PR #9, opened 2022-08-29
DOCS-626: Add `HostEndpointsUpgrade` partial component

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-626

Add `HostEndpointsUpgrade` partial component with dynamic rendering logic based on props.

### DIFF
--- a/src/components/partials/hostendpoints-upgrade.js
+++ b/src/components/partials/hostendpoints-upgrade.js
@@ -1,0 +1,124 @@
+import React from "react";
+import Admonition from "@theme/Admonition";
+import CodeBlock from "@theme/CodeBlock";
+
+function PriorToVersion(props) {
+	if (props.prodname === "Calico") {
+		return "v3.14";
+	}
+
+	return "v3.0";
+}
+
+function ApplyPolicyScript(props) {
+	if (props.prodname === "Calico") {
+		return (
+			<CodeBlock language="bash">
+				calicoctl apply -f - {"<"} allow-all-upgrade.yaml
+			</CodeBlock>
+		);
+	}
+
+	return (
+		<CodeBlock language="bash">
+			{props.orch === "OpenShift"
+				? "oc apply -f - < allow-all-upgrade.yaml"
+				: "kubectl apply -f - < allow-all-upgrade.yaml"}
+		</CodeBlock>
+	);
+}
+
+export default function HostEndpointsUpgrade(props) {
+	return (
+		<>
+			<h2 id="host-endpoints">Host Endpoints</h2>
+			<Admonition type="caution">
+				If your cluster has host endpoints with{" "}
+				<code>interfaceName: *</code> you must prepare your cluster
+				before upgrading. Failure to do so will result in an outage.
+			</Admonition>
+			<p>
+				In versions of {props.prodname} prior to{" "}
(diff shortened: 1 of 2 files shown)


src/components/partials/hostendpoints-upgrade.js
```diff
@@ -0,0 +1,124 @@
...
+				change in default behavior for all-interfaces host endpoints: in
+				the absence of policy the default behavior is to{" "}
+				<strong>drop traffic</strong>. This default behavior is
+				consistent with &quot;named&quot; host endpoints (which specify
+				a named interface such as &quot;eth0&quot;); named host
+				endpoints drop traffic in the absence of policy.
+			</p>
+			<p>
+				{/*
+          TODO [version]: Display correct version. From source:
```
alexeymagdich-tigera, line 61: @doucol I suppose we need to find a way to support `{{page.version}}` variable from source docs in our Docusaurus project, but I noticed that you removed it from your `reqs-sys` component. Is it intentioned and we don't want use it anymore?

src/components/partials/hostendpoints-upgrade.js
```diff
@@ -0,0 +1,124 @@
...
+			</CodeBlock>
+			<p>
+				Now that the nodes with an all-interfaces host endpoint are
+				labeled with <strong>host-endpoint-upgrade</strong>, we can
+				create a policy to log and allow all traffic going into or out
+				of the host endpoints temporarily:
+			</p>
+			<CodeBlock language="bash">
+				{`cat > allow-all-upgrade.yaml <<EOF
+apiVersion: projectcalico.org/v3
```
alexeymagdich-tigera, line 95: It looks like a broken indentation, but it's not. It renders correctly:

image

But if change indentation to desired shape, something like that:

```
<CodeBlock language="bash">
  {`cat > allow-all-upgrade.yaml <<EOF
      apiVersion: projectcalico.org/v3
      kind: GlobalNetworkPolicy
      metadata:
        name: allow-all-upgrade
      spec:
        selector: has(host-endpoint-upgrade)
        types:
        - Ingress
        - Egress
        ingress:
        - action: Log
        - action: Allow
        egress:
        - action: Log
        - action: Allow
      EOF`}
  </CodeBlock>
```

Then the result will be as follows:

image

Actually, it's a JS template strings limitation. Yes, it looks awful, but we can avoid it using https://www.npmjs.com/package/endent or leave it as is if it's not a problem :)

doucol: Leave it as-is for now. Please put a comment in there describing the situation. Thanks!

src/components/partials/hostendpoints-upgrade.js
doucol: yeah, i took it out because it seemed unnecessary - we will likely have versioned docs, in which case, I don't know why adding a version here makes any difference. And, if we don't version our docs, it makes no sense. But, perhaps I'm missing something and I'll later realize we need it! :-)

doucol (approved):

src/components/partials/hostendpoints-upgrade.js
alexeymagdich-tigera: Okay 👍

PR #15, opened 2022-08-31
DOCS-635: Add `ConfigureManagedCluster` partial component

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-635

### DIFF
--- a/src/theme/MDXComponents.js
+++ b/src/theme/MDXComponents.js
@@ -13,18 +13,18 @@ import EnvironmentFile from '@site/src/components/partials/environment-file';
 import ConfigureManagedCluster from '@site/src/components/partials/configure-managed-cluster';
 
 export default {
-	// Re-use the default mapping
-	...MDXComponents,
+  // Re-use the default mapping
+  ...MDXComponents,
 
-	// all of our partials
-	ReqsSys,
-	ReqsKernel,
-	HostEndpointsUpgrade,
-	EnterpriseLink,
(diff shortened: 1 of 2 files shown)


src/theme/MDXComponents.js
```diff
@@ -10,19 +10,21 @@ import InstallOpenshiftBeforeYouBegin from '@site/src/components/partials/instal
 import CalicoWindowsInstall from '@site/src/components/partials/calico-windows-install';
 import PodCidrSed from '@site/src/components/partials/pod-cidr-sed';
 import EnvironmentFile from '@site/src/components/partials/environment-file';
+import ConfigureManagedCluster from '@site/src/components/partials/configure-managed-cluster';
 
 export default {
-    // Re-use the default mapping
```
alexeymagdich-tigera, line 15: @doucol I think we should decide how many spaces we want to use for tabs - 2 or 4. The initial Docusaurus setup uses 2 spaces so that's the reason why I'm following that approach (I've configured auto-formatting on save in my editor). Also, I think 2 spaces are more convenient for such types of projects as documentation since most of the content is just plain text and it means that more text will be visible without a horizontal scroll.

doucol: yeah, sounds good - I use WebStorm and it allows me to follow a well-known standard. I'll set it to "Javascript Standard Style" - which is 2 spaces

doucol (approved):

PR #16, opened 2022-09-01
add support for context vars, can override including multi-path suppo…

doucol:
add support for context vars, can override including multi-path support which picks up partials

### DIFF
--- a/src/remark/variablesPlugin.js
+++ b/src/remark/variablesPlugin.js
@@ -1,64 +1,66 @@
-const path = require("path")
-const visit = require("unist-util-visit")
+const path = require("path");
+const visit = require("unist-util-visit");
 
-const variables = require(path.resolve("variables"))
-const objProp = require(path.resolve("src/utils/objProp"))
-const varRegex = RegExp(/\{\{[ \t]*([\w.]+)[ \t]*}}/, 'g')
+const variables = require(path.resolve("variables"));
+const objProp = require(path.resolve("src/utils/objProp"));
+const varRegex = RegExp(/\{\{[ \t]*([\w.]+)[ \t]*}}/, "g");
 
 function variablesPlugin(_options) {
-	async function transformer(tree, file) {
-		const contextVariables = getContextVariables(file, variables)
-		visit(tree, () => true, (node) => {
-			for (let prop in node) {
-				if (!Object.prototype.hasOwnProperty.call(node, prop)) continue
-				if (prop === 'type' || typeof node[prop] !== 'string') continue
+  async function transformer(tree, file) {
+    const contextVariables = getContextVariables(file, variables);
+    visit(
+      tree,
+      () => true,
+      (node) => {
+        for (let prop in node) {
+          if (!Object.prototype.hasOwnProperty.call(node, prop)) continue;
+          if (prop === "type" || typeof node[prop] !== "string") continue;
 
-				node[prop] = node[prop].replaceAll(varRegex, (match, varName) => {
-						let varValue
-						for (let cv of contextVariables) {
-							varValue = objProp(cv, varName)
-							if (varValue) return String(varValue)
-						}
-						varValue = objProp(variables.global, varName)
-						if (varValue) return String(varValue)
-						return match
-					},
-				)
-			}
-		})
-	}
+          node[prop] = node[prop].replaceAll(varRegex, (match, varName) => {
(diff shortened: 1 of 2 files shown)


src/remark/variablesPlugin.js
```diff
@@ -1,51 +1,64 @@
...
-				? variables.enterprise
-				: posixFriendlyPath.includes(pathPrefixes.openSource)
-					? variables.openSource
-					: null;
-
-		if (!productVariables) {
-			return;
-		}
+		const contextVariables = getContextVariables(file, variables)
+		visit(tree, () => true, (node) => {
```
alexeymagdich-tigera: If I'm not mistaken, we can just omit the second parameter to allow all types of nodes to be processed.

src/remark/variablesPlugin.js
```diff
@@ -1,51 +1,64 @@
-const path = require("path");
-const visit = require("unist-util-visit");
+const path = require("path")
```
alexeymagdich-tigera: I would suggest configuring your editor to allow semicolons for each line. By the way, I'm using `Prettier` - you can install it and then we'll be able to sync up our settings. Or, moreover, we can add `.prettierrc` and enable the same code styles for all of us.

src/remark/variablesPlugin.js
```diff
@@ -1,51 +1,64 @@
...
-					const varValue = objProp(productVariables, varName);
-					if (!varValue) {
-						return match;
+function getContextVariables(file, variables) {
+	let cvars = []
+	const dpName = 'docsPathPrefix'
+	const posixFriendlyPath = convertToPosixFriendlyPath(file.path)
+
+	for (let p in variables) {
+		if (!Object.prototype.hasOwnProperty.call(variables, p)) continue
```
alexeymagdich-tigera: Do we really need this check? `for ... in` should iterate only on the existing properties of an object.

src/remark/variablesPlugin.js
```diff
@@ -1,51 +1,64 @@
...
+		if (typeof variables[p] === 'object') {
+			if (!Object.prototype.hasOwnProperty.call(variables[p], dpName)) continue
+			const docsPathPrefix = variables[p][dpName]
+			if (Array.isArray(docsPathPrefix)) {
```
alexeymagdich-tigera: As I see in `variables.js` the `docsPathPrefix` field is always an array. Do we need this if/else if statement?

variables.js
```diff
@@ -2,36 +2,50 @@ const variables = {
 	cloud: {
 		prodname: "Calico Cloud",
 		prodnamedash: "calico-cloud",
+		docsPathPrefix: [
```
alexeymagdich-tigera: Do we really want to store it in variables? Looks like it's used only for plugin purposes.

src/remark/variablesPlugin.js
doucol: Good to know, thanks. I think we'll just keep it as-is since I know it works :-) - and it's less of a mystery as to what's going on.

src/remark/variablesPlugin.js
doucol: I have added prettier and configured WebStorm to use it for all formatting on save. It's configured to do so for all javascript and markdown oriented files. If you want to add that .prettierrc file, go for it. Personally, I would prefer we get rid of the semicolons, use single quotes when we can, and use tabs. But I don't care that much. It's nice to just be all consistent with each other. I really don't care for the way it formatted the 'visit' function call though - ugly!

src/remark/variablesPlugin.js
doucol: Normally, yes. But, this is safer as I understand things. There can be times when you will iterate through a property that is not a normal property. Perhaps the world as moved on since I was last coding javascript though! 😄

src/remark/variablesPlugin.js
doucol: We could change that so that it doesn't support a single string. I thought it would be nice, but it's probably over engineered in this case, since this isn't going to be a big public API. I'll remove that. Thanks

variables.js
doucol: I think it's better to keep these values in variables.js since the paths are tightly associated with those variables. This is what allows us support the overriding dynamically.

alexeymagdich-tigera (dismissed):

PR #30, opened 2022-09-13
DOCS-633: Add `componentImagePlugin`

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-633

It's implemented as a plugin (not a component) due to MDX limitations: `component_image` include is used only in code blocks (defined via ```) and it's not possible to render components in such syntax constructions. 

Anyway, it's still can be used in two differented cases:

1. Use inside code blocks:

```
{% component_image("some-component") %}
```

As you see, it's a Jekyll-like syntax.

2. Use inside components (if needed):

```
import componentImage from '@site/src/components/componentImage';`
...
<div>
  {componentImage("some-component")
</div>
```

### DIFF
--- a/src/components/utils/componentImage.js
+++ b/src/components/utils/componentImage.js
@@ -1,0 +1,18 @@
+const getProductVariablesByProdname = require('../../utils/getProductVariablesByProdname');
+
+function componentImage(comp, prodname) {
+  const productVariables = getProductVariablesByProdname(prodname);
+
+  if (!productVariables) {
+    console.error(`Invalid "prodname": ${prodname}`);
+
+    return;
+  }
+
+  const component = productVariables.components[comp];
+  const registry = component.registry ? `${component.registry}/` : '';
+
+  return `${registry}${component.image}:${component.version}`;
+}
+
+module.exports = componentImage;
--- a/src/remark/componentImagePlugin.js
+++ b/src/remark/componentImagePlugin.js
@@ -1,0 +1,47 @@
+const path = require('path');
+const visit = require('unist-util-visit');
+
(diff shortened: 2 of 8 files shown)


src/components/utils/componentImage.js
```diff
@@ -0,0 +1,18 @@
+const getProductVariablesByProdname = require('../../utils/getProductVariablesByProdname');
```
alexeymagdich-tigera, line 1: We need to use CJS modules import/export here because this file is used in a plugin that is executed on build time. Build time uses CJS modules - it's Docusaurus default configuration so I'm not sure we can switch to ESM modules quite easily.

The same for some other files below.

src/remark/componentImagePlugin.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (path.includes(pathPrefixes.enterprise)) {
+    return variables.enterprise.prodname;
+  }
+
+  if (path.includes(pathPrefixes.openSource)) {
+    return variables.openSource.prodname;
+  }
+}
```
doucol: Unfortunately, these paths can exist across the different products (like CE having a folder named 'calico-cloud'). You'll need to be more specific about the path by prefixing each path with '/docs/'.

doucol (approved):

src/components/utils/componentImage.js
doucol: Noted, thanks

src/remark/componentImagePlugin.js
alexeymagdich-tigera: Good catch, thank you 👍

PR #40, opened 2022-09-30
DOCS-818: Implement dynamic (runtime) TOC

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-818

> **Caution**: Since this approach is based on swizzling which is potentially [could be risky](https://docusaurus.io/docs/swizzling#what-is-safe-to-swizzle) we need to revert these changes as soon as [Docusaurus TOC issue](https://github.com/facebook/docusaurus/issues/3915) will be resolved and published.

Solution is based on [swizzling](https://docusaurus.io/docs/swizzling) - the way to "override" the behavior of built-in Docusaurus' components.

Every doc page is based on `DocItem` component which is composed of several other components like `DocItemLayout`, `DcoItemFooter`, etc. In our case, we are interested in `DocItemTOCDesktop` and `DocItemTOCMobile` components that are responsible for displaying TOC for desktop and mobile devices respectively. You can explore the source code [here](https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-classic/src/theme/DocItem).

Under the hood, these components are rendering TOC composed in build-time by remark plugin - we want to override it. Current changes parses the page content, detects all necessary headings (based on config/front-matter) and generates correct TOC - it's implemented via `useToc` hook. Then we use this hook inside swizzled components.

On the screenshot below I've shown the generated TOC structure for one of the pages and highlighted TOC items that were missing before the changes (since they are come from imported partials). The second screenshot shows the rendered partial and you can see the TOC item for it.

image
image

### DIFF
--- a/src/theme/DocItem/TOC/Desktop/index.js
+++ b/src/theme/DocItem/TOC/Desktop/index.js
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import { ThemeClassNames } from '@docusaurus/theme-common';
+import { useDoc } from '@docusaurus/theme-common/internal';
+import TOC from '@theme/TOC';
+import useToc from '@site/src/utils/useToc';
+
+export default function DocItemTOCDesktop() {
+  const {
+    frontMatter: { toc_min_heading_level, toc_max_heading_level },
+  } = useDoc();
+  const toc = useToc(toc_min_heading_level, toc_max_heading_level);
+
+  return (
+    <TOC
+      toc={toc}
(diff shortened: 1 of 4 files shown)


ctauchen (approved):

ctauchen:
@alexeymagdich-tigera Is this ready to merge? Or should @denis-alkhelali take a look?

denis-alkhelali:
> 

@ctauchen @alexeymagdich-tigera let me take a look

denis-alkhelali (approved):
Perfecto :)

PR #44, opened 2022-10-05
DOCS-831: Fix TOC for doc pages composed only by components

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-831

Here's the example of the doc page that is composed only of partial components - TOC is displayed correctly. The approach is based on swizzling - the same as https://github.com/tigera/docs/pull/40.

image

### DIFF
--- a/src/theme/DocItem/Layout/index.js
+++ b/src/theme/DocItem/Layout/index.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import clsx from 'clsx';
+
+import { useWindowSize } from '@docusaurus/theme-common';
+import { useDoc } from '@docusaurus/theme-common/internal';
+import DocItemPaginator from '@theme/DocItem/Paginator';
+import DocVersionBanner from '@theme/DocVersionBanner';
+import DocVersionBadge from '@theme/DocVersionBadge';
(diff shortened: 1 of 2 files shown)


src/theme/DocItem/Layout/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+import useToc from '@site/src/utils/useToc';
+
+import styles from './styles.module.css';
+/**
+ * Decide if the toc should be rendered, on mobile or desktop viewports
+ */
+function useDocTOC() {
+  const { frontMatter } = useDoc();
+  const toc = useToc();
```
alexeymagdich-tigera, line 23: This is an actual change. We want to use our ("correct") TOC to detect its size. Previously, the condition below (`const canRender = !hidden && toc.length > 0;`) was based on TOC by Docusaurus. In the case of the page composed only of components - TOC was always empty.

src/theme/DocItem/Layout/styles.module.css
```diff
@@ -0,0 +1,10 @@
+.docItemContainer header + *,
```
alexeymagdich-tigera, line 1: Styles by Docusaurus - no changes.

doucol (approved):

PR #72, opened 2022-11-15
Added proxy config for CE 3.14, 3.13, 3.12

ctauchen:
This adds proxy configurations for CE 3.14, 3.13, and 3.12.

All of our old docs sites are to follow this model. `netlify.toml` will contain proxy definitions for every Jekyll site we want to continue serving in our docs portal.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -141,8 +141,92 @@ package = "@netlify/plugin-lighthouse"
   headers = {X-From = "Netlify"}
 
 [[redirects]]
-  from = "/calico-enterprise/3.12/*"
+  from = "/v3.12/*"
   to = "https://tigera-v3-12.netlify.app/v3.12/:splat"
+  status = 200
+  headers = {X-From = "Netlify"}
+
+[[redirects]]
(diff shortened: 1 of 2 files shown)


ctauchen:
A note: The proxies are working correctly. But the JS files aren't loading. According to @penkeysuresh, may have something to do with `baseURL` in old docs.

doucol (approved):

Makefile
```diff
@@ -10,6 +10,7 @@ PRODUCT?=calico
 
 build: init
 	yarn build
+	cp _redirects build/_redirects
```
doucol: If you put the _redirects file into /static, it will get copied for you automatically

doucol (changes requested):

Makefile
ctauchen: That sounds much more dependable. Will change, thanks!

ctauchen:
Closing and moving changes to https://github.com/tigera/docs/pull/74/

doucol (approved):

PR #80, opened 2022-11-18
Fix versions dropdown

denis-alkhelali:
I noticed that Docusaurus allows to have versions dropdown to only one of the multi-instances and the instance must be set in docusaurus config. So, e.g. if we set Calico, then regardless of the selected product, there're will be only Calico's versions in dropdown with links to Calico.
I wrapped DocsVersionDropdownNavbarItem to pass product id based on the users current location

'+ small refactoring

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -92,7 +92,6 @@ const config = {
           {
             type: 'docsVersionDropdown',
             position: 'right',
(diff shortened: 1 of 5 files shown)


alexeymagdich-tigera (approved):

ctauchen (approved):
This is great -- thanks! lgtm

PR #146, opened 2022-12-16
Test redirects

alexeymagdich-tigera:


doucol:
@ctauchen I was looking at this today... still not sure why it's not working. I would spend more time on it, but I'm questioning whether we should be doing this at all.  I think the way it's working now is good, let's chat next week.

alexeymagdich-tigera:
@ctauchen I suppose, we can close it?

ctauchen:
@alexeymagdich-tigera Yes, thanks for flagging this.

PR #158, opened 2022-12-16
Add container threat defense

bmckercher123:
## Missing entry in sidebars.js for CC


doucol:
@ctauchen We may need to do this in the source docs, if this is needed right now. Or, we'll need to add some logic into the migration script to do this. What are your thoughts?

ctauchen:
Fixed in https://github.com/tigera/docs/pull/159.

Those source files got it by mistake. In an older version of the script, I mixed up the --path-rename for the CC security directory.

PR #175, opened 2022-12-21
Delete-landing-page-policy-lifecycle

bmckercher123:
### Delete landing page for singleton topic for CC and CE

- Updated sidebars.js for CC and CE
- Deleted landing page, plus CE had an extra file (old) that should be removed from the script (Policy impact preview). 
- Able to build without errors
- PR to update source files for CE: https://github.com/tigera/calico-private/pull/5515.
- You just need to update script and coordinate these PRs.


ctauchen:
@bmckercher123 Two changes before we can merge:
* Remove extra files. I think you're running an old version of the script. I've since fixed the bug that creates a few files under a `security` directory at root. A quick fix is to just delete the files and amend your commit. 
* Can you add the navigation changes to CE 3.24 and 3.25?

PR #210, opened 2023-01-17
CC-fix-wording-for-clarity

bmckercher123:
## Request from Support to clarify

- https://tigera.atlassian.net/browse/DOCS-1185
- For CC 3.15 and next

### DIFF
--- a/calico-cloud/users/user-management.mdx
+++ b/calico-cloud/users/user-management.mdx
@@ -36,17 +36,15 @@ The Owner role cannot be assigned to new users. The only Owner is the user who c
 
 :::
 
-## Add your own IdP
+## Add your own identity provider
 
-{{prodname}} supports the following IdPs:
+{{prodname}} works with any identity provider that supports [OpenID Connect](https://openid.net/connect/). For example, OKTA, Google, and Azure AD.
 
-- OIDC-based Auth providers (for example Google, Azure AD)
+To add an identity provider, open a [Support ticket](https://support.tigera.io/).
 
-To add an IdP, open a [Support ticket](https://support.tigera.io/).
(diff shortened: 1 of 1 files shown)


calico-cloud/users/user-management.mdx
```diff
@@ -40,7 +40,7 @@ The Owner role cannot be assigned to new users. The only Owner is the user who c
 
 {{prodname}} supports the following IdPs:
 
-- OIDC-based Auth providers (for example Google, Azure AD)
+- Any OIDC-based Auth providers (for example, OKTA, Google, Azure AD)
```
ctauchen: Suggestion:

{{prodname}} works with any identity provider that supports [OpenID Connect](https://openid.net/connect/).

To add an identity provider, open a [Support ticket](https://support.tigera.io/).

ctauchen: I'm also in favor of s/IdP/identity provider in all instances.

ctauchen (commented):
Thanks!

A few suggestions for you to consider.

If this is for CE 3.15+, then you'll also want to add this change to the file in 3.15 and CC (if applicable).

calico-cloud/users/user-management.mdx
bmckercher123: Agree on "identity provider". Fixed. Though I did leave the examples because Support is saying that users do not know off-hand that OTKA is OIDC/OpenID Connect. They know the names of the idp, but not necessarily the underlying protocol.

bmckercher123: Fixed in CC.

ctauchen (approved):
LGTM

PR #215, opened 2023-01-18
DOCS-1184: Fixes for OS helm chart links

ctauchen:
Changing the URL for OS helm chart links. This reproduces the logic from the old docs site.

`helm repo add projectcalico https://docs.tigera.io/calico/charts`
* https://docs.tigera.io/calico/charts is proxied to an S3 location

We'll maintain the same proxy rule at projectcalico.docs.tigera.io to catch any traffic to that URL.


ctauchen:
CC @caseydavenport

PR #224, opened 2023-01-19
DOCS-1090: Adds versioning configuration

ctauchen:
Fixes https://tigera.atlassian.net/browse/DOCS-1090

* Corrects the versioned URL structure for version 'latest'
* Adds/removes banners with version-related message.
* Sets CE 3.14 as 'latest'
* Adds 'beta' tag and unreleased banner to 3.15

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -330,6 +330,23 @@ const config = {
         path: 'calico',
         routeBasePath: 'calico',
         editCurrentVersion: true,
+        versions: {
+          current: {
+            label: 'Next',
+            path: 'next',
+            banner: 'unreleased',
+          },
+          '3.25': {
+            label: '3.25',
+            path: '3.25',
+            banner: 'none',
+          },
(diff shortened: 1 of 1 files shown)


alexeymagdich-tigera (approved):

denis-alkhelali:
@ctauchen looks good.
one note. Ticket says "These links should always link to the current version.". Not sure what does it mean, if "current" here means currently selected version (or last selected version) then it's fine, but that how it always worked. If "current" means always latest version, then it's not fixed. I.e if I select 3.24 then the link in header will lead me to OS 3.24

ctauchen:
@denis-alkhelali Thanks. I wasn't very precise in my wording, and I forgot entirely about how Docusaurus remembers what you've been viewing. 

My test: When I open in incognito mode and click on Calico Enterprise from the Docs drop-down, it opens 3.14 (and for OS 3.25. 

Thanks for following up and clarifying.

PR #245, opened 2023-01-24
DOCS-1193: Load swagger ui styles only on demand

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-1193


ctauchen (approved):
LGTM!

PR #246, opened 2023-01-24
DOCS-1180: Fixes links to tutorial manifest files

ctauchen:
https://tigera.atlassian.net/browse/DOCS-1180

We don't need tutorial manifest files to be hosted with the proper manifest files. Here, I've added these files to `/static/files/` and changed links accordingly.

None of these files have diffs across product or version. So I'm leaving these unversioned. We can change this later if we need.


denis-alkhelali (commented):
@ctauchen maybe let's put this variable to global variables? variables.js in root folder. Variables from specific variables.js files take precedence over global variables, in case smth must be overriden.
Up to you, maybe you'll want to get rid of the concept of global variables

ctauchen:
I think we'll redo our global variables all at once another time.

doucol:
I'm curious why we wouldn't use just a generic 'filesUrl' - rather than something specific for tutorials?

PR #263, opened 2023-01-27
DOCS-1198: Rename "special" partial components

alexeymagdich-tigera:
https://tigera.atlassian.net/browse/DOCS-1198

Now they look as usual partial components - it'll help to avoid any confusions in the future.


denis-alkhelali (approved):
Nice

PR #266, opened 2023-01-27
Docusaurus 2.3 released, so remove plugin-content-docs bug workaround

denis-alkhelali:


alexeymagdich-tigera (approved):

doucol (approved):
Nice - thanks for following up with this!

PR #273, opened 2023-01-31
Fixes CC codeblock with images

ctauchen:
Adding variable definitions and an image list for CC.

### DIFF
--- a/calico-cloud/get-started/connect/setup-private-registry.mdx
+++ b/calico-cloud/get-started/connect/setup-private-registry.mdx
@@ -51,7 +51,40 @@ helm repo add calico-cloud {{clouddownloadbase}}/charts
 INSTALLER_IMAGE="{{cloudoperatorimage}}:$(helm show chart calico-cloud/calico-cloud | grep version: | sed -e 's/version: *//' -e 's/+/-g/')"
 IMAGES=(
 $INSTALLER_IMAGE
-image-list-v3.14.1-16 )
+docker.io/calico/pod2daemon-flexvol:v3.23.2
+quay.io/tigera/operator:v1.27.11
+quay.io/tigera/cnx-apiserver:v3.14.1
+quay.io/tigera/compliance-benchmarker:v3.14.1
+quay.io/tigera/compliance-controller:v3.14.1
+quay.io/tigera/compliance-reporter:v3.14.1
+quay.io/tigera/compliance-snapshotter:v3.14.1
+quay.io/tigera/deep-packet-inspection:v3.14.1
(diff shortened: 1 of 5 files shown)


alexeymagdich-tigera (commented):
`modules-3.16/_test.mdx` - I suppose this file should be removed?

calico-cloud/get-started/connect/setup-private-registry.mdx
```diff
@@ -51,7 +51,40 @@ helm repo add calico-cloud {{clouddownloadbase}}/charts
 INSTALLER_IMAGE="{{cloudoperatorimage}}:$(helm show chart calico-cloud/calico-cloud | grep version: | sed -e 's/version: *//' -e 's/+/-g/')"
 IMAGES=(
 $INSTALLER_IMAGE
-image-list-v3.14.1-16 )
+docker.io/calico/pod2daemon-flexvol:v3.23.2
```
tmjd, line 54: This doc was populated before as part of doc rendering I believe so the `image-list` file (which was a symlink to the correct version, ie. `image-list-v3.14.1-16`) would be "filled in" to this doc. But now I guess we need to hand copy in this list into this document?
Is there a way with docusaurus to include the contents of one file into another? Because that would be ideally what is done here.

alexeymagdich-tigera: @tmjd something similar is possible: https://docusaurus.io/docs/markdown-features/react#importing-code-snippets

I didn't try it, but I suppose it should work as follows:
- Define file with content (e.g. in `static` directory)
- Import it via raw-loader (as in the example from the link above)
- Inject into the code block

ctauchen: @tmjd For a single image list that appears in one place, I don't see the use in complicating things by keeping a separate file.

We can leave that option open. But for now, I'd like to just slap this in the code block. If it ain't broke ...

tmjd: The reason for having a separate file is that it can be automated to grab that list and include/update it, not to reduce duplication.
Previously we only had to update the cloudmanifests version in the config and then run a command.

Now we've got to:
1. update the cloudmanifests version (because it is needed in other places)
2. know it is necessary to update the image list by hand
3. Find the location of the image list and get the correct version
4. Copy the list of images
5. Find this file that needs to be manually updated
6. Replace the existing lines with the copied list.

Requiring these steps also stops us from being able to have CI that ensures this file has been updated (or at least it would be much more complicated).

So in my mind manually updating this instead of having a separate file is complicating things.

PR #283, opened 2023-02-01
[RS-641] - CTD now supports all CC platforms.

mikestephen:


ctauchen (commented):
Great stuff, thanks!

I'm still getting used to the process, so forgive me. Could you also make the corresponding change in /calico-cloud/threat/container-threat-detection.mdx?

mikestephen:
> Great stuff, thanks!
> 
> I'm still getting used to the process, so forgive me. Could you also make the corresponding change in /calico-cloud/threat/container-threat-detection.mdx?

Ah yes, that makes sense, I'll add it now.

ctauchen (approved):
LGTM

PR #290, opened 2023-02-02
CE-RKE-UI-rework

bmckercher123:
### Combine RKE2 existing doc and Rancher UI into single doc

- Valid for 3.16 and 3.15 because there were no code changes
- Update title to reflect RKE2 Government
- Edit doc


ctauchen (dismissed):
There are diffs between next and 3.15. Appears that next has some edits and 3.15 needs updating.

LGTM, pending alignment of next and 3.15.

PR #298, opened 2023-02-03
Fix IPPools in EGW docs.

sridhartigera:
EGW docs mention the ippool cidr as /31. As a result, we can have only 2 EGW pods. Whenever a change is made to the EGW resource, a roll-out is triggered. This results in a new pod, but the IPPool doesn't have enough IPs. 

Hence the IPPools should be large enough to include terminating pods.

### DIFF
--- a/calico-enterprise/networking/egress/egress-gateway-aws.mdx
+++ b/calico-enterprise/networking/egress/egress-gateway-aws.mdx
@@ -501,16 +501,16 @@ For the "west-1" availability zone:
 - IP pool "egress-red-west-1", CIDR `100.64.0.4/30` (the first non-reserved /30 CIDR in the VPC subnet). These
   addresses will be used for "red" egress gateways in the "west-1" AZ.
 
-- IP pool "egress-blue-west-1", CIDR `100.64.0.6/30` (the next 4 IPs from the "west-1" subnet). These addresses
+- IP pool "egress-blue-west-1", CIDR `100.64.0.8/30` (the next 4 IPs from the "west-1" subnet). These addresses
   will be used for "blue" egress gateways in the "west-1" AZ.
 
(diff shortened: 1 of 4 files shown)


calico-enterprise/networking/egress/egress-gateway-aws.mdx
```diff
@@ -498,18 +498,18 @@ it will take an IP from the appropriate pool.
 
 For the "west-1" availability zone:
 
-- IP pool "egress-red-west-1", CIDR `100.64.0.4/31` (the first non-reserved /31 CIDR in the VPC subnet). These
+- IP pool "egress-red-west-1", CIDR `100.64.0.4/30` (the first non-reserved /30 CIDR in the VPC subnet). These
   addresses will be used for "red" egress gateways in the "west-1" AZ.
 
-- IP pool "egress-blue-west-1", CIDR `100.64.0.6/31` (the next 2 IPs from the "west-1" subnet). These addresses
+- IP pool "egress-blue-west-1", CIDR `100.64.0.6/30` (the next 4 IPs from the "west-1" subnet). These addresses
```
fasaxc:
```suggestion
- IP pool "egress-blue-west-1", CIDR `100.64.0.8/30` (the next 4 IPs from the "west-1" subnet). These addresses
```
A /30 has 4 IPs in it so `100.64.0.4/30` ends at `100.64.0.7` and the next /30 is `100.64.0.8/30`.  Need to be super careful and double check with a CIDR calculator; took me 2-3 attempts last time I adjusted these CIDRs!

fasaxc (approved):
LGTM, thanks for fixing the AZ confusion.

bmckercher123 (approved):
LGTM

PR #300, opened 2023-02-03
CC: Adjust version variables

tmjd:
I noticed the Image Assurance version needed to be updated when reviewing the CC docs so I've done that. Also I noticed we still had v3.14 in the variables files for CC and I don't think that is correct so I've updated those too.


ctauchen (dismissed):
Lgtm

tmjd:
The image assurance version and the version for the cloud manifests will need a bump.

tmjd:
I've updated the changes here to include the manifest update that was done today (v3.15.1-0 to v3.15.1-1) and that included an update of image assurance to 1.3.4.
I've updated both the unversioned and the version-3.15 for CC.

penkeysuresh (approved):
LGTM from Image Assurance perspective

bmckercher123 (approved):
Approve

PR #303, opened 2023-02-06
CC-CE-Add policy order example

bmckercher123:
### Add policy or example to network policy tutorial

- Applies to CE and CC (latest and next)
- Request from Support to provide example of policy order


ctauchen (dismissed):
LGTM, but should be included for all relevant products/versions

PR #305, opened 2023-02-06
Sk update yaml manifest private registry

sujeet-kr:
Fix the manifest example rendering.

image


sujeet-kr:
cc @ctauchen @bmckercher123

ctauchen:
@sujeet-kr Thanks for pointing this out! It's an error in the migration from Jekyll to Docusaurus.

We're going to take a different approach to fixing this here and elsewhere in the docs. One that lets us keep the emphasis on the line. See https://tigera.atlassian.net/browse/DOCS-1237.

FYI @girishshankaran

PR #306, opened 2023-02-07
BPF Include Topology Aware Hints docs

ghost:
Reference : https://tigera.atlassian.net/browse/CORE-8593
Here is docs PR to accompany the Topology Aware Hints task code implementation https://github.com/projectcalico/calico/pull/7241


tomastigera:
Is this something we want to include in calico docs? Seems rather like an internal howto. For the product, this should just work out of the box the same as with iptables and k8s' kube-proxy :thinking:

ctauchen:
After discussion with @StevenTigera, I think we're going to keep this as RN-only. The code appears to remove a limitation, rather than add a new Calico feature.

PR #322, opened 2023-02-08
fix broken link with Calico Enterprise 3.14 Openshift install docs

radTuti:


ctauchen (approved):
LGTM! Thanks for finding this!

PR #325, opened 2023-02-08
CC-fix-graphic-alignment

bmckercher123:
### Fix misaligned graphic for connect cluster

- CC 3.15 and 3.16


ctauchen:
@bmckercher123 I don't see a difference in the image alignment. Are you trying to put the image in line with the next, but not appearing after the line finishes?

bmckercher123:
The published version is not aligned. This fixes the problem.

ctauchen:
I see now -- you fixed the one, but I was looking at another. See https://github.com/tigera/docs/pull/392

PR #326, opened 2023-02-08
IA scanner bulk image upload

ghost:


calico-cloud/image-assurance/scan-image-registries.mdx
```diff
@@ -134,6 +136,23 @@ You can scan images and report results back to {{prodname}}, or scan images loca
...
+  "images": [
+    "ubuntu:latest",
+    "alpine:latest"
+  ]
+}
+```
+
+```shell
+./tigera-scanner scan --input_file images.json --output_file results.json
+```
```
stevegaossou, line 154: Should we also include an example somewhere to demonstrate scanning multiple images using the command-line arguments method of specifying multiple images?

cut-new-version.sh
```diff
@@ -5,7 +5,7 @@ version=$2
 
 echo 'Cutting a new version...'
 
-docusaurus docs:version:$product $version
+npx docusaurus docs:version:$product $version
```
stevegaossou: Was this change intended? 

I'm not sure (I'm completely unfamiliar with the new docs tooling), but just wanted to check.

calico-cloud/image-assurance/scan-image-registries.mdx
stevegaossou: Should we also include an example somewhere to demonstrate scanning multiple images using the command-line arguments method of specifying multiple images?

calico-cloud/image-assurance/scan-image-registries.mdx
```diff
@@ -153,6 +172,8 @@ Key names must match the full name of arguments passed to the CLI. The configura
 | --warn_threshold        | -w        | CC_WARN_THRESHOLD        | CVSS threshold for Warn scan results. Range from 0.0 - 10.0.                                                                |
 | --fail_threshold        | -f        | CC_FAIL_THRESHOLD        | CVSS threshold for Fail scan results. Range from 0.0 - 10.0.                                                                |
 | --vulnerability_db_path | -p        | CC_VULNERABILITY_DB_PATH | Path to a folder to store vulnerability data (defaults to `$XDG_CACHE_HOME`; if it is not set, defaults to `$HOME/.cache`). |
+| --input_file            | -i        | CC_INPUT_FILE            | Input file containing image URLs.                                                                                           |
```
stevegaossou: I think we should have some mention in the [export scan results section](https://github.com/tigera/docs/blob/main/calico-cloud/image-assurance/scan-image-registries.mdx#export-scan-results) that there is an option to export a batch of images and that file can then be fed into the CLI scanner using this input option.

cut-new-version.sh
ghost: Yes it's intended. Npx runs node binaries. Instead of devs having different versions on docusaurus installed on their local machine, npx ensures a specified version is used.

PR #335, opened 2023-02-10
Fixes bad xref in calico cloud

ctauchen:
We removed a file. But there were still cross reference to that file. The PR removes the cross references.


ctauchen:
@girishshankaran Can you take a look?

girishshankaran (approved):
lgtm

PR #339, opened 2023-02-12
Always use calico kubeconfig for eks as eks set KUBECONFIG env by

song-jiang:
Since EKS v1.23, Windows nodes set KUBECONFIG Env by default and point it to c:\ProgramData\kubernetes\kubeconfig. 
We need to set calico-kube-config explicitly in our config.ps1, otherwise it will take the default ENV variable. 

We can modify config.ps1 but that means we need to have a patch release. This PR walk around the issue by modifying install-windows-scripts.ps1 only.

### DIFF
--- a/static/calico-enterprise/3.15/scripts/install-calico-windows.ps1
+++ b/static/calico-enterprise/3.15/scripts/install-calico-windows.ps1
@@ -406,6 +406,16 @@ function InstallCalico()
     Write-Host "`n{{site.prodnameWindows}} installed`n"
 }
 
+function SetConfig {
+    param(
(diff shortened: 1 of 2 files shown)


coutinhop (approved):
LGTM

ctauchen:
@coutinhop I've merged this one, given the urgency, but I have a few follow-up questions:
* Does this same change need to go in the similar file for /next?
* Do similar changes need to be added for OS?
* (IMPORTANT!) Are the same changes being made elsewhere? Hosting these files in the docs site is a TEMPORARY fix until we get a permanent downloads location set up. I plan to delete these files as soon as those new download locations are ready.

PR #343, opened 2023-02-13
DOCS-1265: Threat feed updates to CE 3.15, CC 3.15 and CC next releases

girishshankaran:
https://tigera.atlassian.net/browse/DOCS-1265

See also https://tigera.atlassian.net/browse/DOCS-979

Adding the updates to these versions: CE 3.15, CC 3.15, and CC next.

And


ctauchen (approved):
LGTM

PR #353, opened 2023-02-13
Add note for OCP users using federation

rene-dekker:


calico-cloud/multicluster/kubeconfig.mdx
```diff
@@ -340,6 +340,14 @@ spec:
 
 Congratulations! You have completed configuration of federated endpoint identity.
 
+:::note
+
+Openshift users are advised to disable the `network.openshift.io/RestrictedEndpointsAdmission` admission controller
```
fasaxc: I think you should be more forceful.  IIUC, service IP federation simply doesn't work with that controller enabled so we should be explicit.

calico-cloud/multicluster/kubeconfig.mdx
```diff
@@ -53,6 +53,11 @@ Although this example applies remote cluster configurations symmetrically across
 
 ## Before you begin
 
+**OpenShift users**
+
+{{prodname}} supports federated endpoint identity for OpenShift. However, the [federated services feature](services-controller.mdx)
+cannot be used, because OpenShift currently does not allow users to disable the `RestrictedEndpointsAdmission` plugin.
```
bmckercher123, line 59: {{prodname}} supports federated endpoint identity for OpenShift. However, the [federated services feature](services-controller.mdx) is not supported because OpenShift does not currently support disabling the `RestrictedEndpointsAdmission` plugin.
```suggestion
cannot be used, because OpenShift currently does not allow users to disable the `RestrictedEndpointsAdmission` plugin.
```

bmckercher123 (changes requested):
Looks like you updated CE 3.14 (not necessary, but ok) and 3.15. But you missed CE 3.16. (Root is 3.17).

bmckercher123 (approved):
Suggestions only.

PR #355, opened 2023-02-14
CC-new-architecture-doc

bmckercher123:
### Add Calico Cloud architecture diagram

- For 3.15/next


calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+{{prodname}} architecture is based on the Calico Enterprise multi-cluster management feature. {{prodname}} manages the control plane, and you connect your clusters (called **managed clusters**) to the control plane. Communication between the {{prodname}} control plane and managed clusters is secured using TLS tunnels. 
+
+![calico-architecture](/img/calico-cloud/cc-architecture.svg)
+
+The components that secure communications between the {{prodname}} control plane and managed clusters are:
+
+- **{{prodname}} tunnel server** - accepts secure TLS connections from managed clusters
+- **Guardian** - an agent that runs in each managed cluster that proxies communication between {{prodname}} components and managed cluster components
+
+All connections go through the {{prodname}} tunnel server and Guardian. The only exception is installation and upgrade when managed clusters connect to {{prodname}} using TLS connections to register and get install/update resources, and report the status of the install/update.
```
tmjd: I'd suggest the 2nd sentence be changed to 
```
The only exception is during installation and upgrade when managed clusters connect to {{prodname}} using TLS connections to get install/update resources, register the cluster, and report the status of the install/update.
```

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+The {{prodname}} tunnel is initiated by Guardian on the managed cluster. The {{prodname}} control plane does not initiate new connections to the managed cluster outside of the tunnel. However, there are connections that go through the {{prodname}} tunnel server that are initiated from the control plane; for example, when a user does something in the Manager UI, or when the Tigera operator installs or configures access to a component.
```
tmjd: `or when the Tigera operator installs or configures access to a component`
I don't think we should specifically call out the tigera operator here, we don't need to convey that we use the tigera/operator in the management cluster.
I think we should make it more general, something like `or when configuration needs to be pushed into the managed cluster`.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+![calico-architecture-diagram](/img/calico-cloud/cc-arch-diagram.png)
+
+| Component                      | Description                                                  | Ports/Protocol                                               |
+| ------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| User access to Manager UI      | Authenticated users can access the browser-based Manager UI, which provides network traffic visibility and troubleshooting, centralized multi-cluster management, threat-defense, container threat detection, policy lifecycle management, scan images for vulnerabilities, and compliance for multiple roles/stakeholders. | &bull; Port 443 to {{prodname}} tunnel server                |
+| Logs                           | Provides a write API for an Elasticsearch backend, ensuring secure access for all components across all managed clusters.  Data for each managed cluster is isolated and protected against unauthorized access. | n/a                                                          |
```
tmjd: We should replace the first sentence with something like. No need to call out that we're using Elasticsearch here.
`Storage where all logs will be stored.`

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+| {{prodname}} tunnel server     | Communicates with managed clusters by creating secure TLS tunnels. | &bull; Port 9000 to managed clusters                         |
```
tmjd: I think `Port 9000 to managed clusters` should be changed to `Port 9000 from managed clusters` since the tunnel server does not initiate the connections.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+| Guardian                       | An agent running in each managed cluster that proxies communication between the {{prodname}} tunnel server and your managed cluster. Secured using TLS tunnels. | &bull; Port 9000 to tunnel server<br />&bull; TCP 6443 to Kubernetes API server<br />&bull; TCP 6443 from {{prodname}} components |
+| Intrusion detection controller | Handles integrations with threat intelligence feeds and {{prodname}} custom alerts. Installs Kibana dashboards for viewing jobs through the Kibana UI. | &bull; TCP 8080 to Guardian<br />&bull; TCP 6443 to Kubernetes API server |
```
tmjd: I don't think we need to include `Installs Kibana dashboards for viewing jobs through the Kibana UI.`,  that happens in the management cluster.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
+---
+description: Understand the main components of Calico Cloud.
+---
+
+# Calico Cloud architecture
+
+## {{prodname}} security
```
ctauchen, line 7: This title, which is quite broad, doesn't quite fit the material. It seems that this section describes the basic architecture of the product. And maybe focuses on the connection between managed/management clusters (is that why Security?).

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+|                                | **compliance-server** <br />Provides the API for listing, downloading, and rendering reports, and RBAC by performing authentication and authorization through the Kubernetes API server. RBAC is determined from the users RBAC for the Global report type and Global report resources. |                                                              |
+|                                | **compliance-benchmarker**<br />A daemonset that runs checks in the CIS Kubernetes Benchmark on each node so you can see if Kubernetes is securely deployed. |                                                              |
```
tmjd: For Image Assurance all these different components are included in the same row. Should we do the same for Compliance?

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+|                                | **compliance-reporter**<br />Handles report generation. Reads configuration history from Elasticsearch and determines time evolution of cluster-wide configuration, including relationships between policies, endpoints, services, and Network sets. Data is then passed through a zero-trust aggregator to determine the “worst-case outliers” in the reporting interval.compliance-controllerReads report configuration, and manages creation, deletion, and monitoring of report generation jobs. |                                                              |
+|                                | **compliance-server** <br />Provides the API for listing, downloading, and rendering reports, and RBAC by performing authentication and authorization through the Kubernetes API server. RBAC is determined from the users RBAC for the Global report type and Global report resources. |                                                              |
```
tmjd: compliance-server is not ran in the managed cluster so it can be removed from this table.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+|                                | **API and API proxy**<br />Isolates tenant data and ensures that all external access to Image Assurance data is through the API Proxy, which enforces authorization through Kubernetes RBAC. <br /><br />**Note** {{prodname}} does not store registry credentials in its database and does not pull customer images into the {{prodname}} control plane. {{prodname}} uses Google’s encryption at rest feature to encrypt data across the entire datastore. |                                                              |
```
tmjd: Did you mean to create a new row for these? I think this makes it unclear that they are part of Image Assurance.

tmjd: I think the sentence `{{prodname}} uses Google’s encryption at rest feature to encrypt data across the entire datastore.` should be removed. I think it is sharing too much.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+| Kubernetes API server          | A Kubernetes component that validates and configures data for the API objects (for example, pods, services, and others). | &bull; TCP 6443 (from all components)                        |
+| kube-controllers               | Monitors the Kubernetes API and performs actions based on cluster state. The {{prodname}} kube-controllers container includes these controllers:<br/><br />&bull; Node<br />&bull; Service<br />&bull; Federated services<br />&bull; Authorization<br />&bull; Elasticsearch configuration<br />&bull; Managed cluster | &bull;  TCP 9094 from Prometheus API service<br />&bull;  TCP 6443 to Kubernetes API server |
```
tmjd: We should remove `Elasticsearch configuration` and `Managed cluster` because they don't run in a managed cluster.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,51 @@
...
+| Prometheus API service         | Collects metrics from {{prodname}} components and makes the metrics available to Manager UI. | &bull;  TCP 6443 to Kubernetes API server<br />&bull;  TCP 9080 to Fluentd<br />&bull; TCP 9900 & 9081 to Prometheus API service |
+| calico-node                    | Bundles key components that are required for networking containers with {{prodname}}:<br/><br />&bull;  Felix<br />&bull;  BIRD<br />&bull;  confd | &bull; TCP 5473 to Typha<br />&bull;  TCP 9900 & 9081 from Prometheus API service |
+| Fluentd                        | Open-source data collector for unified logging. Collects and forwards {{prodname}} logs (flows, DNS, L7) to Elasticsearch. | &bull;  TCP 8080 to Guardian<br />&bull;  TCP 9080 from Prometheus API service |
```
tmjd: I think we should remove `Elasticsearch` and replace it with `log storage`.

tmjd (changes requested):
In several places `install/update` is used, is that a good practice? I know I use that regularly but it feels like shorthand that I don't know if it is appropriate for the docs. If the docs team is good with it then I am too.

One other general comment is the order of rows in the table, do we want to have some kind of order to them? Right now they seem in random order. If you're going to change the order,  please lets fix everything else first and after everything else is ok then we can change the order, otherwise follow up reviews will be difficult.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
stevegaossou: @bmckercher123, based a point that @tmjd made about focusing the documentation on what is in the customer's managed cluster only, you should reconsider the inclusion of the following components: 

- Pod watcher is going away in the next release (but even if it wasn't, it's not part of the managed cluster). 

- API and API proxy are not in the managed cluster. We definitely shouldn't mention API proxy, if we don't want to expose any details about the internals of the management cluster. We probably want to mention the API, but only at a high-level (no details about where it runs). I think we want to keep details on how we isolate tenant data hidden from users.  

- CLI scanner might be something we want to mention, but it also does not run in the managed cluster. Or more specifically, the user gets to decided where to run it, but it is not something that is assumed to be running in the managed cluster.

Also, the top of this Image Assurance section says: 

> Identifies vulnerabilities in container images that you deploy to Kubernetes clusters. **Includes these components:**

With that wording (the part in **bold**), it feels like its saying that is a complete list of the components. This might be minor, but I wonder if it would be clearer to say something like "The components of interest are ...".

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Fixed.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Agree that general language is better.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Agree. Storage for logs.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Good catch.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Deleted sentence.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Removed

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Removed.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Removed.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Fixed. Updated diagram to read "Log storage"

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Addressed all comments: deleted pod watcher, left API, deleted CLI scanner. Used "components of interest".

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Fixed.

tmjd (dismissed):
LGTM

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
```diff
@@ -0,0 +1,48 @@
...
+| Installation endpoints         | Endpoints at  `*.calicocloud.io` and `*.projectcalico.org`.  | TCP 443 for both                                             |
+| Intrusion detection controller | Handles integrations with threat intelligence feeds and {{prodname}} custom alerts. | &bull; TCP 8080 to Guardian<br />&bull; TCP 6443 to Kubernetes API server |
+| Image Assurance                | Identifies vulnerabilities in container images that you deploy to Kubernetes clusters.  Components of interest are: <br /><br />**Admission controller**<br />Uses Kubernetes Validating Webhook Configuration https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/ to control which images can be used to create pods based on scan results. <br /><br />**API**<br />Isolates tenant data and authorizes all external access to Image Assurance data. **Note:** {{prodname}} does not store registry credentials in its database and does not pull customer images into the {{prodname}} control plane.<br /> | &bull; TCP 8080 to Guardian<br /> &bull;  TCP 6443 to Kubernetes API server |
```
ctauchen: Do you really want the full link text in the doc?

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: Security teams are the audience for this document. They want to understand the architecture in the context of "how do you keep my cluster data safe" in SaaS product. So the first section addresses two components that keep clusters safe: the tunnel server and Guardian. The rest is the other ask, which is to show the components in the managed cluster and the communication via ports and protocol. I gave the initial draft to Support and customer to ensure this is what they wanted. We cannot provide details on the "management cluster" aka Calico Cloud part.

calico-cloud_versioned_docs/version-3.15/get-started/connect/cc-arch-diagram.mdx
bmckercher123: No, good catch! Removed.

Thakurvaibhav (dismissed):

stevegaossou:
@bmckercher123 it looks fine to me (from an Image Assurance perspective). 👍🏼

PR #385, opened 2023-02-21
DOCS-1277: Fix 404 Redirects

girishshankaran:
Fixes https://tigera.atlassian.net/browse/DOCS-1277

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -270,6 +270,25 @@
 /archive/v1.5/getting-started/kubernetes/installation/*          https://calico-legacy.netlify.app/archive/v1.5/getting-started/kubernetes/installation/:splat      200
 /archive/v1.5/getting-started/kubernetes/cloud-config/*          https://calico-legacy.netlify.app/archive/v1.5/getting-started/kubernetes/cloud-config/:splat      200
 /archive/v1.5/getting-started/kubernetes/tutorials/*             https://calico-legacy.netlify.app/archive/v1.5/getting-started/kubernetes/tutorials/:splat         200
+                                
+# Some specific pages are throwing 404 because of the docs restructuring
+/master/getting-started/kubernetes/minikube                      https://docs.tigera.io/calico/next/getting-started/kubernetes/minikube                                      301!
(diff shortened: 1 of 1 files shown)


girishshankaran:
@ctauchen. 

These paths are not resolving because of the fragment identifiers in the path, hence I did not add them. We may need to address them differently:

```
/reference/node/configuration#configuring-the-default-ip-pools
/reference/node/configuration#configuring-bgp-networking
/reference/node/configuration#configuring-bgp-networking
/reference/node/configuration#appendix
```

This path is already covered in the proxy rules, so I did not make an extra entry. However, this path is not working either.

`/master/getting-started/kubernetes/installation/*         https://calico-master.netlify.app/master/getting-started/kubernetes/installation/:splat      200`

_redirects
```diff
@@ -270,6 +270,24 @@
...
+/maintenance/ebpf/use-cases-ebpf                                 https://docs.tigera.io/calico/latest/operations/ebpf/use-cases-ebpf                                         301!
+/reference/contribute                                            https://docs.tigera.io/calico/latest/reference/involved                                                     301!
+/security/calico-network-policy                                  https://docs.tigera.io/calico/latest/network-policy/get-started/calico-policy/calico-network-policy         301!
+/maintenance/troubleshoot/vpp                                    https://docs.tigera.io/calico/latest/operations/troubleshoot/vpp                                            301!
+/getting-started/docker/tutorials/docker-swarm                   https://docs.tigera.io/calico/latest/getting-started                                                        301!
+/master/maintenance/monitor/monitor-component-visual             https://docs.tigera.io/calico/next/operations/monitor/monitor-component-visual                              301!
```
ctauchen, line 288: The list from our test included this same article for both latest and master. Can you add the same rule for latest? but without `/master`. Should be:
`/maintenance/monitor/monitor-component-visual https://docs.tigera.io/calico/latest/operations/monitor/monitor-component-visual 301`

ctauchen (commented):
One more to add, otherwise LGTM!

PR #392, opened 2023-02-22
Fixes image alignment

ctauchen:
Follow-up to https://github.com/tigera/docs/pull/325/


bmckercher123 (approved):
Approve

ctauchen:
@bmckercher123 Turns out the image problem occurred in two places in the file. I thought your fix hadn't worked because I was looking at the second. This PR fixes the second.

PR #394, opened 2023-02-22
batch scanning docs

ghost:


calico-cloud/image-assurance/scan-image-registries.mdx
```diff
@@ -153,6 +178,8 @@ Key names must match the full name of arguments passed to the CLI. The configura
 | --warn_threshold        | -w        | CC_WARN_THRESHOLD        | CVSS threshold for Warn scan results. Range from 0.0 - 10.0.                                                                |
 | --fail_threshold        | -f        | CC_FAIL_THRESHOLD        | CVSS threshold for Fail scan results. Range from 0.0 - 10.0.                                                                |
 | --vulnerability_db_path | -p        | CC_VULNERABILITY_DB_PATH | Path to a folder to store vulnerability data (defaults to `$XDG_CACHE_HOME`; if it is not set, defaults to `$HOME/.cache`). |
+| --input_file            | -i        | CC_INPUT_FILE            | Path to the JSON file containing image URLs.                                                                                           |
```
WilliamTigera, line 181: Nit: For the configuration file, it was unclear to me if I should have put int the option without the -- or the environment variable when editing the configuration yaml. If there's an easy way to differentiate which one to use for the configuration file, that might be worth a quick add.

WilliamTigera (dismissed):
Other than my nit, docs are pretty clear to me. I'm not sure whether others would agree or not with my comment.

calico-cloud/image-assurance/scan-image-registries.mdx
ghost: I agree it's a bit confusing. To keep the scope of this PR to the batch scanning, we can fix this in a separate PR.

PR #396, opened 2023-02-23
DOCS-1277: Missing images fix

girishshankaran:
Fixes https://tigera.atlassian.net/browse/DOCS-1279


girishshankaran:
@ctauchen Can you review this one?

This is what I did:

- Dropped "master" from the image path for v3.15 and v3.14
- We have used "master" image path in two other places in v3.16 and version next. Those paths are working because the images are there in the master folder, and those images were recently added after docs migration. I dropped "master" from these paths as well and then moved these two images to the `/img/calico-enterprise` path.

ctauchen (dismissed):
One thing: It looks like you've added the waf images to img/calico-enterprise, but you haven't removed them from the old location in img/calico-enterprise/master

Apart from that, LGTM!

PR #406, opened 2023-02-27
Update Calico cloud versioned to 315 images

tmjd:


doucol (approved):
It looks like a couple new components related to csi-driver and a rename of packetcapture-api to packetcapture, mostly - is that right?

LGTM

tmjd:
I copied the one from the enterprise v3.15 docs. I don't think I really care about the additions, it is mainly just the update of all the images to v3.15.1.

PR #407, opened 2023-02-27
Fix EE install-calico-windows.ps1 scripts

coutinhop:
The `install-calico-windows.ps1` script had this `layout: null` docs metadata in it which was preventing it from running properly.

Additionally, jekyll liquid markup was not correctly being substituted for the correct values.

Finally, there was a step missing in the Calico for Windows EE quickstart guides for some configuration options.


ctauchen (dismissed):
Lgtm

PR #410, opened 2023-02-27
CE-OCP-install-steps

bmckercher123:
### Fix missing steps in OCP install

- For CE OCP 3.14, 3.15, 3.16, 3.17


radTuti (commented):
for 3.14, it is still missing the step to install policies

PR #425, opened 2023-03-03
DOCS-1312: Fix Redirects

girishshankaran:


girishshankaran:
@doucol Can you review this one?

doucol (changes requested):
I think you will likely need to change all these redirects from 301 to 301!   -- this is because these paths exist in the source location which means you will need to "force" redirect on netlify (301!)

doucol (changes requested):
Same as other PR:

Do you think you could simplify all of these with a smaller set of directives? Such as:
/archive/* https://docs.tigera.io/archive/:splat 301!

ctauchen (approved):
So it's the same comments as I left in https://github.com/tigera/docs/pull/431. These need to be moved higher up, so they appear before the targeted 200 proxies for manifests.

Once that's done, and tests show these aren't appearing as 200s, then we're good to merge. LGTM

PR #444, opened 2023-03-09
Fix wrong use of deny when describing allow policy in secure metrics

sgaist:
The secure metrics documents  describe all Allow rules as deny.

This pull request fixes them.


ctauchen:
@sgaist This seems like a good fix. Thanks! Let me double-check a few things. I'll get back to you with any questions. Thanks again for the contribution!

ctauchen (approved):
LGTM, thanks!

PR #447, opened 2023-03-09
Fixes InstallOpenShift.js for CE 3.14

ctauchen:
To compliment https://github.com/tigera/docs/pull/410

### DIFF
--- a/calico-enterprise_versioned_docs/version-3.14/_includes/components/InstallOpenShift.js
+++ b/calico-enterprise_versioned_docs/version-3.14/_includes/components/InstallOpenShift.js
@@ -242,18 +242,31 @@ spec:
         When it shows all components with status <code>Available</code>, proceed to the next section.
       </p>
 
-      {props.clusterType == 'managed' && (
-        <>
-          <Heading
-            as='h4'
-            id={`sucure-${prodnamedash}-components-with-network-policy`}
-          >
-            Secure {prodname} components with network policy
-          </Heading>
-          <p>To secure the components that make up {prodname}, install the following set of network policies.</p>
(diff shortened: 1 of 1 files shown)


bmckercher123:
I noticed these things (as I work on the 3.16+ versions) of the same:
- We need to add an H4 for !==managed section or change existing h3 to h4 because maybe we don't need to account for legacy? (view in netlify)
- Update this line with "components" (to match h3) or the link doesn't work in the How to section.
<When condition={props.clusterType !== 'managed'}>
            <>
                <Heading
                    as='h4'
                    id={`secure-${prodnamedash}-components-with-network-policy`}

ctauchen:
> I noticed these things (as I work on the 3.16+ versions) of the same:
> 
> * We need to add an H4 for !==managed section or change existing h3 to h4 because maybe we don't need to account for legacy? (view in netlify)
> * Update this line with "components" (to match h3) or the link doesn't work in the How to section.
>   <When condition={props.clusterType !== 'managed'}>
>   <>
>   <Heading
>   as='h4'
>   id={`secure-${prodnamedash}-components-with-network-policy`}

Yes, the headings are a bit of a mess. Constructed based on the old source, which was always one h level lower that it should have been. I'd originally changed everything, but realized after that things were getting all messed up where the component was being reused in other sections.

I think we should address the heading business separately, as it may get very complicated. For now, we should just keep it consistent. Which means yes, I'll fix the h3 ==> h4. Long answer!

Yes, let's change the page ref.

ctauchen:
Closing this -- changes are to be added to https://github.com/tigera/docs/pull/410

PR #454, opened 2023-03-13
Eks

MichalFupso:


calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -113,12 +113,34 @@ Before you get started, make sure you have downloaded and configured the [necess
    kubectl apply -f {{manifestsUrl}}/manifests/calico-vxlan.yaml
    ```
 
-1. Configure {{prodname}} to disable AWS src/dst checks.
+2. Configure {{prodname}} to disable AWS src/dst checks.
```
mgleung: Do the numbers here need to be changed manually? I thought we always put `1` and liquid handles the explicit numbering for us?

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -113,12 +113,34 @@ Before you get started, make sure you have downloaded and configured the [necess
...
 
    ```batch
    kubectl -n kube-system set env daemonset/calico-node FELIX_AWSSRCDSTCHECK=Disable
    ```
 
+</TabItem>
+
+<TabItem label="Helm" value="Helm-2">
```
mgleung, line 135: I might be misunderstanding what this does, but I'm kind of confused why we have tabs here for helm when I don't see any other tabs. Are there other tabs that I'm missing?

mgleung: Nvm, I see that I misread things. This should be fine.

mgleung (dismissed):

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
bmckercher123: Although already introduced by previous author, using inline tabs inside a procedure has issues. The final step for all tabs comes out as "1." in netlify. I see two options. Duplicate the content for each tab (which is the original intent of tabs), or make the final step to be a non-numbered step. Like this:

Finally, add nodes to the cluster.

image

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -119,6 +119,28 @@ Before you get started, make sure you have downloaded and configured the [necess
...
+
+1. Add {{prodname}} into your Helm repository.
+
+    ```batch
+    helm repo add projectcalico https://docs.projectcalico.org/charts
+    ```
+
+1. If you already have {{prodname}} added, you can update it to get the latest released version.
```
bmckercher123:
```suggestion
1. If {{prodname}} is already added, update it to get the latest released version.
```

bmckercher123 (dismissed):
Minor nits.

bmckercher123 (approved):
LGTM

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
ctauchen: There are circumstances where they need to be explicit, yes. I'm not sure of the exact cause. But having multiple paragraphs for a single step, particularly with different markup, such as that for notes and codeblocks, significantly increases the likelihood that automatic numbering will fail.

ctauchen:
@MichalFupso If all comments have been addressed, then we're about ready to merge this one. 

Could you please squash your commits and rebase?

PR #463, opened 2023-03-15
Add SELinux documentation for main and v3.16

hjiawei:


hjiawei:
Calico Enterprise `v3.16` versioned docs are copies. I will update them once review comments are addressed on `main` docs.

calico-enterprise/_includes/components/ReqsSys.js
```diff
@@ -29,9 +29,11 @@ function NodeRequirementsEnt(props) {
           <ul>
             {(props.orch === orchestrators.Kubernetes || props.orch === orchestrators.HostProtection) && (
               <>
-                <li>CentOS 8</li>
```
hjiawei, line 32: [CentOS 8 is EoL'ed](https://www.centos.org/centos-linux-eol/). CentOS (stream) in general is just one of the RHEL related or derived OSes. I would like to loosely categorize it into the RHEL "binary-compatible operating systems".

ctauchen: Thanks -- good to get this reference out.

I don't know that we can stand behind a "loose category". Shouldn't the rule be that we list support only for systems we've tested? I'd much prefer a long list.

Can you give an example of one of these binary-compatible systems?

calico-enterprise/operations/logstorage/create-storage.mdx
```diff
@@ -64,9 +64,15 @@ volumeBindingMode: WaitForFirstConsumer
 reclaimPolicy: Retain
 ```
 
+:::note
+
+If local persistent volumes are provisioned on a SELinux-enabled host, you can use `/mnt/tigera` host path created by {{prodname}} policy package.
```
ctauchen, line 69:
```suggestion
If local persistent volumes are provisioned on an SELinux-enabled host, you can use the `/mnt/tigera` host path created by {{prodname}} policy package.
```

calico-enterprise/_includes/components/ReqsSys.js
```diff
@@ -29,9 +29,11 @@ function NodeRequirementsEnt(props) {
...
+                <li>Red Hat Enterprise Linux 7 and 8 or binary-compatible operating systems</li>
+                  <ul>
```
ctauchen, line 33: I think we should remove this reference. SELinux isn't a requirement, but rather a special installation option for RHEL.

calico-enterprise/operations/selinux.mdx
```diff
@@ -0,0 +1,42 @@
+---
```
ctauchen, line 1: I think this would be better placed under `getting-started/install-on-clusters/`. 

It's best to get this information directly to users as part of the install process. I'd suggest we add this as either an optional step or a prerequisite to every relevant installation procedure.

ctauchen: I'll hold specific edits on this section until after we've settled on a location.

ctauchen (commented):
@hjiawei Thanks for this. A few comments and questions for you.

calico-enterprise/_includes/components/ReqsSys.js
hjiawei: SELinux is enabled by default on RHEL and binary-compatible systems. It is not a special installation.

calico-enterprise/_includes/components/ReqsSys.js
hjiawei: [RockyLinux](https://rockylinux.org/) and [AlmaLinux](https://almalinux.org/) are the two main ones to replace CentOS 8.

ctauchen:
Closing this PR, re-opening here: https://github.com/tigera/docs/pull/500

PR #464, opened 2023-03-15
Add env CALICO_IPV6POOL_NAT_OUTGOING to calico-node for IPv6

carezkh:
Releated PR: https://github.com/projectcalico/calico/pull/7268

### DIFF
--- a/calico/networking/ipam/ipv6.mdx
+++ b/calico/networking/ipam/ipv6.mdx
@@ -117,6 +117,14 @@ spec:
    | `IP6`               | `autodetect` |
    | `FELIX_IPV6SUPPORT` | `true`       |
 
+   :::note
+
+   If your IPv6 IP pools include private IP addresses, pods that are assigned private IP addresses cannot perform outbound NAT by default. 
+
+   To enable outbound NAT for these pods, add `CALICO_IPV6POOL_NAT_OUTGOING: true` to the environment of the `calico-node` container.
+
+   :::
+
 1. For clusters **not** provisioned with kubeadm (see note below), configure the default IPv6 IP pool by adding the following variable setting to the environment for the `calico-node` container:
 
    | Variable name          | Value                                                                                                   |
@@ -199,6 +207,14 @@ spec:
    | ------------------- | ------------ |
    | `IP6`               | `autodetect` |
    | `FELIX_IPV6SUPPORT` | `true`       |
+
+   :::note
+
+   If your IPv6 IP pools include private IP addresses, pods that are assigned private IP addresses cannot perform outbound NAT by default. 
+
+   To enable outbound NAT for these pods, add `CALICO_IPV6POOL_NAT_OUTGOING: true` to the environment of the `calico-node` container.
+
+   :::
 
 1. For clusters **not** provisioned with kubeadm (see note below), configure the default IPv6 IP pool by adding the following variable setting to the environment for the `calico-node` container:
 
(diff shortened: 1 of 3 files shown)


ctauchen:
@carezkh Thanks for this. We'll add this to the review queue and return to you with questions.

calico/networking/ipam/ipv6.mdx
```diff
@@ -117,6 +117,12 @@ spec:
    | `IP6`               | `autodetect` |
    | `FELIX_IPV6SUPPORT` | `true`       |
 
+   :::note
+
+   For IPv6 IP Pools with private IPs, the environment `CALICO_IPV6POOL_NAT_OUTGOING: true` should be set to perform outbound NAT for connections from pods to outside of the cluster.
```
ctauchen: I want to make sure I understand this correctly. 

You're trying to document how to configure an IPv6-only cluster that uses only private IPs for its IP pools. 

To do this, users must add an environment variable to the `calico-node` container: `CALICO_IPV6POOL_NAT_OUTGOING: true`.

Is that right?

ctauchen: Some other questions:
1. Why would someone want to do this?
2. Why would someone NOT want to do this?
3. How important is it for this to be set here (during this step?)

ctauchen (commented):
A few questions for you @carezkh. Thanks again for sending this in!

ctauchen:
@caseydavenport @mgleung I see you had questions for the submitter in https://github.com/projectcalico/calico/pull/7268. Are you satisfied with how they've been addressed?

calico/networking/ipam/ipv6.mdx
carezkh: @ctauchen Thanks for your review! 

> You're trying to document how to configure an IPv6-only cluster that uses only private IPs for its IP pools.

The cluster here can be either IPv6-only or dual-stack cluster.

> Why would someone want to do this?

For pods with private IPv6 address (also named IPv6 Unique Local Address), They can not communicate with the outside world over the private IPv6 address when there is no SNAT rule on host. Please see this issue (https://github.com/projectcalico/calico/issues/7183) for more details. 

> Why would someone NOT want to do this?

There is no need to do this for the few who does not want to communicate with the outside world over the private IPv6 address in pods.

> How important is it for this to be set here (during this step?)

IIUC, the outgoing NAT support can be added by IPPool resource after CNI calico is installed.

But there is a describe in `manifest tab` tells us the pods can communicate with the outside cluster over IPv6, actually they can not.
https://github.com/tigera/docs/blob/0607542585638af863fc60688e394fe050882ced/calico/networking/ipam/ipv6.mdx?plain=1#L132-L135

And the field `natOutgoing` is `Enable` in `operator tab`,  so i guess maybe someone forgot to update the `manifest tab`.

caseydavenport (approved):

caseydavenport:
@ctauchen this looks good to me - thanks for the ping, and thanks @carezkh for sticking with this one :)

calico/networking/ipam/ipv6.mdx
ctauchen: Ok, thanks for that detailed information. Let suggest a slightly different wording:

```
:::note

If your IPv6 IP pools include private IP addresses, pods that are assigned private IP addresses cannot perform outbound NAT by default. 

To enable outbound NAT for these pods, add `CALICO_IPV6POOL_NAT_OUTGOING: true` to the environment of the `calico-node` container.

:::
```

carezkh: done.

PR #491, opened 2023-03-21
DOCS-fix-TOC

bmckercher123:
Fix install CE TOC to before migration with abbreviations. 

Product Version(s):
CE 3.16, 3.17

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:


ctauchen (commented):
We mentioned something like this a while back. But as I recall, the idea was to change just the nav title, not the actual title.

What do you think about making these changes in the frontmatter? Something like:
`title: AKS`. That way, the sidebars change, but the title (and head metadata) remain the same.

bmckercher123:
Closing this out. Will start a new one that fixes both CE and OSS.

PR #512, opened 2023-03-27
Fix a redirect

girishshankaran:
Doug's request on docops channel.

this URL is broken - can we fix the redirect for this?  https://docs.tigera.io/calico/latest/calico-enterprise/  (it is coming from this origin:  https://docs.projectcalico.org/calico-enterprise/ )

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -505,6 +505,8 @@
 /v1.6/* https://docs.tigera.io/archive/v1.6/:splat 301
 /v1.5/* https://docs.tigera.io/archive/v1.5/:splat 301
 
+/calico-enterprise/  https://docs.tigera.io/calico-enterprise/latest/about-calico-enterprise 301
+
 # Keeping felix alive and well for all the old READMEs.
 /images/felix.png /images/felix.png 200
 


_redirects
```diff
@@ -505,6 +505,9 @@
 /v1.6/* https://docs.tigera.io/archive/v1.6/:splat 301
 /v1.5/* https://docs.tigera.io/archive/v1.5/:splat 301
 
+# Redirect to the root folder of Calico Eneterprise
```
ctauchen: We can leave out the comment here. It's actually just a directory that grouped information about CE, but in the OS docs.

_redirects
```diff
@@ -505,6 +505,9 @@
...
+/calico-enterprise/ https://docs.tigera.io/calico-enterprise/ 301
```
ctauchen:
```suggestion
/calico-enterprise/ https://docs.tigera.io/calico-enterprise/latest/about-calico-enterprise 301
```
Always helpful to be specific and save an extra redirect.

ctauchen (approved):
Two small nits, otherwise LGTM

_redirects
girishshankaran: Done

PR #513, opened 2023-03-27
Mention ordering with Kubernetes NetworkPolicies

remram44:
Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

The documentation for NetworkPolicies mention the `order` field and give example of its use, but doesn't say how those policies are ordered relative to Kubernetes NetworkPolicies. I was able to find that information on StackOverflow, but I believe it should definitely be mentioned in your documentation.

Source: https://stackoverflow.com/a/68134768/711380

### DIFF
--- a/calico_versioned_docs/version-3.25/network-policy/get-started/calico-policy/calico-network-policy.mdx
+++ b/calico_versioned_docs/version-3.25/network-policy/get-started/calico-policy/calico-network-policy.mdx
@@ -217,7 +217,7 @@ spec:
 
 ### Apply network policies in specific order
 
-To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint.
+To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint. Because Kubernetes NetworkPolicy don't have an explicit order field, they are treated as though they had an implicit order of 1000.
 
 In the following example, the policy **allow-cluster-internal-ingress** (order: 10) will be applied before the **policy drop-other-ingress** (order: 20).
 
--- a/calico_versioned_docs/version-3.25/network-policy/get-started/calico-policy/calico-network-policy.mdx
+++ b/calico_versioned_docs/version-3.25/network-policy/get-started/calico-policy/calico-network-policy.mdx
@@ -217,7 +217,7 @@ spec:
 
 ### Apply network policies in specific order
 
-To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint. Because Kubernetes NetworkPolicy don't have an explicit order field, they are treated as though they had an implicit order of 1000.
+To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint. Because Kubernetes NetworkPolicies don't have an explicit order field, they are treated as though they had an implicit order of 1000.
 
 In the following example, the policy **allow-cluster-internal-ingress** (order: 10) will be applied before the **policy drop-other-ingress** (order: 20).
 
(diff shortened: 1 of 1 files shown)


remram44:
Do I need to sign this CLA? It is very unclear what I am agreeing to:

> grant to (...) recipients of software distributed by the Project Manager a (...) copyright license to reproduce, (...), sublicense, and distribute Your Contributions and such derivative works

The top of the page says "agree to the CLA for multiple repositories or organizations"

Who is the Project Manager? Which organizations and repositories does this apply to?

remram44:
> You agree to notify the Project Manager

How would I do this?

ctauchen:
@tigera/core-maintainers Can anyone help answer @remram44's questions above?

I can confirm this following:
* Yes, signing the CLA is a requirement for any submission to be accepted. 
* From the document itself, I can see that the "Project Manager" is defined as "Tigera, Inc".

ctauchen:
@nelljerram Are you happy with this change? Seems to have come from [your own answer on Stack Overflow](https://stackoverflow.com/a/68134768/711380).

remram44:
Does this need a CLA? The information comes from your team anyway.

nelljerram:
@ctauchen @remram44 Yes, looks good to me, except I would change "don't" to "doesn't", considering "NetworkPolicy" in this context to be an [uncountable noun](https://en.wikipedia.org/wiki/Mass_noun).

calico_versioned_docs/version-3.25/network-policy/get-started/calico-policy/calico-network-policy.mdx
```diff
@@ -217,7 +217,7 @@ spec:
 
 ### Apply network policies in specific order
 
-To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint.
+To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint. Because Kubernetes NetworkPolicies don't have an explicit order field, they are treated as though they had an implicit order of 1000.
```
ctauchen:
```suggestion
To control the order/sequence of applying network policies, you can use the **order** field (with precedence from the lowest value to highest). Defining policy **order** is important when you include both **action: allow** and **action: deny** rules that may apply to the same endpoint. Because Kubernetes network policies don't have an explicit order field, they are treated as though they had an implicit order of 1000.
```

ctauchen:
> @ctauchen @remram44 Yes, looks good to me, except I would change "don't" to "doesn't", considering "NetworkPolicy" in this context to be an [uncountable noun](https://en.wikipedia.org/wiki/Mass_noun).

I'm not sure what I'd do with a bucket of NetworkPolicy! 

@remram44 You're on the right track, but let's just call it "network policies" rather than make the field name plural.

@nelljerram Any chance you can offer insight into the CLA question above?

nelljerram:
@ctauchen I believe our CLA policy is strict.  @remram44 will remain credited as the contributor of this change, regardless of where the information originated from, and hence the CLA is needed.

Independently, if there are issues of clarity with the CLA process, that is an important matter that should be raised and addressed.

remram44:
Signing that I agree to "notify the Project Manager" without giving me a clear channel to do so is a non-starter. Especially if you "may freely [...] assign or transfer this Agreement". The top of the page already indicates that this agreement applies to multiple repositories or organizations without listing them, so I have no idea what I'm signing.

ctauchen:
> Signing that I agree to "notify the Project Manager" without giving me a clear channel to do so is a non-starter. Especially if you "may freely [...] assign or transfer this Agreement". The top of the page already indicates that this agreement applies to multiple repositories or organizations without listing them, so I have no idea what I'm signing.

I'm not a lawyer, but it seems to me that it's simple enough to find the project manager and get in touch. Is it standard for these agreements to get into granular detail? I don't know.

@remram44 If your aim is to modify the CLA, rather than to get information about it, I suggest you [open an issue with projectcalico/calico](https://github.com/projectcalico/calico/issues). Modifying this CLA is not something we can help with directly.

remram44:
My aim was just to get this one sentence in your docs.

It seems that your are powerless to explain or fix this contract that you require me to sign, so I don't see a way forward. It is unfortunate that your process blocks even the simplest of contributions.

PR #543, opened 2023-04-05
Correct EKS getting started docs

lwr20:
EKS getting started docs are wrong:
- the Helm install doesn't mention needing to create and specify a namespace
- nodegroup can be added once aws-node daemonset is deleted, and this means the Helm instructions won't fail at the `helm install` step.

Product Version(s):

Calico v3.24 upwards.

Issue:

Raised by CU Byron Mansfield in slack thread: https://calicousers.slack.com/archives/CPTH1KS00/p1680705961183599

Link to docs preview:

SME review:

DOCS review:

Additional information:

### DIFF
--- a/calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
+++ b/calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
@@ -70,10 +70,16 @@ Before you get started, make sure you have downloaded and configured the [necess
    eksctl create cluster --name my-calico-cluster --without-nodegroup
    ```
 
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
+   ```
+
+1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
+
+   ```batch
+   eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
    ```
 
 1. Now that you have a cluster configured, you can install {{prodname}}.
@@ -104,12 +110,6 @@ Before you get started, make sure you have downloaded and configured the [necess
    EOF
    ```
 
-1. Finally, add nodes to the cluster.
-
-   ```batch
-   eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
-   ```
-
 </TabItem>
 <TabItem label="Manifest" value="Manifest-1">
 
@@ -123,12 +123,6 @@ Before you get started, make sure you have downloaded and configured the [necess
 
    ```bash
    kubectl -n kube-system set env daemonset/calico-node FELIX_AWSSRCDSTCHECK=Disable
-   ```
-
-1. Finally, add nodes to the cluster.
-
-   ```bash
-   eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
    ```
 
 </TabItem>
@@ -146,16 +140,15 @@ Before you get started, make sure you have downloaded and configured the [necess
     helm repo update
     ```
 
+1. Create the tigera-operator namespace.
+    ```batch
+    kubectl create namespace tigera-operator
+    ```
+
 1. Install version {{releaseTitle}} of the {{prodname}} operator and custom resource definitions.
 
     ```batch
-    helm install calico projectcalico/tigera-operator --version {{releaseTitle}}
-    ```
-
-1. Finally, add nodes to the cluster.
-
-    ```batch
-    eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
+    helm install calico projectcalico/tigera-operator --namespace tigera-operator --version {{releaseTitle}}
     ```
 
 </TabItem>
--- a/calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
+++ b/calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
@@ -70,15 +70,15 @@ Before you get started, make sure you have downloaded and configured the [necess
    eksctl create cluster --name my-calico-cluster --without-nodegroup
    ```
 
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing a conflicting CNI config to your nodes.
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
    ```
 
-1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
+1. Now that the `aws-node` daemonset is gone, you can add nodes to the cluster.
 
-   ```batch
+   ```bash
    eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
    ```
 
@@ -145,7 +145,7 @@ Before you get started, make sure you have downloaded and configured the [necess
     kubectl create namespace tigera-operator
     ```
 
-1. Install version {{releaseTitle}} of the {{prodname}} operator and custom resource definitions.
+1. Install version {{releaseTitle}} of the tigera-operator and custom resource definitions.
 
     ```batch
     helm install calico projectcalico/tigera-operator --namespace tigera-operator --version {{releaseTitle}}
--- a/calico_versioned_docs/version-3.24/getting-started/kubernetes/managed-public-cloud/eks.mdx
+++ b/calico_versioned_docs/version-3.24/getting-started/kubernetes/managed-public-cloud/eks.mdx
@@ -60,10 +60,16 @@ Before you get started, make sure you have downloaded and configured the [necess
    eksctl create cluster --name my-calico-cluster --without-nodegroup
    ```
 
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
+   ```
+
+1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
+
+   ```batch
+   eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
    ```
 
 1. Now that you have a cluster configured, you can install {{prodname}}.
@@ -94,12 +100,6 @@ Before you get started, make sure you have downloaded and configured the [necess
    EOF
    ```
 
-1. Finally, add nodes to the cluster.
-
-   ```batch
-   eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
-   ```
-
 </TabItem>
 <TabItem label="Manifest" value="Manifest-1">
 
@@ -113,12 +113,6 @@ Before you get started, make sure you have downloaded and configured the [necess
 
    ```bash
    kubectl -n kube-system set env daemonset/calico-node FELIX_AWSSRCDSTCHECK=Disable
-   ```
-
-1. Finally, add nodes to the cluster.
-
-   ```bash
-   eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
    ```
 
 </TabItem>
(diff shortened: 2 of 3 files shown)


girishshankaran (approved):
LGTM

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -146,16 +140,15 @@ Before you get started, make sure you have downloaded and configured the [necess
...
-    helm install calico projectcalico/tigera-operator --version {{releaseTitle}}
+    kubectl create namespace tigera-operator
     ```
 
-1. Finally, add nodes to the cluster.
+1. Install version {{releaseTitle}} of the {{prodname}} operator and custom resource definitions.
 
     ```batch
-    eksctl create nodegroup --cluster my-calico-cluster --node-type t3.medium --max-pods-per-node 100
+    helm install calico projectcalico/tigera-operator --namespace tigera-operator --version {{releaseTitle}}
```
byronmansfield, line 151: If this is eks specific. And it is for using tigera-operator helm chart. I believe the values file [here](https://github.com/projectcalico/calico/blob/master/charts/tigera-operator/values.yaml#L5) might want to be set to `"EKS"`. In which case you would need to pass in the `--set installation.kubernetesProvider="EKS"` argument to the helm install command.

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -70,12 +70,18 @@ Before you get started, make sure you have downloaded and configured the [necess
...
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
    ```
 
+1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
+
+   ```batch
```
ctauchen:
```suggestion
   ```bash
```
We're back to 'bash' again.

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -70,12 +70,18 @@ Before you get started, make sure you have downloaded and configured the [necess
    eksctl create cluster --name my-calico-cluster --without-nodegroup
    ```
 
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
```
ctauchen:
```suggestion
1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing a conflicting CNI config to your nodes.
```

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -70,12 +70,18 @@ Before you get started, make sure you have downloaded and configured the [necess
...
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
    ```
 
+1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
```
ctauchen:
```suggestion
1. Now that the `aws-node` daemonset is gone, you can add nodes to the cluster.
```

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -146,16 +140,15 @@ Before you get started, make sure you have downloaded and configured the [necess
...
-1. Install version {{releaseTitle}} of the {{prodname}} operator and custom resource definitions.
-
+1. Create the tigera-operator namespace.
     ```batch
-    helm install calico projectcalico/tigera-operator --version {{releaseTitle}}
+    kubectl create namespace tigera-operator
     ```
 
-1. Finally, add nodes to the cluster.
+1. Install version {{releaseTitle}} of the {{prodname}} operator and custom resource definitions.
```
ctauchen: `releaseTitle` returns the current version of Calico. If you're looking for the current version of the operator, you'll need `tigeraOperatorVersionShort` or `tigeraOperator`. See variables.js and releases.json in the product/version root directory.

ctauchen: the {{prodname}} operator

Isn't it still called the Tigera operator?

ctauchen (approved):
A few comments, which should be carried through globally.

Is there anyone you'd like to nominate for a technical review?

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
lwr20: According to https://github.com/tigera/operator/blob/master/pkg/controller/installation/core_controller.go#L499
installation.kubernetesProvider="EKS" causes tigera-operator to:
- pick AmazonVPC CNI
- specify default IP pool of "172.16.0.0/16" + VXLAN
- disable BGP
- set autodetection to CanReach: "8.8.8.8"

Since the point of this section of the doc is to set up Calico CNI, not AmazonVPC CNI, I think we do NOT want `kubernetesProvider="EKS"`

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
lwr20: Not sure I agree with this one.  In my head "CNI config" is a collective noun, but not sure its worth arguing :)

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
lwr20:
```suggestion
1. Install version {{releaseTitle}} of the tigera-operator and custom resource definitions.
```

calico_versioned_docs/version-3.24/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -60,12 +60,18 @@ Before you get started, make sure you have downloaded and configured the [necess
...
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
    ```
 
+1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
+
+   ```batch
```
lwr20:
```suggestion
   ```bash
```

calico_versioned_docs/version-3.25/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -70,12 +70,18 @@ Before you get started, make sure you have downloaded and configured the [necess
...
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
+1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemonset to prevent it from writing conflicting CNI config to your nodes.
 
    ```bash
    kubectl delete daemonset -n kube-system aws-node
    ```
 
+1. Now that the aws-node daemonset is gone, you can add nodes to the cluster.
+
+   ```batch
```
lwr20:
```suggestion
   ```bash
```

lwr20:
> Is there anyone you'd like to nominate for a technical review?

Perhaps @caseydavenport?  I have a recollection of Casey adding the namespace and helm installing into that namespace before.

caseydavenport:
per @lwr20 's request, there are a few other things we need to do for our helm docs across the board:

Our generic helm docs [here](https://docs.tigera.io/calico/latest/getting-started/kubernetes/helm) have some content that is just wrong.

> If you are installing on a cluster installed by EKS, GKE, AKS or Mirantis Kubernetes Engine (MKE), or you need to customize TLS certificates, you must customize this Helm chart by creating a values.yaml file. Otherwise, you can skip this step.

This is misleading and both too specific and too general. It should be re-written to be more precise and explain that for basic setups, default settings will just work. However there are many many reasons you might want to customize your settings. I'm not sure why it specifically calls out TLS as a reason there - it's one reason, but the implication by omission is that it's the only reason, which is wrong!

Our helm upgrade docs are also wrong - they omit the step where you need to manually replace the CRDs prior to upgrade.

We can handle those in this PR, or in separate follow-ons depending on how you want to do it. But it would be great to attack these while we're thinking about helm.

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
caseydavenport: It will only pick those settings if no other settings are provided - they are the deafults. Even if you set the provider to empty, the operator will know it is running on EKS and default that field to EKS anyway, and thus default to AmazonVPC CNI.

To use Calico on EKS, you will need to explicitly tell it to use Calico CNI as part of your Installation (via values.yaml, or explicit CLI flags to set particular values).

caseydavenport: I think we will want to create a values.yaml file and pass it to helm, with something like this:

```yaml
cat > values.yaml <<EOF
installation:
  kubernetesProvider: EKS
  cni:
    type: Calico
  calicoNetwork:
    bgp: Disabled
EOF
```

Potentially also include IP pools, if we need to:

```yaml
    ipPools:
    - cidr: <default cidr>
      encapsulation: VXLAN
```

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
ctauchen: Yikes, that's way too cool for me. I'm reading config = configuration, which is the whole thing. All or nothing.

You're saying that this usage is collective, meaning we can write some config to the nodes, meaning we can write/rewrite _some discrete portion of the total configuration_ (more than 0 but less than all)?

lwr20: Yeah, the CNI config directory `/etc/cni/net.d/` on the host gets config files written to it by various CNI plugins.  

In this (error) case there will be 2 config files - one from aws-node, and one from calico-node.  Sadly, the tiebreaker for which one to use is alphabetical, which calico loses to aws.  https://github.com/containernetworking/cni/blob/main/SPEC.md

bmckercher123:
@https://github.com/byronmansfield Looks like code change that the docs PR is for still hasn't been merged yet - it's failing tests. We had someone restart it. If you can rebase, we can get this merged.

ctauchen:
@lwr20 I'm trying to clear out stale PRs. This one seems both stale and necessary. I get the impression that the changes are sound, but need to be updated to later versions. Casey's comments may have slowed folks down, but the general Helm changes can be considered separately. 

Have I got the right idea here? I'm happy to put up a new PR with these changes in the later versions. And put something in the backlog for the other Helm business.

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
```diff
@@ -70,12 +70,18 @@ Before you get started, make sure you have downloaded and configured the [necess
    eksctl create cluster --name my-calico-cluster --without-nodegroup
    ```
 
-1. Since this cluster will use {{prodname}} for networking, you must delete the `aws-node` daemon set to disable AWS VPC networking for pods.
```
caseydavenport, line 73: I don't understand why the new wording is more desirable - the former was more abstractly correct, whereas the latter is one particular reason among several for why we don't want the aws-node daemonset.

caseydavenport: Maybe "to ensure that nodes are not configured with AWS VPC networking for pods" would be better than "to disable"?

calico/getting-started/kubernetes/managed-public-cloud/eks.mdx
caseydavenport: @ctauchen this particular bit of feedback needs to be addressed before merge. I think it's the only outstanding technical feedback.

I think we need the following as part of this step in order to install correctly.:

---

1. Create a values.yaml file to configure Calico as your network provider.

```
cat > values.yaml <<EOF
installation:
  kubernetesProvider: EKS
  cni:
    type: Calico
  calicoNetwork:
    bgp: Disabled
    ipPools: 
    - encapsulation: VXLAN
      cidr: 192.168.0.0/16
EOF
```

1. Install calico using helm

```
helm install calico projectcalico/tigera-operator --namespace tigera-operator --version {{releaseTitle}} -f values.yaml
```

ctauchen:
I'm closing this old PR. Some of the suggestions here have been made already. I'll be working through OSS install docs this quarter and plan to review an incorporate changes suggested here as part of [DOCS-2366](https://tigera.atlassian.net/browse/DOCS-2366).

[DOCS-2366]: https://tigera.atlassian.net/browse/DOCS-2366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #608, opened 2023-05-01
Update: Docs banner update

frozenprocess:
Calico survey banner update.


doucol (approved):

doucol:
@ctauchen let's go forward with this - at least until we have eBPF better represented in the home page. Thanks

PR #616, opened 2023-05-02
Cloud: Add make target to update private image list

tmjd:
This PR adds a make target to update the private image list from the clouddownloadurl when it is updated.

This PR also includes CI to ensure that if the image list is not updated a PR will fail CI.

Product Version(s):
Calico Cloud

Issue:
The Calico Cloud docs include a list of images needed to setup a private registry. Before the unified docs the list of images were included in the docs by fetching the image-list from the cloudversion that was specified. That functionality was dropped when the unified docs were introduced, which left room for the image list to become out of date because of forgetting to update the image list manually.

Link to docs preview:

SME review:

DOCS review:

Additional information:


tmjd:
This PR includes an update to the v3.15 CC version, which shouldn't matter since we don't use it anymore.
I think we should probably remove the `version-3.15` docs but I didn't know if we're keeping it for some reason.

PR #623, opened 2023-05-04
Fix rbac version - EV-3491

vara2504:
Product Version(s):

Issue:

Updating the missed files with rbac version.

https://tigera.atlassian.net/browse/EV-3491

SME review:

DOCS review:

Additional information:

Merge checklist:

- [x] Tests have passed

### DIFF
--- a/calico-cloud/compliance/overview.mdx
+++ b/calico-cloud/compliance/overview.mdx
@@ -97,7 +97,7 @@ The following sample manifest creates RBAC for three users: Paul, Candice and Da
 
 ```yaml
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tigera-compliance-manage-report-config
 rules:
@@ -110,7 +110,7 @@ rules:
 
 ---
 kind: ClusterRoleBinding
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tigera-compliance-manage-report-config
 subjects:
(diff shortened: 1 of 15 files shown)


ctauchen:
@vara2504 The bug's fixVersion is 3.16.2. Is that because this docs change is relevant only for that patch? If so, then we'll need to be careful about timing for how we merge these changes. But it kinda looks like this is just a bug that ought to be changed immediately.
CC @bmckercher123

vara2504:
@ctauchen :  it is applicable for both next and 3.16.2 .. 
The fix was made in calico-private manifest -https://github.com/tigera/calico-private/pull/6044 .. RBAC version is updated from v1betav1 to v1.  Docs had few yaml blocks with cluster role binding which is missed at that point of time.

initial fix - https://github.com/tigera/calico-private/pull/6044 - is available in both next and 3.16.2

bmckercher123:
3.16.2 will not release until end of May. Hold this PR until then.

vara2504:
@bmckercher123 Sorry i conveyed it wrong.. This change is applicable for both 3.16 and 3 17.. it is a bug in the existing docs which need to be fixed.. 
RBAC version should be` rbac.authorization.k8s.io/v1` instead of `apiVersion: rbac.authorization.k8s.io/v1beta1` 
For example: 3.16 federation yaml have a correct version.
https://docs.tigera.io/calico-enterprise/latest/multicluster/federation/kubeconfig#create-access-to-secrets-for-clusters

ctauchen:
@vara2504 So it sounds like all instances of `rbac.authorization.k8s.io/v1beta1` must be changed to `rbac.authorization.k8s.io/v1`.

This applies to 
* CE 3.16
* CE 3.17
* CC/3.16
* CC/3.17

If that's the case, then let's make sure those changes are made throughout those versions. The current changes do not capture everything in the relevant versions.

For example: https://deploy-preview-623--tigera.netlify.app/calico-cloud/compliance/overview#configure-report-permissions

vara2504:
@ctauchen yes you are correct.. Sorry for missing it earlier.. i will update all the needed files now. Also update 3.15 files as @rene have made the changes for 3.15 manifest -https://github.com/tigera/calico-private/pull/6046

bmckercher123 (approved):
LGTM.

PR #654, opened 2023-05-16
document FeatureGates and BPFConnectTimeLoadBalancingWorkaround

tomastigera:
Product Version(s):
CE and OSS next 

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [x] Test have passed


bmckercher123 (approved):
I found this explanation hard to follow...is this correct?

* `BPFConnectTimeLoadBalancingWorkaround` 

By default, this parameter is disabled. However, if host-networked processes cannot reach services, use the `enable` flag to fix this. When CTLB is enabled, UDP clients may get stuck sending traffic to endpoints that no longer exist. In this case, disable CTLB using the `udp` flag.  
	* `enabled` - enables CTLB so host-networked processes can reach services
	* `udp` - disables CTLB  for UDP, but still ensures that services are always accessible. The preferred solution is to make sure that DNS works.

bmckercher123 (commented):
As the PR stands, the fix is only for CE 3.17 (next) and OSS 3.26 (next). If relevant for CC, make the same changes in CC root (3.16). If the fix applies to previous versions, go to the versioned folders and make the same changes there.

calico-enterprise/reference/component-resources/node/felix/configuration.mdx
```diff
@@ -100,6 +100,14 @@ The full list of parameters which can be set is as follows.
...
+| `FeatureGates`			| `FELIX_FEATUREGATES`			| Is used to enable or disable tech-preview Calico features. Values are specified in a comma separated list with no spaces. This is used to enable features that are not fully production ready. Example; "AbC=enabled,XyZ=false" [Default: not set] | string
+
+#### Feature Gates
+
+* `BPFConnectTimeLoadBalancingWorkaround` - Use when connect-time loadblalancer (CTLB) is turned off or if you want to turn it of for UDP only. When CTLB is turned off, host networked processes cannot always reach services. This workaround makes sure that they can. When CTLB is turned on, UDP clients may get stuck sending traffic to endpoint that does not exist anymore. So CTLB needs to be turned off sometimes.
```
bmckercher123:
```suggestion
* `BPFConnectTimeLoadBalancingWorkaround` - Use when connect-time loadblalancer (CTLB) is turned off or if you want to turn it off for UDP only. When CTLB is turned off, host networked processes cannot always reach services. This workaround makes sure that they can. When CTLB is turned on, UDP clients may get stuck sending traffic to endpoint that does not exist anymore. So CTLB needs to be turned off sometimes.
```

bmckercher123 (approved):
Just a typo.

PR #679, opened 2023-05-24
Operator based calico is necessary for upgrading to EE

rene-dekker:
https://tigera.atlassian.net/browse/EV-3717


bmckercher123 (approved):
LGTM

PR #696, opened 2023-05-31
OSS operator migration

MichalFupso:
Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


bmckercher123 (approved):
LGTM

calico_versioned_docs/version-3.26/operations/operator-migration.mdx
```diff
@@ -42,6 +42,7 @@ and calculate how to take ownership of them. The operator will maintain existing
 ## Before you begin
 
 - Ensure that your {{prodname}} installation is configured to use the Kubernetes datastore. If your cluster uses etcdv3 directly, you must follow [the datastore migration procedure](datastore-migration.mdx) before following this document.
+- Migration to operator-managed installation is supported only between the same {{prodname}} versions
```
ctauchen: I find this statement confusing. Migrating between the same version? Can we be more specific about what this means? It might be clearer if we also say explicitly what you can't do.

I suspect what's going on here is that we have an existing installation that does not use the operator. We want to start using the operator, which requires installing the operator with tigera-operator.yaml. The procedure gives a link to a version-specific file. So perhaps this line is telling us we can't use a different tigera-operator.yaml?

bmckercher123: @MichalFupso Is this the gist? You can migrate to operator-based installation as long as you are migrating within the same product version you are currently in. For example, if you are using 3.24, you can migrate to operator-managed installation in any version of 3.24, but you cannot migrate directly to 3.25 and use operator unless you first upgrade?

bmckercher123: Changes should be copied to 3.27 (root folder) as well.

bmckercher123:
New PR: https://github.com/tigera/docs/pull/879

PR #713, opened 2023-06-06
DOCS-Add-redirect-IA

bmckercher123:
Add a redirect for CC Candu hardcoded links for IA to the latest docs.

Product Version(s):

Issue:
n/a

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/static/_redirects
+++ b/static/_redirects
@@ -33,6 +33,9 @@
 /calico-enterprise/3.15/ /calico-enterprise/3.15/about-calico-enterprise 301
 /calico-enterprise/3.14/ /calico-enterprise/3.15/about-calico-enterprise 301
 
+# Redirects for CC pages
+/calico-cloud/latest/image-assurance/scanners/cli-based-scanner#start-the-cli-scanner /calico-cloud/latest/image-assurance/overview 301
+
 # Redirects for pages we've moved after the docs migration
 ## Reorganizing 'About' section for OS:
 /calico/latest/about/about-k8s-networking /calico/latest/about/kubernetes-training/about-k8s-networking 301


static/_redirects
```diff
@@ -33,6 +33,9 @@
 /calico-enterprise/3.15/ /calico-enterprise/3.15/about-calico-enterprise 301
 /calico-enterprise/3.14/ /calico-enterprise/3.15/about-calico-enterprise 301
 
+# Redirects for CC pages in Candu
+/calico-cloud/image-assurance/scan-image-registries#start-the-cli-scanner /calico-cloud/latest/image-assurance/overview
```
stevegaossou: @bmckercher123 the direct target is not correct. 

We want to redirect the user to: 
- https://docs.tigera.io/calico-cloud/image-assurance/scanners/cli-based-scanner#start-the-cli-scanner

It's also not specific to Candu. Candu is only one of the instances where the old URL was used. 

I would label this one as an Image Assurance redirect.

bmckercher123: Fixed.

static/_redirects
```diff
@@ -33,6 +33,9 @@
 /calico-enterprise/3.15/ /calico-enterprise/3.15/about-calico-enterprise 301
 /calico-enterprise/3.14/ /calico-enterprise/3.15/about-calico-enterprise 301
 
+# Redirects for CC pages
+/calico-cloud/latest/image-assurance/scanners/cli-based-scanner#start-the-cli-scanner /calico-cloud/latest/image-assurance/overview 301
```
stevegaossou, line 37: Hi @bmckercher123, I think we're still not on the same page. 

If I understand the syntax in this redirect file it is: 
```
source_url dest_url 
```

If my understanding above is correct, then the target URL you have is: 
- /calico-cloud/latest/image-assurance/overview

That target is not correct, as we don't want to redirect the user to the overview page. We want to go to the exact section that outlines the steps to download the CLI scanner. 

We want the target (destination) URL to be: 
- /calico-cloud/image-assurance/scanners/cli-based-scanner#start-the-cli-scanner

Let me know if that doesn't make sense to you. 

Also, I noticed you have `/latest` in the URLs. That doesn't seem correct either. 

The original URL that was used, did not have `/latest` in it.

static/_redirects
```diff
@@ -33,6 +33,9 @@
...
+
```
stevegaossou, line 38: @bmckercher123 I also just realized there were two slightly different URLs that were being used. 

The Image Assurance page had a button called "Download Scanner", which used the URL: 
- https://docs.calicocloud.io/image-assurance/scan-image-registries#start-the-cli-scanner

I believe that is what your current redirect entry covers. 

On the other hand, the Candu URL is set to: 
- https://docs.tigera.io/calico-cloud/image-assurance/scan-image-registries

Which is a slightly different URL. 

I think we want both old URLs to redirect to the same place: 
- https://docs.tigera.io/calico-cloud/image-assurance/scanners/cli-based-scanner#start-the-cli-scanner

So I think this PR needs to have a second entry. 

I can chat in-person with you tomorrow to make sure everything makes sense.

ctauchen:
Closing, fixed elsewhere.

PR #773, opened 2023-06-26
DOCS-1627: Fixes to the install steps

girishshankaran:
Product Version(s):
 Calico 3.24 through version next

Issue:
https://tigera.atlassian.net/browse/DOCS-1627

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico/getting-started/kubernetes/self-managed-onprem/onpremises.mdx
+++ b/calico/getting-started/kubernetes/self-managed-onprem/onpremises.mdx
@@ -40,13 +40,13 @@ installed directly on the cluster as a Deployment, and is configured through one
 <Tabs>
 <TabItem label="Operator" value="Operator-0">
 
-1. First, install the operator on your cluster.
+1. Install the operator on your cluster.
 
    ```
    kubectl create -f {{manifestsUrl}}/manifests/tigera-operator.yaml
    ```
 
-1. Download the custom resources necessary to configure {{prodname}}
+1. Download the custom resources necessary to configure {{prodname}}.
 
    ```
    curl {{manifestsUrl}}/manifests/custom-resources.yaml -O
@@ -58,6 +58,17 @@ installed directly on the cluster as a Deployment, and is configured through one
 
    ```
    kubectl create -f custom-resources.yaml
+   ```
+1. Verify {{prodname}} installation in your cluster.
+
+   ```
+   watch kubectl get pods -n calico-system
+   ```
+   You should see a result similar to the below.
+
+   ```
+   NAMESPACE     NAME                READY   STATUS                  RESTARTS         AGE
+   kube-system   calico-node-txngh   1/1     Running                   0              54s
    ```
 
 <GeekDetails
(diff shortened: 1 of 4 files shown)


girishshankaran:
@lwr20 Can you please review this one?

calico/getting-started/kubernetes/self-managed-onprem/onpremises.mdx
```diff
@@ -59,6 +59,12 @@ installed directly on the cluster as a Deployment, and is configured through one
    ```
    kubectl create -f custom-resources.yaml
    ```
+1. Verify {{prodname}} installation in your cluster.
+
+   ```
+   watch kubectl get pods -n calico-system
+   ```
```
lwr20, line 66: IMO, this is only verification if you know what the expected output is. Perhaps include an example output from a known-good system?

ctauchen: Yes! 

We should have a code block with the title: Example output. And it would be nice to get more information. Something like, If this worked, then you should see this ...

calico/getting-started/kubernetes/self-managed-onprem/onpremises.mdx
```diff
@@ -59,6 +59,12 @@ installed directly on the cluster as a Deployment, and is configured through one
...
+1. (Optional) Verify that the pods can talk to each other and that the network policy works. For more information, see [Kubernetes policy](../../../network-policy/get-started/kubernetes-policy/kubernetes-policy-basic.mdx).
```
ctauchen: Why is this one optional? 

Do we really want to assume that people can figure this out on their own? How do they check that the network policy works? And while we're at it, what network policy have we installed?

girishshankaran: The procedure is for installing Calico. When you install, you should verify if the installation works. Nothing beyond that. Checking the Kubernetes network policy does not seem to be part of the install workflow.

calico/getting-started/kubernetes/self-managed-onprem/onpremises.mdx
```diff
@@ -59,6 +59,18 @@ installed directly on the cluster as a Deployment, and is configured through one
...
+
+   ```
+   watch kubectl get pods -n calico-system
+   ```
+   You should see a result similar to the below.
+
+   ```
+   NAMESPACE     NAME                READY   STATUS                  RESTARTS         AGE
+   kube-system   calico-node-txngh   1/1     Running                   0              54s
+   ```
```
ctauchen, line 72: What do you think of this?
```suggestion
   ```bash title="Example output"
   NAMESPACE     NAME                READY   STATUS                  RESTARTS         AGE
   kube-system   calico-node-txngh   1/1     Running                   0              54s
   ```
```

calico/getting-started/kubernetes/self-managed-onprem/onpremises.mdx
ctauchen: If you want to include a verification step, then you should include information on how to perform the verification. If checking the network policy is not part of the workflow, then why include it here? Right now this seems to be uncomfortably in between.

girishshankaran: Optional step is removed.

PR #883, opened 2023-08-18
Update makefile and run make autogen

rene-dekker:


bmckercher123 (approved):
LGTM. The dead links should clear if you rebase.

bmckercher123:
It is still failing. Please add the following link to skip list: __tests/crawler.test.js

http://www.gnu.org/philosophy/why-not-lgpl.html in

doucol:
> It is still failing. Please add the following link to skip list: __tests/crawler.test.js
> 
> http://www.gnu.org/philosophy/why-not-lgpl.html in

This is getting a 302 for me and redirects you to http://www.gnu.org/licenses/why-not-lgpl.html  -- please fix the link rather than skipping.

rene-dekker:
@doucol  @bmckercher123  I fixed the dead url, but I wonder if this will return at some point. Is this page generated by code from fossa scans?

doucol:
> @doucol @bmckercher123 I fixed the dead url, but I wonder if this will return at some point. Is this page generated by code from fossa scans?

yeah... the test should handle redirects, so perhaps we just caught it at a point where it didn't exist and then the redirect was added.  idk!  I would just leave it as-is for now though.

coutinhop (approved):

PR #923, opened 2023-09-06
EV-4156 add upstream prometheus operator versions to the docs

rene-dekker:


bmckercher123 (approved):
LGTM. Shouldn't this be updated in CC 3.18 and 3.17 as well?

bmckercher123 (approved):
LGTM

PR #936, opened 2023-09-11
[CORE-9833] Document the performanceHints field in (Global)NetworkPolicy.

fasaxc:
Product Version(s):

Calico OS v3.27; Calico Enterprise v3.19

Issue:

Documents https://github.com/projectcalico/calico/pull/7963
CORE-9833
CORE-9780
DOCS-1807

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [x] Test have passed


calico/reference/resources/globalnetworkpolicy.mdx
```diff
@@ -140,6 +142,18 @@ for how `doNotTrack` and `preDNAT` and `applyOnForward` can be useful for host e
 
 <Servicematch />
 
+### Performance Hints
+
+Performance hints provide a way to tell {{prodname}} about the intended use of the policy so that it may
```
bmckercher123, line 147: by "may" do you mean "it might possibly", or should this be "will" or "can" process it more efficiently?

bmckercher123 (approved):
LGTM. Also, copy to CC and CE in appropriate versions.

calico/reference/resources/globalnetworkpolicy.mdx
fasaxc: Yes, "it might possibly".  The hint may help or not, we'll do our best!

calico/reference/resources/globalnetworkpolicy.mdx
```diff
@@ -75,18 +75,20 @@ spec:
...
+| preDNAT\*\*            | Indicates to apply the rules in this policy before any DNAT.                                                                                                                                                                         | true, false               | boolean                | false                                         |
+| applyOnForward\*\*     | Indicates to apply the rules in this policy on forwarded traffic as well as to locally terminated traffic.                                                                                                                           | true, false               | boolean                | false                                         |
+| performanceHints       | Contains a list of hints to Calico's policy engine to help process the policy more efficiently.  Hints never change the enforcement behaviour of the policy.                                                                         | `AssumeNeededOnEveryNode` | List of strings        |                                               |
```
ctauchen:
```suggestion
| performanceHints       | Contains a list of hints to the policy engine to help process the policy more efficiently.  Hints never change the enforcement behavior of the policy.                                                                         | `AssumeNeededOnEveryNode` | List of strings        |                                               |
```

ctauchen (commented):
Just a few things, and a general question.

I presume it's too late to bring this up. But I'm confused by the mental model we've constructed around the idea of a 'hint'. In real life, we give someone a hint to help them understand something, which they may or may not actually get. 

But surely that can't be the case here, right? Any 'hint' you provide to the policy engine will presumably achieve the same result every time. The function described here seems to be more of an extra rule.

Why are we calling this a hint?

fasaxc:
@ctauchen the analogy I'd use is that I hand you a policy written on paper and then say "oh and by the way, we're planning to use this on every node".  You can use that information as you wish or ignore it completely if there's some other more pressing concern.

We could have represented this as "programOnEveryNode: Always" instead (one of the ideas we discussed in the design doc) but  that didn't sit right with me.  If I set `programOnEveryNode` to Always and then looked in iptables/BPF policy dump I'd expect to see it(!)  but it wouldn't be there because another dataplane optimisation kicks in and avoids programming iptables rules that can't be reached.  OTOH, what this flag is really for, is to make sure that the IP sets for the policy get pre-calculated (because those are the expensive things to calculate) and those _would_ be there.  So then we come back to the idea of a hint: you tell me that this policy is likley to be needed and then the (current) policy engine decides to pre-calculate IP sets, but defers iptables writing until later.

ctauchen:
> @ctauchen the analogy I'd use is that I hand you a policy written on paper and then say "oh and by the way, we're planning to use this on every node". You can use that information as you wish or ignore it completely if there's some other more pressing concern.
> 
> We could have represented this as "programOnEveryNode: Always" instead (one of the ideas we discussed in the design doc) but that didn't sit right with me. If I set `programOnEveryNode` to Always and then looked in iptables/BPF policy dump I'd expect to see it(!) but it wouldn't be there because another dataplane optimisation kicks in and avoids programming iptables rules that can't be reached. OTOH, what this flag is really for, is to make sure that the IP sets for the policy get pre-calculated (because those are the expensive things to calculate) and those _would_ be there. So then we come back to the idea of a hint: you tell me that this policy is likley to be needed and then the (current) policy engine decides to pre-calculate IP sets, but defers iptables writing until later.

Thanks for the detailed explanation. It's not an easy one to describe succinctly in a variable name, and what you describe makes sense enough. Sometimes I get to feeling that every one of these things needs a footnote with several diagrams and a chapter or two of explanation ...

ctauchen:
As for moving this along --- would you mind:
* copying to ./calico-enterprise/ and ./calico-cloud/ 
* fetch and rebase (hoping that kicks the build error)
* add a Jira ticket to title for reference

PR #952, opened 2023-09-15
add note block to configuring/dual-tor regarding IP autodetection

aaaaaaaalex:
Product Version(s):

CE:
 - vnext
 - v3.18
 - v3.17
 - v3.16
 - v3.15
 - v3.14

CC:
 - vnext
 - v3.18
 - v3.17
 - v3.16
 - v3.15

Issue:
CI-1316

Link to docs preview:
(To be added)

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-cloud/networking/configuring/dual-tor.mdx
+++ b/calico-cloud/networking/configuring/dual-tor.mdx
@@ -646,6 +646,15 @@ as their source and destination. For example:
 
 The only connections using interface-specific addresses should be BGP.
 
+:::note
+
+If you plan to use [Egress Gateways](../egress/egress-gateway-on-prem.mdx) in your cluster, you must also adjust the {{nodecontainer}} IP
+auto-detection method to pick up the stable IP, for example using the `interface: lo` setting
+(The default first-found setting skips over the lo interface). This can be configured via the
+{{prodname}} [Installation resource](../../reference/installation/api.mdx#operator.tigera.io/v1.NodeAddressAutodetection).
+
+:::
+
 If you look at the Linux routing table on any cluster node, you should see ECMP routes
 like this to the loopback address of every other node in other racks:
 
--- a/calico-cloud/networking/egress/egress-gateway-on-prem.mdx
+++ b/calico-cloud/networking/egress/egress-gateway-on-prem.mdx
@@ -337,6 +337,14 @@ If you are deploying `egress-gateway` on a RKE2 CIS-hardened cluster, its `PodSe
 The Tigera Operator configures egress gateways to use the same iptables backend as `calico-node`.
 To modify the iptables backend for egress gateways, you must change the `iptablesBackend` field in the [Felix configuration](../../reference/resources/felixconfig.mdx).
 
+### Configure IP autodetection for dual-ToR clusters.
+
+If you plan to use Egress Gateways in a [dual-ToR cluster](../configuring/dual-tor.mdx), you must also adjust the {{nodecontainer}} IP
+auto-detection method to pick up the stable IP, for example using the `interface: lo` setting
+(The default first-found setting skips over the lo interface). This can be configured via the
+{{prodname}} [Installation resource](../../reference/installation/api.mdx#operator.tigera.io/v1.NodeAddressAutodetection).
+
+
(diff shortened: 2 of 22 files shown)


calico-cloud/networking/configuring/dual-tor.mdx
```diff
@@ -646,6 +646,16 @@ as their source and destination. For example:
 
 The only connections using interface-specific addresses should be BGP.
 
+:::note
+
+- Some Calico features, such as Egress Gateways, will not work correctly if per-interface
+  IP addresses are used instead of a node's stable IP address.
+- If interface-specific addresses are seen on connections where the stable IP address is
+  expected, the {{nodecontainer}} IP auto-detection method must be adjusted to detect the stable IP.
+  This can be configured via the {{prodname}} [Installation resource](/calico-cloud/reference/installation/api#operator.tigera.io/v1.NodeAddressAutodetection).
```
ctauchen:
```suggestion
  This can be configured via the {{prodname}} [Installation resource](../../reference/installation/api.mdx#operator.tigera.io/v1.NodeAddressAutodetection).
```
This should work and fix your build errors. We prefer to use relative file links. The page ref should work too.

ctauchen, line 655:
```suggestion
  This can be configured via the {{prodname}} [Installation resource](/calico-cloud/reference/installation/api#operator.tigera.io/v1.NodeAddressAutodetection).
```

aaaaaaaalex: ah, nice one :raised_hands:

aaaaaaaalex: I think I've added this change to all the links now :ok_hand:

calico-cloud/networking/configuring/dual-tor.mdx
```diff
@@ -646,6 +646,16 @@ as their source and destination. For example:
 
 The only connections using interface-specific addresses should be BGP.
 
+:::note
+
+- Some Calico features, such as Egress Gateways, will not work correctly if per-interface
+  IP addresses are used instead of a node's stable IP address.
```
nelljerram: As an uninformed reader, I think my reaction to that bullet would be "OK.  But why are you telling me that, and what can I do about it?  Isn't the whole doc above this point about what IP addresses are used?"

I think it shouldn't be a bullet on its own - because it isn't a standalone note - and the wording can be improved.  Unless we can immediately think of another relevant feature, let's make it simpler by just saying egress gateway:

"If you plan to use Egress Gateways in your cluster, you must also adjust the {{nodecontainer}} IP auto-detection method to pick up the stable IP, for example using the `interface: lo` setting.  (The default `first-found` setting skips over the `lo` interface.)  This can be configured ..."

nelljerram (changes requested):

ctauchen (commented):
@aaaaaaaalex I'm wondering about where this information is best placed. Supposing we added this information only in the egress gateway docs -- would something be missed? Is it essential that this gets set up as part now, and not later when the user configures EGW?

aaaaaaaalex:
@ctauchen 
I see no reason we couldn't put it with the egress gateway docs in fairness - my main reason for putting it here was I felt it really is a part of "correctly" configuring dual-tor to have that autodetection working right. But in practice, problems only manifest when you go to provision an EG, so I'm happy to move it over.

PS - the setting can be applied at any time, yes.

aaaaaaaalex:
Added a section to the egress gateway docs now.
Should I leave this info in both pages or remove it from the dual-ToR pages? Personally I think both makes sense

nelljerram:
I agree that it makes sense to document this from both places.  But considering the exact wording, will it be clear enough what "stable IP" means, to someone reading through the egress gateway page?  OTOH, on the other hand, perhaps it's fine because there's a link to dual ToR, hence it's obvious that someone should go look there if they need more context.

Anyway, perhaps this might be polished even further, but it's already a great improvement, so will approve...

nelljerram (approved):

aaaaaaaalex:
Checked the preview and all looks good. Squashing commits.

PR #957, opened 2023-09-19
[RS-1186] Update security events FAQ to mention restore option

gantony:
Update security events FAQ to mention restore option. This was added as part of a bug fix for 3.18.1. It is not currently available in 3.18.0 or CC. 

Will need another PR to update the remaining versions (or update here and hold the merge, **please advise and best way to proceed**).

Product Version(s):
Applies to CC 18 (upcoming) and EE 3.18 GA (not in "early preview")

Issue:
https://tigera.atlassian.net/browse/RS-1186

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


ctauchen:
@gantony Let's go ahead an add this to all relevant versions -- CE next/3.18, CC next/18. We'll hold on to the PR and merge as part of the CE 318.1 process.

calico-cloud/threat/security-event-management.mdx
```diff
@@ -83,12 +83,12 @@ No. {{prodname}} security events do not encompass all types of alerts nor all se
 
 **What does dismissing a security event do?**
 
-Dismissing a security event hides it from view. Use the toggle button to show/hide. There is no option to undo a dismissed alert.
+Dismissing a security event hides it from view. Use the toggle button to show/hide. If dismissed by mistake, a security event can be restored.
```
bmckercher123: If the board has been updated and both dismissing and restoring are intuitive, I'd get rid of the entire section. But if restoring is not intuitive, let's be more specific:

If you dismiss an event in error, you can restore it by...

bmckercher123 (approved):
Overall, LGTM. Just a question about how to restore an event.

calico-cloud/threat/security-event-management.mdx
gantony: It's intuitive and uses the same workflow (if an item is dismissed, you can click "Restore Security Event"; if not dismissed you can click "Dismiss Security Event").

I think we could leave it as is to keep "Dismissing a security event hides it from view" as some people may be wondering "what does dismiss do?"

bmckercher123: Agree. The bullet should read:

- Dismissing a security event hides it from view

Because CE 3.18.1 is already published, you can go ahead and make the updates to CC 3.18 as well in the same PR. Then, we can merge this. Unless you disagree, I don't believe it warrants a Release note entry because it is too low-level.

ctauchen:
LGTM, thanks!

PR #1005, opened 2023-10-05
DOCS-1824: Remove all docs for Anomaly Detection in Calico Enterprise

nelljerram:
Part of PMREQ-473.


ctauchen:
LGTM.

To be followed up with CC fixes and redirects.

PR #1072, opened 2023-11-03
DOCS-1959-epbf-use-cases

bmckercher123:
Product Version(s):
CE, CC, OSS multiple versions

Issue:
https://tigera.atlassian.net/browse/DOCS-1959

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [x] Test have passed

### DIFF
--- a/calico-enterprise_versioned_docs/version-3.18/operations/ebpf/use-cases-ebpf.mdx
+++ b/calico-enterprise_versioned_docs/version-3.18/operations/ebpf/use-cases-ebpf.mdx
@@ -66,10 +66,6 @@ Let’s look at some specific cases where it would make sense to use eBPF, and s
 
 ### When not to use eBPF
 
-### ✘ Building a service mesh control plane
-
-Service mesh relies on proxies like Envoy. A lot of thought has gone into designing this process over the years. The main reason for doing it this way is that, in many cases, it is not viable to do inline processing for application protocols like HTTP at the high speeds seen inside a cluster. Therefore, you should think of using eBPF to route traffic to a proxy like Envoy in an efficient way, rather than using it to replace the proxy itself.
-
 ### ✘ Packet-by-packet processing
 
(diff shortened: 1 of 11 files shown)


calico-enterprise_versioned_docs/version-3.18/operations/ebpf/use-cases-ebpf.mdx
```diff
@@ -82,7 +82,7 @@ With eBPF, you can load balance at the source using a program you’ve loaded in
...
+Service mesh relies on proxies like Envoy. A lot of thought has gone into designing this process over the years. The main reason for doing it this way is that, in many cases, it is not viable to do inline processing for application protocols like HTTP at the high speeds seen inside a cluster. Therefore, you should think of using eBPF to route traffic to a proxy like Envoy in an efficient way, rather than using it to replace the proxy itself. However, you do need to turn off connect-time load balancing so sidecars can see the service addresses. Given you are already taking a significant performance hit, using Cross Task Load Balancing (CTLB) is not a big deal.
```
tomastigera:
```suggestion
Service mesh relies on proxies like Envoy. A lot of thought has gone into designing this process over the years. The main reason for doing it this way is that, in many cases, it is not viable to do inline processing for application protocols like HTTP at the high speeds seen inside a cluster. Therefore, you should think of using eBPF to route traffic to a proxy like Envoy in an efficient way, rather than using it to replace the proxy itself. However, you do need to turn off connect-time load balancing  (CTLB) so sidecars can see the service addresses. Given you are already taking some performance hit by the extra hop to the sidecar, not usign CTLB performance optimization to avoid NAT overhead, is likely not a big deal.
```

bmckercher123: Fixed.

PR #1091, opened 2023-11-14
Add 'make start-with-vnext' command

coutinhop:
Add 'make start-with-vnext' command that includes the 'current' version in onlyIncludeVersions in the docusaurus config, so that one can visualize the 'next' version of the locally rendered website (which doesn't show up in 'make start' since this PR: https://github.com/tigera/docs/pull/1052/)

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,10 @@ build: init
 start: init
 	$(YARN) start$(YARN_ACTION_SUFFIX)
 
+.PHONY: start-with-vnext
+start-with-vnext: init
+	$(YARN) start-with-vnext$(YARN_ACTION_SUFFIX)
+
 .PHONY: test
 test: init
 	./scripts/serve-test.sh
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,9 @@ build: init
 start: init
 	$(YARN) start$(YARN_ACTION_SUFFIX)
 
-.PHONY: start-with-vnext
-start-with-vnext: init
-	$(YARN) start-with-vnext$(YARN_ACTION_SUFFIX)
+.PHONY: start-next
+start-next: init
+	$(YARN) start-next$(YARN_ACTION_SUFFIX)
 
 .PHONY: test
 test: init
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,9 @@ CALICO_CLOUD_BRANCHES=calico-cloud__operator_reference $(addsuffix __operator_re
 build: init
 	$(YARN) build
 
+build-next: init
+	$(YARN) build-next
+
 .PHONY: start
 start: init
 	$(YARN) start$(YARN_ACTION_SUFFIX)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -26,6 +26,10 @@ build: init
 start: init
 	$(YARN) start$(YARN_ACTION_SUFFIX)
 
+.PHONY: start-with-vnext
```
ctauchen: Perhaps just `make start-next`?

I can see wanting a similar `build-next`, but we can wait for now.

docusaurus.config.with-vnext.js
```diff
@@ -0,0 +1,17 @@
+// This file imports the config from docusaurus.config.js and adds the 'current' version to
```
ctauchen: I'm thinking the best place for this file would be `./scripts`.

docusaurus.config.with-vnext.js
```diff
@@ -0,0 +1,17 @@
...
+// onlyIncludeVersions, so that the 'next' version is available when running
+// 'make start-with-vnext' locally.
+
+let config = require('./docusaurus.config.js');
```
ctauchen: We're in the process of moving to Docusuarus v3. Among other things, this drops support for CommonJS. Can you rewrite this to be compatible with ESM?

ctauchen (commented):
@coutinhop I think this is really useful, thanks! A few comments below.

ctauchen:
@ronanc-tigera @HugoJLFerreira Anything to add?

docusaurus.config.with-vnext.js
coutinhop: @ctauchen Apologies for the extremely long delay. Let me begin by admitting js is as far as it can get from my strong suite, so let me know if I'm missing something. I tried updating into using `import / export` instead of `require()`, but kept getting errors, like this one:
```
[ERROR] Error [ERR_REQUIRE_ESM]: require() of ES Module docusaurus.config.with-vnext.mjs not supported.
```
(This was after I changed the extension to `.mjs` to indicate that this was a module, otherwise I'd get a 'cannot import outside a module' error.)

I think this is because of the original `docusaurus.config.js` file that this one imports, so it's like a chicken and egg problem. I did add a 'TODO' comment with the new syntax (as far as I can tell), but I don't think I can update this before the docusaurus v3 upgrade...

docusaurus.config.with-vnext.js
coutinhop: Done, thanks!

Makefile
coutinhop: Done, thanks!

coutinhop: Just pushed a commit adding `build-next`, please take a look!

docusaurus.config.with-vnext.js
ctauchen: Thanks, I've had a look and tested and all looks good. We'll adjust the script for the Docusuarus 3 upgrade when we get there.

ctauchen:
LGTM

PR #1141, opened 2023-11-27
Document that data is retained when a managed cluster is deleted

rene-dekker:


calico-enterprise/multicluster/change-cluster-type.mdx
```diff
@@ -171,6 +171,14 @@ If you wish to retain LogStorage data for your managed cluster, verify that the
 
    :::
 
+   Follow the following step to also disconnect your managed cluster from your management cluster. This will not result in any
+   data loss. If you reconnect your cluster using the same name, existing data will be available again.
+
+   1. In your management cluster, delete the `ManagedCluster` resource for your managed cluster.
```
bmckercher123:
```suggestion
   1. In your management cluster, delete the `ManagedCluster` resource for your managed cluster. Note that deleting the managed cluster does not result in loss of data. Whenever you reconnect your cluster using the same name, existing data is available again. 
```

bmckercher123 (commented):
Good catch to add this info. Suggest to move the blurb with the step.

bmckercher123:
@rene-dekker Chris just cut 3.19 so you'll need to rebase and copy changes to 3.19.

PR #1154, opened 2023-11-30
Update helm install customization instructions

ti-afra:
What:
User should be providing values.yaml file during helm installs. Thus, we need to highlight it more as a "must have", not optional.

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
+++ b/calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
@@ -38,17 +38,22 @@ In this guide, you install the Tigera Calico operator and custom resource defini
        : `curl -O -L https://downloads.tigera.io/ee/charts/tigera-operator-{{chart_version_name}}.tgz`}
   </CodeBlock>
 
-### Customize the Helm chart
+### Prepare the Installation Configuration
 
-If you are installing on a cluster installed by EKS, GKE, AKS or Mirantis Kubernetes Engine (MKE), or you need to customize TLS certificates, you **must** customize this Helm chart by creating a `values.yaml` file. Otherwise, you can skip this step.
+You **must** provide the desired configuration for your cluster via the `values.yaml`, otherwise installation will use the default settings based on the auto-detected provider.
+The configurations you need to provide depends on your cluster's settings and your desired state.
 
-1. If you are installing on a cluster installed by EKS, GKE, AKS or Mirantis Kubernetes Engine (MKE), set the `kubernetesProvider` as described in the [Installation reference](../../../reference/installation/api.mdx#operator.tigera.io/v1.Provider). For example:
+Some important configurations you might need to provide to the installer (via `values.yaml`) includes (but not limited to): _kubernetesProvider_, _cni type_, or if you need to customize _TLS certificates_.
+
+Here are some examples for updating `values.yaml` with your configurations:
+
+Example 1. Providing `kubernetesProvider`:  if you are installing on a cluster installed by EKS, set the `kubernetesProvider` as described in the [Installation reference](../../../reference/installation/api.mdx#operator.tigera.io/v1.Provider)
 
   ```bash
   echo '{ installation: {kubernetesProvider: EKS }}' > values.yaml
   ```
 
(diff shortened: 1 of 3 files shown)


calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
```diff
@@ -64,13 +69,7 @@ If you are installing on a cluster installed by EKS, GKE, AKS or Mirantis Kubern
   EOF
   ```
 
-1. Add any other customizations you require to `values.yaml`. To see values that can be customized in the chart, see the [helm docs](https://helm.sh/docs/) or run the following command:
-
-  <CodeBlock language='bash'>
```
dimitri-nicolo, line 69: Why are we removing the section indicating how to examine your `values.yaml`?

ti-afra: 1. The link to helm customization doc page is missing from this paragraph and we have general helm docs page instead (not sure how general docs can help users figure out how they can customize their calico installation values.yaml)

2. The command in the docs ``helm show values ./tigera-operator-{{chart_version_name}}.tgz`}` just shows the current values set in valuels.yaml. I am again not sure how that helps users figure out what are the options they can define in values.yaml and how they can initialize them.

So I just replaced this whole section with:
```
For more information about configurable options via values.yaml please see [Helm installation reference](../../reference/installation/helm_customization).
```

dimitri-nicolo (approved):

ctauchen (commented):
After reading this a few times, I'm still not certain exactly what you're trying to change here. In your main comment, you say that you're trying to make it clear that the `values.yml` file is mandatory.

But then: "otherwise installation will use the default settings based on the auto-detected provider."

This makes it sound like a MUST only if you don't want the default settings. Is that right?

The current text says clearly that you need to customize the Helm chart if you're on certain providers. 

Are you saying now that we need to customize in all circumstances?

ti-afra:
> After reading this a few times, I'm still not certain exactly what you're trying to change here. In your main comment, you say that you're trying to make it clear that the `values.yml` file is mandatory.
> 
> But then: "otherwise installation will use the default settings based on the auto-detected provider."
> 
> This makes it sound like a MUST only if you don't want the default settings. Is that right?
> 
> The current text says clearly that you need to customize the Helm chart if you're on certain providers.
> 
> Are you saying now that we need to customize in all circumstances?

This is feedback coming from people within our org who have tried helm installation using this doc. It's not obvious for people that they have to provide settings for their cluster when using helm installation. People think we dont have special settings, but the truth is that default setting might differ based on the provider people use. For example, you don't always get Calico as the CNI. If you don't specify it in the values.yaml the code might set something else as you CNI and then the whole cluster is broken and people don't know what's going on...

The goal here is to convey this message that users have to provide values.yaml with their installation and they have to think about the settings for their cluster and define it at the time of installation, otherwise who knows what they are getting as defaults. Unless they are 100% sure that defaults will work for them.

ctauchen:
> > After reading this a few times, I'm still not certain exactly what you're trying to change here. In your main comment, you say that you're trying to make it clear that the `values.yml` file is mandatory.
> > But then: "otherwise installation will use the default settings based on the auto-detected provider."
> > This makes it sound like a MUST only if you don't want the default settings. Is that right?
> > The current text says clearly that you need to customize the Helm chart if you're on certain providers.
> > Are you saying now that we need to customize in all circumstances?
> 
> This is feedback coming from people within our org who have tried helm installation using this doc. It's not obvious for people that they have to provide settings for their cluster when using helm installation. People think we dont have special settings, but the truth is that default setting might differ based on the provider people use. For example, you don't always get Calico as the CNI. If you don't specify it in the values.yaml the code might set something else as you CNI and then the whole cluster is broken and people don't know what's going on...
> 
> The goal here is to convey this message that users have to provide values.yaml with their installation and they have to think about the settings for their cluster and define it at the time of installation, otherwise who knows what they are getting as defaults. Unless they are 100% sure that defaults will work for them.

Ok, thanks for that detailed explanation. I'll schedule a meeting so we can talk through the nuances here. I think I'm still missing something. But I'm sure we can clear it up and make sure that our users understand your intent clearly.

PR #1222, opened 2023-12-27
Fix Calico OSS variables.js

coutinhop:
Add missing windows variables and update versions to v3.27 where needed.

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


calico_versioned_docs/version-3.27/variables.js
```diff
@@ -9,12 +9,14 @@ const variables = {
   filesUrl: 'https://projectcalico.docs.tigera.io/v3.27',
   tutorialFilesURL: 'https://docs.tigera.io/files',
   calicoReleasesURL: 'https://github.com/projectcalico/calico/releases/download',
-  tmpScriptsURL: 'https://docs.tigera.io/calico/3.26',
+  tmpScriptsURL: 'https://docs.tigera.io/calico/3.27',
+  windowsScriptsURL: 'https://raw.githubusercontent.com/kubernetes-sigs/sig-windows-tools/master/hostprocess',
```
danudey, line 13: I'm very uncomfortable linking to `master` branches and/or external repositories; as we've seen in other cases recently, they can change arbitrarily and in incompatible ways. Is this the best place to link to? Would it make sense to vendor this script ourselves to ensure compatibility? Or is this a known-stable file path that we can rely on long-term?

danudey (approved):

calico_versioned_docs/version-3.27/variables.js
coutinhop: Thanks @danudey! Yes, the risk has definitely been considered, but we ended up deciding against vendoring+maintaining these scripts ourselves. They are and have been stable for quite some time. (We could consider pointing to specific hash versions of their tree when cutting our versions, though... What do you think?)

PR #1248, opened 2024-01-17
[RS-1564] Doc for anomaly detection reinstatement

nelljerram:
We are recreating some anomaly detection function within the Runtime Security architecture.  Initially only the DGA detection algorithm.  As the design doc, for how we are choosing to present this:

Overall, our Calico Cloud documentation will continue to present “Anomaly Detection” (aka “Detect and alert on anomalies”) and “Container Threat Detection” as distinct features.  There is no anticipated need for the user to know that we now do these within the same architecture.  That also means that the UI control for “Enable Container Threat Detection” will continue to be accurate and valid.

More specifically, in the Calico Cloud docs, the previous page about anomaly detection “Detect and alert on anomalies” will be restored, with the following changes.

- Material about configuration will be removed, as configuration is now unnecessary.

- Material about specific algorithms - only DGA, in the first release - will be incorporated into this page, instead of in a linked reference page (previously https://docs.tigera.io/calico-cloud/reference/anomaly-detection).

- Page will be generally reviewed and revised to ensure correctness for the new implementation.

The user will see security events for DNS lookups for algorithm-generated domain names, if such lookups are occurring in their cluster.  In the previous AD architecture this would have been subject to the existence of a GlobalAlert resource with type: AnomalyDetection and detector: dga, but now it will be unconditional.

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


calico-cloud/threat/security-anomalies.mdx
```diff
@@ -0,0 +1,82 @@
...
+Generate alerts for anomalous activity within a cluster.
+
+## Value
+
+The {{prodname}} anomaly detection engine analyzes patterns and alerts on potential
+threats and issues such as:
+
+- **Security issues** - threats like DGA, reconnaissance threats like IP sweep, but also issues with servers that can affect daily operations.
+
+- **Performance issues** - spikes in data transmission, and anomalous degradation in network communication that may impact application workloads.
```
peterkellydev: Technically we do not have any of these types of performance anomaly detection in the new CC engine yet. We could remove this or leave it there just as a reference to different possible types of anomalies.

calico-cloud/threat/security-anomalies.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+This how-to guide uses the following {{prodname}} features:
+
+- **Anomaly detection**
+
+## Concepts
+
+### About anomalies
+
+Anomaly detection uses {{prodname}} Elasticsearch logs ([flow](../visibility/elastic/flow/index.mdx) logs, [L7](../visibility/elastic/l7/index.mdx) logs, and [DNS](../visibility/elastic/dns/index.mdx) logs) to learn the behavior of cluster nodes, pods, services, and other entities that send log records (applications, load balancers, databases, etc.).
```
peterkellydev: Could possibly drop "Elasticsearch" and just mention logs. I think we try not to reference ES in the docs now (I know this is an old doc being moved over).

peterkellydev (approved):
LGTM, couple of minor ideas / feedback

calico-cloud_versioned_docs/version-18-2/threat/security-anomalies.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+Anomaly detection uses {{prodname}} Elasticsearch logs ([flow](../visibility/elastic/flow/index.mdx) logs, [L7](../visibility/elastic/l7/index.mdx) logs, and [DNS](../visibility/elastic/dns/index.mdx) logs) to learn the behavior of cluster nodes, pods, services, and other entities that send log records (applications, load balancers, databases, etc.).
+
+The current Calico Cloud release includes the following anomaly detection algorithms.
+
+#### Domain Generation Algorithms (DGA)
+
+Looks for DNS requests for domain names that are likely to be created by [Domain Generation Algorithms (DGA)](https://en.wikipedia.org/wiki/Domain_generation_algorithm), frequently used by malware. Generated domain names (URLs) are used to communicate between the malware code and the malware's command and control servers. Presence of DGA may indicate presence of malware.
+
+## FAQ
```
bmckercher123: FAQ entry is still good info, but consider moving it to Concepts so it fits more with the standard:

## Concepts

### About anomalies

Anomaly detections run by default (no enabling required) and are preconfigured with reasonable defaults that are optimized for frequency and severity. Anomaly detection uses {{prodname}} logs ([flow](../visibility/elastic/flow/index.mdx) logs, [L7](../visibility/elastic/l7/index.mdx) logs, and [DNS](../visibility/elastic/dns/index.mdx) logs) to learn the behavior of cluster nodes, pods, services, and other entities that send log records (applications, load balancers, databases, etc.).

calico-cloud_versioned_docs/version-18-2/threat/security-anomalies.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+**Do I need to configure anomaly values?**
+
+> No.  Anomaly detections run unconditionally and are preconfigured with reasonable defaults that are optimized for frequency and severity.
+
+## Before you begin
+
+**Supported**
+
+- All Calico Cloud platforms
```
bmckercher123: You can delete the **Supported** section. Several years ago we switched to a model of calling out only platforms that were not supported and other limitations. The user can assume it is supported on all platforms unless otherwise documented. The change happened when we created the Support and Compatibilty matrix which states this policy. (This doc is old so it just wasn't updated.)

calico-cloud/threat/security-anomalies.mdx
bmckercher123: Given this is tech preview, I think the paragraphs Peter mentions are ok to leave as they give some context of what the feature is trying to solve.

calico-cloud_versioned_docs/version-18-2/threat/security-anomalies.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+## How To
+
+- [Monitor anomaly alerts in Manager UI](#monitor-anomaly-alerts-in-manager-ui)
```
bmckercher123: With only one section, you can delete the inline link.

bmckercher123 (commented):
A few spots to consider, but looks good. Kudos for also updating the associated references to anomaly detection.

ctauchen:
After discussing with @nelljerram, I recommend that we reduce the scope to 1) a release note and 2) updating the table of Mitre tactics on the [Security Event Management](https://docs.tigera.io/calico-cloud/threat/security-event-management) page.

There are a few reasons here, mostly to do with what users need to do and understand. 
* This alert is on by default and cannot be modified or disabled.
* A user who receives a DGA alert may want more information. The Mitre technique number is the best lead, but that information is outside our docs ( We don't currently document sub-techniques such as this one).
* The table of Mitre techniques should include the category of attack that DGA belongs to (command and control).
* It's really not relevant that this detection came from a system once called Anomaly Detection. This is now just one of many Security Event Alerts and should be treated as such.

Docs team will take care of the release note, which can include more specifics about DGA.

CC @peterkellydev @bmckercher123

calico-cloud/threat/security-anomalies.mdx
nelljerram: This page is completely reverted now.

nelljerram:
@bmckercher123 FYI I have made this change a lot smaller and simpler after prompting from @ctauchen and agreement from @peterkellydev .  Please see Chris's comment above at https://github.com/tigera/docs/pull/1248#issuecomment-1900154802 and my summary of the change of direction in the message of [this commit](https://github.com/tigera/docs/pull/1248/commits/d5356026abb3bcb7ab726f4e3041b5ff573954f6).

bmckercher123 (approved):
LGTM

PR #1249, opened 2024-01-17
DOCS-Add-nftables-known-issue

bmckercher123:
Doug and Erik approved adding to CC RN.

Product Version(s):
OSS, CE, CC unarchived version RNs

Issue:

Link to docs preview:
https://deploy-preview-1249--tigera.netlify.app/calico/latest/release-notes/

SME review:

DOCS review:

Additional information:

Merge checklist:

- [x] Test have passed


calico_versioned_docs/version-3.27/release-notes/index.mdx
```diff
@@ -97,10 +97,14 @@ For more information, see [Enable the eBPF data plane](../operations/ebpf/enabli
  - eBPF: handles IPv4 packets with IP options. XDP only does so in kernels 5.18+ it drops them in older kernels. [calico #7701](https://github.com/projectcalico/calico/pull/7701) (@tomastigera)
  - eBPF - only report errors setting /proc/sys/net/../accept_local etc. if the device is present [calico #7631](https://github.com/projectcalico/calico/pull/7631) (@tomastigera)
 
+## Known issues
+
+- Calico panics if kube-proxy or other components are using nftable mode
```
fasaxc:
```suggestion
- Calico panics if kube-proxy or other components are using native `nftables` rules instead of the `iptables-nft` compatibility shim.
```

calico_versioned_docs/version-3.27/release-notes/index.mdx
```diff
@@ -97,10 +97,14 @@ For more information, see [Enable the eBPF data plane](../operations/ebpf/enabli
...
+
+   Until Kubernetes nftables mode is fully released (currently in Alpha v1.29), we recommend that you use iptables-nft for all your components (rather a mix of legacy iptables and nftables) for production environments. Although using a mix of Calico iptables and nftables does not fail (assuming your kernel supports both), the behaviour is confusing, and policies may not be properly enforced because nftables can "undo" the verdict made by iptables-legacy.
```
fasaxc:
```suggestion
   Until Calico supports native nftables mode, we recommend that you continue to use the iptables-nft compatibility layer for all components.  (This was the only option before Kubernetes v1.29 added alpha-level `nftables` support.)
   
    Do not run calico in "legacy" iptables mode on a system that is also  using `nftables`.  While this combination does not panic or fail (at least on kernels that support both), the interaction between `iptables` "legacy" mode and `nftables` is confusing: both `iptables` and `nftables` rules can be executed on the same packet, leading to policy verdicts being "overturned". 
```

fasaxc (changes requested):

ctauchen (commented):
Playing devil's advocate: Do we really need to document the fact that we don't support a Kubernetes alpha feature that Kubernetes itself doesn't support?

Is there a strong reason for this particular case? I would want to go down the path of vetting all the pre-GA Kubernetes features.

@fasaxc

bmckercher123:
@ctauchen In this case, it is an important bit of information because it shows the root issue is because Calico is lock step with Kubernetes, and until they fix their side, our interim solution is the best we can do. If we do not mention it, users may believe we are not fixing the root cause and sidestepping a true fix. Users who brought up this issue in the first place were quite peeved that we could not play nicely with nftables. So I do think it is warranted to dial down the reaction to the issue.

fasaxc (approved):

PR #1250, opened 2024-01-18
[EV-4549] Add a note about setting DelayDNSResponse for Staged DNS Policies

ti-afra:
Product Version(s):

Issue:

https://tigera.atlassian.net/browse/EV-4549

Link to docs preview:

SME review:

DOCS review:

Additional information:

policy egress rule matches for domain names are not consistent for DNS staged policies when the domain address in policy rule is hit. The reason for that is felix is matching based on IP address not domain name and when we are using domain name instead of ip address, the first call will fail but will include the dns response that felix will use to update the rules. Based on the time to live of this domain to ip address matching, next hits will be successful. 
To resolve this inconsistency, there is a flag defined in felix called DNSPolicyMode. By setting this flag to DelayDNSResponse felix will replace domain name will ip address in policy rules before matching rules for the traffic.

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/network-policy/domain-based-policy.mdx
+++ b/calico-enterprise/network-policy/domain-based-policy.mdx
@@ -227,6 +227,24 @@ spec:
       destination:
         selector: color == 'red'
 ```
+:::note
+
+Newly initialized pods sometimes fail to connect to domains that are allowed by a DNS policy (staged or enforced).
+In these cases, the pod tries to connect to a domain before {{prodname}} has finished processing the DNS policy that allows the connection.
+As soon as the processing is complete, the pod is able to connect.
+
+To avoid these failed connections, you can add the following to your `FelixConfiguration` resource:
+
+```yaml
+...
+spec:
+  DNSPolicyMode: DelayDNSResponse
+...
+```
+
+For more information, see [DNSPolicyMode](../reference/resources/felixconfig#dnspolicymode).
+
+:::
 
 ## Additional resources
 


ctauchen:
@ti-afra Can you provide some context here? Link to ticket?

ti-afra:
> @ti-afra Can you provide some context here? Link to ticket?

@ctauchen  Sorry for forgetting to link to the ticket. I have update all the info in the description section above.

ctauchen:
Thanks for the extra information!

calico-enterprise/network-policy/domain-based-policy.mdx
```diff
@@ -229,6 +229,13 @@ spec:
       destination:
         selector: color == 'red'
 ```
+## Staged DNS Policy
```
bmckercher123, line 232: As written, it sounds like we are covering up a bug and forcing users to set a flag unnecessarily. I suspect we just need to emphasize the reason for choosing to align staged and enforced policy. For example:

When you stage a DNS policy, the goal is to preview and test flows to ensure the policy is correctly allowing or blocking traffic before enforcement. When a DNS policy first tries to reach a specified domain address or networkset, there can be a delay in reaching the domain that can result in a “no match” in flow logs in troubleshooting. Although this condition does not occur in enforced policies, you may want to align flow log results between staged and enforced policy by changing the `FelixConfiguration` parameter from `DNSPolicyMode` to `DelayDNSResponse`.

bmckercher123: As written, it sounds like we are covering up a bug making them change a flag for some inconsistency on our part. I suspect we just need to emphasize that aligning results in flow logs between the two types of policy is a choice. For example:

When you stage a DNS policy, the goal is to preview and test traffic flows to ensure the policy is correctly allowing or blocking before enforcement. When a DNS staged policy first tries to reach a specified domain address or networkset, there can be a delay in the host reaching the domain that can result in a “no match” for the domain in flow logs. This flow log condition does not occur in enforced policies, but if you want to align flow log behavior between staged and enforced policy, change the `FelixConfiguration` parameter from `DNSPolicyMode` to `DelayDNSResponse`.

calico-enterprise/network-policy/domain-based-policy.mdx
```diff
@@ -229,6 +229,13 @@ spec:
...
+
+If you define a staged DNS policy the rule to allow egress on the domains specifid in the staged policy or the networkset
```
nelljerram: Add comma after first "policy"
sp: "specified"

calico-enterprise/network-policy/domain-based-policy.mdx
```diff
@@ -229,6 +229,13 @@ spec:
...
+used is not always matched. This is because of the hostname to ip address conversion for the first time you try to reach
+to the domain address and the ttl for that conversion. This will not happen for enforced policies, but if you prefer to
+always have the consistency in your staged policies, you need to update `FelixConfiguration` and set `DNSPolicyMode` to
+`DelayDNSResponse`.
```
nelljerram: Agree with Leon that this is expected for enforced policies as well.

calico-enterprise/network-policy/domain-based-policy.mdx
```diff
@@ -229,6 +229,13 @@ spec:
       destination:
         selector: color == 'red'
 ```
+## Staged DNS Policy
+
+If you define a staged DNS policy the rule to allow egress on the domains specifid in the staged policy or the networkset
+used is not always matched. This is because of the hostname to ip address conversion for the first time you try to reach
+to the domain address and the ttl for that conversion. This will not happen for enforced policies, but if you prefer to
```
nelljerram: I think this can be expressed better, and with a narrower scope - making clear that it only applies to connection setups just after a DNS lookup - as follows:

"When a pod first connects to a domain that should be allowed by DNS policy, you may observe that the first connection setup packet is denied.  This is because {{site.prodname}} is still updating iptables or eBPF programming to allow the IP addresses that were indicated in the DNS lookup response, _at the same time_ as the pod is connecting to one of those IP addresses.  If the connection is TCP-based - such as an HTTP request - the TCP level will automatically send another connection setup packet after 1 second, and that packet should be allowed (unless the node concerned is so heavily loaded that {{site.prodname}} is struggling to get any CPU!), which means that the connection is established after a latency of 1 second.  If the connection is UDP-based, it is up to the application to retry connection setup - i.e. what happens will depend on the detailed coding of the particular application that the pod is using.

"You can eliminate the above concerns by setting `DNSPolicyMode` in `FelixConfiguration` to `DelayDNSResponse`.  This works by ensuring that the pod doesn't see the DNS lookup response - and so doesn't try to connect to one of its IP addresses - until {{site.prodname}} has fully updated iptables or eBPF to allow those IP addresses.  But it has a small downside of adding a slight latency to every DNS lookup that pods make, and that is why it is not the default setting of `DNSPolicyMode`."

nelljerram (changes requested):

calico-enterprise/network-policy/domain-based-policy.mdx
ti-afra: Thanks @nelljerram - One more thing that I wanted to clarify here was that why the behaviour between staged and enforces dns policies are different. Do you think it would help if we first give them a high level explanation and then provide a more elaborate technical explanation?

calico-enterprise/network-policy/domain-based-policy.mdx
ti-afra: makes sense! I did no have enough background info and complete picture :+1: thank you!

calico-enterprise/network-policy/domain-based-policy.mdx
nelljerram: I am not aware of what the additional problems with staged policies are.  Can you describe them?

calico-enterprise/network-policy/domain-based-policy.mdx
ti-afra: Thank you @bmckercher123 good point - To elaborate more on this, the reason we see this behaviour is because flowlog matching works with ip address. Thus, host names need to be converted to ip address which happens after the first call to the host and then iptables are updated - But also this conversion is only valid sometimes for a short time and expires. So even if the connection to host doesn't take much it will happen from time to time.

ctauchen:
LGTM!

PR #1254, opened 2024-01-22
[RS-1698] SecurityEventWebhook resource definition.

bartolini:
Original Jira ticket: https://tigera.atlassian.net/browse/RS-1698


calico-enterprise/reference/resources/webhook.mdx
```diff
@@ -0,0 +1,151 @@
+---
+description: API for this Calico Enterprise resource.
+---
+
+# Security Event Webhook
+
+A Security Event Webhook (`SecurityEventWebhook`) is a cluster scoped resource that represents instances
```
ctauchen:
```suggestion
A Security Event Webhook (`SecurityEventWebhook`) is a cluster-scoped resource that represents instances
```

calico-enterprise/reference/resources/webhook.mdx
```diff
@@ -0,0 +1,151 @@
...
+| consumer | Specifies intended consumer of the webhook.                                                                     | Slack, Jira, Generic                | string                                                                  | yes                         |
+| state    | Defines current state of the webhook.                                                                           | Enabled, Disabled, Debug            | string                                                                  | yes                         |
+| query    | Defines query used to retrieve security events from Calico.                                                     | [see Query](#query)                 | string                                                                  | yes                         |
+| config   | Webhook configuration, required contents of this structure is determined by the value of the `consumer` field.  | [see Config](#configuration)        | list of [SecurityEventWebhookConfigVar](#securityeventwebhookconfigvar) | yes                         |
+
+### SecurityEventWebhookConfigVar
```
gantony: Find the heading a bit odd. Would `Config` or `Spec.Config` be better? I think `Config` woud work better here.

gantony: Actually SecurityEventWebhookConfigVarSource` is the name of the struct/type used in the code, so that's probably OK. I understand the logic behind it, makes sense, ignore my comment.

calico-enterprise/reference/resources/webhook.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+A selector consists of a key, comparator, and value. Keys and values
+may be identifiers consisting of alphanumerics and underscores (`_`)
+with the first character being alphabetic or an underscore, or may be
+quoted strings. Values may also be integer or floating point numbers.
+Comparators may be `=` (equal), `!=` (not equal), `<` (less than),
+`<=` (less than or equal), `>` (greater than), or `>=` (greater than
+or equal).
+
+Keys must be indexed fields in their corresponding data set. See the
```
gantony: This paragraph makes no sense here (I assume it was taken from https://docs.tigera.io/calico-enterprise/3.16/reference/resources/globalalert#query). In this case the only index we deal with is the events index (there is no appendix in this page :wink:). The source of truth for the fields we validate is in https://github.com/tigera/calico-private/blob/master/libcalico-go/lib/validator/v3/query/validate_events.go#L12 but I don't think we want to go in that level of details here.

calico-enterprise/reference/resources/webhook.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+The query language is composed of any number of selectors, combined
+with boolean expressions (`AND`, `OR`, and `NOT`), set expressions
+(`IN` and `NOTIN`) and bracketed subexpressions. These are translated
+by {{prodname}} to Elastic DSL queries that are executed on the backend.
+
+Set expressions support wildcard operators asterisk (`*`) and question mark (`?`).
+The asterisk sign matches zero or more characters and the question mark matches a single character.
+Set values can be embedded into the query string or reference the values
+in the global alert substitution list.
```
gantony: The "global alert substitution list" makes no sense in the context of this page.

calico-enterprise/reference/resources/webhook.mdx
bartolini: Yeah, the Query section was copied. I have missed a couple of references - removed now.

calico-enterprise/reference/resources/webhook.mdx
bartolini: Removed now. Thanks @gantony!

calico-enterprise/reference/resources/webhook.mdx
```diff
@@ -0,0 +1,146 @@
+---
+description: API for this Calico Enterprise resource.
+---
+
+# Security Event Webhook
```
ctauchen:
```suggestion
# Security event webhook
```
Global change: Let's call this a security event webhook, all lowercase.

calico-enterprise/reference/resources/webhook.mdx
```diff
@@ -0,0 +1,146 @@
...
+
+### Metadata
+
+| Field | Description                                               | Accepted Values                                     | Schema |
+| ----- | --------------------------------------------------------- | --------------------------------------------------- | ------ |
+| name  | Unique name to describe this resource instance. Required. | Alphanumeric string with optional `.`, `_`, or `-`. | string |
+
+#### Annotations
+
+SecurityEventWebhooks provide an easy way to add arbitrary data to the webhook generated HTTP payload through the metadata annotation.
```
ctauchen:
```suggestion
Security event webhooks provide an easy way to add arbitrary data to the webhook generated HTTP payload through the metadata annotation.
```
Like this when speaking about the idea. If you're talking about the object, use code font.

calico-enterprise/reference/resources/webhook.mdx
ctauchen: OK if this is standard in other docs.

ctauchen (commented):
A few comments. One significant change: our convention is to name the file the same name as the resource: webhook.mdx --> securityeventwebhook.mdx. Remember to change the sidebar id accordingly.

When will this go in other products? I'd like to make sure I have a note somewhere.

@bartolini

calico-enterprise/reference/resources/webhook.mdx
bartolini: Corrected.

calico-enterprise/reference/resources/webhook.mdx
bartolini: Corrected.

calico-enterprise/reference/resources/webhook.mdx
bartolini: Corrected.

bartolini:
@ctauchen all done - thank you for the review.

> When will this go in other products? I'd like to make sure I have a note somewhere.

This will be Calico Enterprise only for the 3.19 release.
CC will be next, I don't know the exact time schedule at this point.

ctauchen:
Thanks, @bartolini. LGTM and merging.

PR #1367, opened 2024-03-12
DOCS-2124: Moving third-party licenses to static hosting for OSS

ctauchen:


ctauchen:
@ronanc-tigera I'm getting some strange behavior on this one. I'm changing these large license files so that they're just static html in our static directory. I've changed the sidebars to link to the URL. But when you click on the link from the sidebars (in Reference/Attributions), it returns a 404. But then, if you refresh the page, the correct licenses page loads just fine. 

Is this something you've seen before? The only thing I can think of at this point is to use the full URL (https://docs.tigera.io/...) and merge, hope that that resolves. Not at all sure why it would.

PR #1395, opened 2024-03-28
Enterprise: Drop prometheus operator patch and pull secret

tmjd:
With https://github.com/tigera/calico-private/pull/7403 we can stop patching the prometheus operator deployment and the creation of the tigera-prometheus pull secret is already created by the operator so it can be removed too.

In Calico Cloud we have been depending on this behavior (for quite a while) of the operator creating the pull secret in the tigera-prometheus namespace and it has worked quite well. Generally that pull secret is only needed for the prometheus-service image (not part of the prometheus operator deployment) but even if using private images (like in testing) the operator creates the pull secret as soon as the tigera-prometheus namespace is created and prometheus CRDs so it will be present so even if it is needed for the prometheus-operator the pull secret will be there.

Product Version(s):
Calico Enterprise

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/_includes/components/InstallAKS.js
+++ b/calico-enterprise/_includes/components/InstallAKS.js
@@ -45,17 +45,6 @@ export default function InstallAKS(props) {
     --type=kubernetes.io/dockerconfigjson -n tigera-operator \\
     --from-file=.dockerconfigjson=<path/to/pull/secret>`}
           </CodeBlock>
-          <p>
-            For the Prometheus operator, create the pull secret in the <code>tigera-prometheus</code> namespace and then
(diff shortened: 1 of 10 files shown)


ctauchen:
LGTM. Happy to merge?

rene-dekker (approved):
lgtm

tmjd:
Seems like we might need a tweak to this for MKE. So lets hold off on merging it for now.

tmjd:
This has been addressed with a tweak to the manifest so the changes here are good to go.

PR #1407, opened 2024-04-04
Updates CC support page to give clear path for both paid and trial users

ctauchen:


ctauchen:
@tmjd Something like this is what I had in mind.

PR #1423, opened 2024-04-12
Microsegmentation use case doc

ctauchen:
- **Adds config for use cases section**
- **Adds microsegmentation use case material**


ctauchen:
CC @mapgirll

PR #1427, opened 2024-04-12
Fix upgrade from oss using helm

rene-dekker:
In OSS, the helm installation is called `calico`.


ctauchen:
Seems okay to me, Rene. I'll leave it to you to merge.

dimitri-nicolo (approved):
lgtm

PR #1478, opened 2024-05-15
RS-1864 Add recommended default value for pods per node on AKS

peterkellydev:
Product Version(s): Calico Enterprise 3.18, 3.19, main

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


nelljerram (approved):

ctauchen (commented):
@peterkellydev This got in before I checked my notifications. A few thoughts:

* You're adding this in a requirements section, but the language makes it sound like a suggestion. Which is it? I think something like this, which uses requirements language with a qualification, may be what you're looking for: `For best performance, set the maximum pods per node to 60 or more.`
* Have the headline specify the requirement: `The maximum pods per node is set to 60 or more`.
* If it's not a do-this-or-else requirement, a hint at an explanation will help users decide how important it is. `For best performance,...` is enough, but if that's not accurate or there's another reason, consider saying that.

tmjd:
It seems odd to make this recommendation since it also works fine to add more nodes to ensure there are enough pods for all needed pods.

PR #1507, opened 2024-06-04
Clarify non-support for VXLAN/IPIP on "OpenStack"

nelljerram:
As discussed in https://github.com/projectcalico/calico/pull/7009, this non-support refers to when Calico is used as the networking plugin at the OpenStack level, but NOT to when Calico is used as the Kubernetes CNI and Kubernetes is running on OpenStack infrastructure.  This change clarifies that.


ctauchen:
LGTM, thanks! Ready to merge?

nelljerram:
Yes, thanks @ctauchen .

PR #1522, opened 2024-06-17
Update IA API reference to latest release 1.17.7 and fix api generation

Brian-McM:
IA API reference docs are quite out of date, mainly because generating the API docs has been broken.

I've changed the generation logic to not use custom forks anymore, and to use a much newer version of github.com/ahmetb/gen-crd-api-reference-docs The latest one is compatible with the latest go versions, and this new process (well the process outlined in the repo itself) doesn't require modifying the go.mod file.

We now check out the gen repo, build it, check out the ia repo, then run the gen tool in the operator repo.

This commit also omits internal packages from IA using the config.json in this repo. The template is also stored in this repo in the crd-gen-template, so any modifications to the generated code structure can be handled there.

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,17 +20,18 @@
 ##         - Outputs all of the branch-related targets that you can
 ##           use to update a specific branch's operator docs.
 
-GO_BUILD_VER?=v0.87
+GO_BUILD_VER?=v0.91
 CALICO_BUILD?=calico/go-build:$(GO_BUILD_VER)
 LOCAL_USER_ID?=$(shell id -u $$USER)
 PACKAGE_NAME?=github.com/projectcalico/calico/calico
 API_GEN_REPO?=tmjd/gen-crd-api-reference-docs
 API_GEN_BRANCH?=kb_v2
+API_GEN_VERSION=v0.3.0
 OPERATOR_REPO?=tigera/operator
 PRODUCT?=calico
 
 IA_OPERATOR_REPO?=tigera/image-assurance
-IA_OPERATOR_VERSION?=v1.7.3
+IA_OPERATOR_VERSION?=v1.17.7
 
 NODE_VER=20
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -162,14 +163,12 @@ build-ia-operator-reference:
 			--env SSH_AUTH_SOCK=/ssh-agent\
 			-w /go/src/$(PACKAGE_NAME) \
 			$(CALICO_BUILD) /bin/bash -c '$(GIT_CONFIG_SSH) rm -rf builder && mkdir builder && cd builder && \
-				git clone --depth=1 -b ia_kb_v2 https://github.com/Brian-McM/gen-crd-api-reference-docs api-gen && cd api-gen && \
-				go mod edit -replace github.com/tigera/image-assurance=github.com/$(IA_OPERATOR_REPO)@$(IA_OPERATOR_VERSION) && \
-				go mod tidy && go mod download all && go build && \
-				./gen-crd-api-reference-docs \
-					-api-dir github.com/tigera/image-assurance/operator/api \
+				git clone --depth=1 -b $(API_GEN_VERSION) git@github.com:ahmetb/gen-crd-api-reference-docs.git && cd gen-crd-api-reference-docs && go build && \
```
stevegaossou, line 166: I understand which we need a separate section here just for IA operator, but it feels like we should at least reconcile with the one that is being used for the Operator: 

- https://github.com/tigera/docs/blob/main/Makefile#L27-L28
- https://github.com/tigera/docs/blob/main/Makefile#L141

stevegaossou, line 166: This comment is not within scope for this PR, but I wonder if we should consider switching off of `gen-crd-api-reference-docs` to one of the alternatives: 
1. Kubernetes reference-docs generator
2. crd-ref-docs by Elastic

It's suggested here on the repo, since the author doesn't have enough time to regularly maintain it: 
- https://github.com/ahmetb/gen-crd-api-reference-docs?tab=readme-ov-file#alternatives

The first alternative is apparently the version of the tool used in the official Kubernetes reference docs.

crd-gen-template/members.tpl
```diff
@@ -0,0 +1,48 @@
...
+        <table>
+            {{ template "members" .Type }}
+        </table>
+    {{ end }}
+    </td>
+</tr>
+{{ end }}
+{{ end }}
+
+{{ end }}
```
stevegaossou, line 48: Question: 
- Why did we need to use custom/modified templates? 
    - Does the IA reference doc format need to differ from the one used by Tigera operator?

stevegaossou (commented):
Added a couple of comments.

crd-gen-template/members.tpl
Brian-McM: We always used customer templates, they were just committed to the individual forked repos that casey, erik, and I had. So this is the same template we were using previously (well updated because the base repo was updated).

Makefile
Brian-McM: We could consider that, but I'm just trying to fix what we have and use already.

Brian-McM: I'll leave that up to @tmjd since we were all using custom forks previously. Likely we should just use the base repo for everything, but I'm not invested into updating the tigera operator API docs right now for cloud.

stevegaossou: > We could consider that, but I'm just trying to fix what we have and use already.

Yep, it's a bigger conversation. Probably only worth it when there is a need to change things (e.g. if `gen-crd-api-reference-docs` gets too out of date and one of the alternatives has a build feature we need).

stevegaossou (approved):

WilliamTigera:
One of the preview docs looks correct, but this one still has the management cluster components.
https://deploy-preview-1522--tigera.netlify.app/calico-cloud/reference/installation/ia-api

https://deploy-preview-1522--calico-docs-preview-next.netlify.app/calico-cloud/next/reference/installation/ia-api (Correct)

Is that just a preview issue?

PR #1527, opened 2024-06-19
add id to the code-block for openshift manifests install instruction

sujeet-kr:
The `code` block for OCP manifests need a uniquely identifiable `attr` in the page for installation automation to work. We used to use the `text` earlier but because of the change in the formatting of the line above the codeblock (adding `<!---`) RegEx does not work and does not successfully search the line in the HTML. This change adds an id to the `code` tag.

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


radTuti (changes requested):
I think prefixing this with some value so it more recognisable why it is set e.g. `data_installOCPManifests` or something similar

otherwise lgtm

radTuti (approved):

PR #1540, opened 2024-07-04
This PR Fixes BGP reference resources and some Openshift steps.

frozenprocess:
This PR fixes a couple of problems with BGP reference pages and fixes a step in Openshift documentation.


calico-enterprise/_includes/components/InstallOpenShift.js
```diff
@@ -57,9 +57,19 @@ export default function InstallOpenShift(props) {
...
 
+      <Admonition type='note'>
+        <p>By default openshift-installer creates 3 replicas, you can change these settings by modifying the cloud-provider part in the install-config.yaml</p>
+        <p>The following example changes the default deployment instance type and replica quantity.</p>
+        <CodeBlock language='yaml'>
+        platform: 
+          aws:
+            type: m5.xlarge
+        replicas: 2
+        </CodeBlock>
```
ctauchen: This is currently being processed without line breaks or indentation. I'd suggest you do something like this:

```suggestion
        <CodeBlock language='yaml' title="install-config.yaml">
        {`...
platform: 
  aws:
    type: m5.xlarge
replicas: 2
...`}
        </CodeBlock>
```
(I think that's the right way, but haven't built it myself.)

This also adds a title and ellipses to make it clear that this is a snippet from that YAML. 

Would it be easier to see where you are by adding the next level up from `platform`?

calico-enterprise/reference/resources/bgpconfig.mdx
```diff
@@ -66,6 +66,7 @@ spec:
...
+| ignoredInterfaces      | List of network interfaces to be excluded when reading device routes.                                                                                                                                                                                                                                                                                                                                                          | A list of network interface names. The names can contain the wildcard character asterisk `*` to specify groups of interface names. | List of string                                        | `nil` (no extra interfaces to be ignored)                       |
```
ctauchen: List of string --> List of strings?

ctauchen (commented):
@frozenprocess Thanks! A few comments, to be applied globally.

Would be nice for someone else who knows the subject to give it a quick look over. Can you nominate someone?

calico-enterprise/_includes/components/InstallOpenShift.js
frozenprocess: Easier for user? yes
But it will be more difficult to keep it uptodate for us too.

calico-enterprise/reference/resources/bgpconfig.mdx
frozenprocess: Well, my initial thought was same as you but then I saw this one and thought maybe there is a merit to "string"
https://github.com/tigera/docs/blob/ddc5f485f7269a28018c2e10c1da34a9a92bd8d7/calico/reference/resources/bgpconfig.mdx?plain=1#L66
I'm happy to change it to `strings` given that, it is the correct "word" for this scenario in my opinion as well.

ctauchen:
LGTM

PR #1546, opened 2024-07-08
DOCS-2215: Publishes CC 19.4.0

ctauchen:


calico-cloud_versioned_docs/version-19-2/variables.js
```diff
@@ -6,17 +6,17 @@ const variables = {
   prodnamedash: 'calico-cloud',
   baseUrl: '/calico-cloud',
   filesUrl: 'https://docs.calicocloud.io',
-  filesUrl_CE: 'https://downloads.tigera.io/ee/v3.19.0-2.0',
+  filesUrl_CE: 'https://downloads.tigera.io/ee/v3.19.0-2.10',
```
tmjd: I don't think this should be changing. It looks similar to the 'cloudversion' but it isn't the same.

ctauchen: Reverted.

tmjd (approved):
LGTM

PR #1549, opened 2024-07-09
Fixes lookback for IA global alerts examples

ctauchen:
Adds `lookback` line to IA global alerts example.
Fixes description for `lookback` in reference


ctauchen:
PTAL @davselliTigera

PR #1563, opened 2024-07-23
DOCS-2265: Fixes bug with attributions page for OSS 3.28. Updates attributions."

ctauchen:
The attributions link from the sidebar went to the vNext placeholder rather than to the correct OSS 3.28 attributions page.


ctauchen:
CC @lwr20 I only discovered this today. It's a process error on my part, something that must change in the sidebars file as part of the version cutting process. I plan to add a fix to the version cutting script.

PR #1602, opened 2024-08-15
nftables getting-started guide

caseydavenport:
Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico/getting-started/kubernetes/nftables.mdx
+++ b/calico/getting-started/kubernetes/nftables.mdx
@@ -1,0 +1,193 @@
+---
+description: Install Calico using the tech-preview nftables data plane.
+---
+
+# Calico nftables data plane
+
+## Big picture
+
+Try the tech-preview {{prodname}} nftables data plane.
+
+## Value
+
+Kubernetes introduced a beta status [kube-proxy Service implementation based on nftables](https://kubernetes.io/docs/reference/networking/virtual-ips/#proxy-mode-nftables)
+in Kubernetes v1.31 which promises better performance than the default iptables implementation.
+Installing {{prodname}} in nftables mode provides a networking and network policy implementation that is compatible with the upstream nftables kube-proxy.
+
+## Before you begin
+
+**Required**
+
+- A Linux host that meets the following requirements:
+
+  - x86-64, arm64, ppc64le, or s390x processor
+  - 2CPU
+  - 2GB RAM
+  - 10GB free disk space
+  - Linux kernel version 5.13 or later with `nft` >= 1.0.1
+
+- {{prodname}} can manage `cali` and `tunl` interfaces on the host
+
+  If NetworkManager is present on the host, see [Configure NetworkManager](../../operations/troubleshoot/troubleshooting.mdx#configure-networkmanager).
+
+:::note
+
+{{prodname}} nftables dataplane is currently in technical preview and may not be suitable for production use. Please report any issues you experience
+while using this feature.
+
+:::
+
+## How to
+
+### Create a single-host Kubernetes cluster with the nftables kube-proxy enabled.
+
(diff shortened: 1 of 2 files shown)


sidebars-calico.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
             'getting-started/kubernetes/requirements',
             'getting-started/kubernetes/community-tested',
             'getting-started/kubernetes/quickstart',
+            'getting-started/kubernetes/nftables',
```
ctauchen: There's a larger conversation about how and where we document dataplane guides like this. But for now it makes most sense to group this further down next to the VPP guide.

calico/getting-started/kubernetes/nftables.mdx
```diff
@@ -0,0 +1,193 @@
+---
+description: Install Calico using the tech-preview nftables data plane.
```
ctauchen: I can see this easily being missed when it goes GA. Can we scrap the tech preview language here?

calico/getting-started/kubernetes/nftables.mdx
```diff
@@ -0,0 +1,193 @@
...
+---
+
+# Calico nftables data plane
+
```
ctauchen, line 9:
```suggestion
:::note

This feature is tech preview. Tech preview features may be subject to significant changes before they become GA.

:::
```
We're using this boilerplate for page-level tech preview notices.

calico/getting-started/kubernetes/nftables.mdx
```diff
@@ -0,0 +1,193 @@
...
+- {{prodname}} can manage `cali` and `tunl` interfaces on the host
+
+  If NetworkManager is present on the host, see [Configure NetworkManager](../../operations/troubleshoot/troubleshooting.mdx#configure-networkmanager).
+
+:::note
+
+{{prodname}} nftables dataplane is currently in technical preview and may not be suitable for production use. Please report any issues you experience
+while using this feature.
+
+:::
```
ctauchen: Remove, in favor of boilerplate above.

calico/getting-started/kubernetes/nftables.mdx
```diff
@@ -0,0 +1,193 @@
...
+   After installing kubeadm, do not power down or restart
+   the host. Instead, continue directly to the next step.
+
+   :::
+
+1. As a regular user with sudo privileges, open a terminal on the host that you installed kubeadm on.
+
+1. Create a kubeadm configuration file that enables the nftables kube-proxy mode.
+
+   ```
```
ctauchen:
```suggestion
   ```bash
```
Globally, for syntax highlighting.

ctauchen (commented):
A few suggestions for you. This is a good doc to get folks up and running with the new dataplane. It would be worth discussing how we ought to be presenting dataplane information generally in the docs. For both eBPF and VPP, we're relying on bespoke guides, not unlike this one, that were written to present the tech preview feature. But we should see how we can integrate these procedures as part of a standard installation procedure.

caseydavenport:
@ctauchen thanks for the review! I think I've addressed your comments.

> But we should see how we can integrate these procedures as part of a standard installation procedure.

Definitely agree, but I think it's larger than this PR.

It might be nice to have data plane selection more seamless integrated into the other install docs rather than tucked away in its own set. The only concern I see is complicating what would otherwise be a pretty straightforward and opinionated guide with information that isn't relevant to every user.

ctauchen:
> It might be nice to have data plane selection more seamless integrated into the other install docs rather than tucked away in its own set. The only concern I see is complicating what would otherwise be a pretty straightforward and opinionated guide with information that isn't relevant to every user.

Yes, I think it's all worth doing, but it will have to be part of a much larger effort. 

For now, do you want to shop this doc around a bit, or are you happy to merge? @caseydavenport

caseydavenport:
I think I'm good to merge this :+1:

PR #1607, opened 2024-08-20
Updating to node v20.16.0

ctauchen:
During D3 upgrade testing I was unable to get a Netlify build because of a node dependency: https://app.netlify.com/sites/tigera/deploys/66c459730829770008b4ade8. 

We're currently on node 18, which is the minimum required for Docusaurus 3. Node 18 will be EOL on 2025-04-30.

This updates our build to node v20.16.0.


ctauchen:
@ronanc-tigera @doucol 

This builds fine, but is it a good idea? Are there reasons to stick with 18?

doucol:
> @ronanc-tigera @doucol
> 
> This builds fine, but is it a good idea? Are there reasons to stick with 18?

Should be fine, and possibly recommended to do the upgrade. would like to hear what Ronan thinks though

ronanc-tigera:
If you're upgrading Node, you should also update the package.json with
```json
  "engines": {
    "node": ">=20.0"
  }
```
Assuming you ran & tested locally as well?

ctauchen:
> ```json
>   "engines": {
>     "node": ">=20.0"
>   }
> ```

Thanks, I've added that and updated the readme. 

Local build and tests pass just fine.

ronanc-tigera (approved):

PR #1609, opened 2024-08-20
DOCS-2019: Undo global availability for versioned components

ctauchen:
Removes the convoluted versioning logic in `src/theme/MDXComponents.js`. This backlog item sat for some time. It may be causing trouble with our D3 builds, so I figured it was worth getting out of the way.

From now on, you must import versioned components directly in your MDX files.


ctauchen:
@ronanc-tigera This has been on my to-do list for some time. I think our current D3 issues with `releases` not being initialized may have something to do with this convoluted mechanism, which I've been wanting to do away with anyway.

I'd appreciate it if you could double-check that this approach is sound. 

Background: this is a port of the functionality that was built into the old docs sites. 22 components are included here. The idea is that we can call a versioned component without getting caught up in versioning imports.

But it's clunky, and has caused problems for me for CE releases after we started our ep1/2 structure.

Once I'm happy this approach works, I'll continue with the remaining components.

src/theme/MDXComponents.js
```diff
@@ -124,10 +123,6 @@ function getCalicoVersionedComponent(version, componentName) {
       return require(`../../calico_versioned_docs/version-${version}/_includes/components/PodCidrSed`).default;
     case 'ReqsKernel':
       return require(`../../calico_versioned_docs/version-${version}/_includes/components/ReqsKernel`).default;
-    case 'ReqsSys':
-      return require(`../../calico_versioned_docs/version-${version}/_includes/components/ReqsSys`).default;
-    default:
```
ronanc-tigera, line 129: did you mean to remove the `default` case?

ctauchen: Nope, thanks! Fixed.

ctauchen:
>**Note**: Netlify doesn't like it, even though it builds and tests just fine on my machine. Troubleshooting likely to wait until next week.

@ronanc-tigera This could do with a second pair of eyes. Realistically, verifying the output manually isn't going to happen with so many changes. I've done some spot checks, which looked fine. However, mistakes here will have a significant impact on the docs, so would be good to make sure we get this right.

A thought: perhaps you could test the visual regression tooling with this PR? WDYT?

No special rush for review, we can dig into this next week.

CC @doucol

calico-enterprise/getting-started/install-on-clusters/openshift/installation.mdx
```diff
@@ -2,9 +2,12 @@
 description: Install Calico Enterprise on an OpenShift 4 cluster.
 ---
 
+import InstallOpenShift from '@site/calico-enterprise/_includes/components/InstallOpenShift';
+import InstallOpenshiftBeforeYouBegin from '@site/calico-enterprise/_includes/components/InstallOpenshiftBeforeYouBegin';
+
 # Install Calico Enterprise on OpenShift
 
-import InstallOpenShiftManifests from '@site/calico-enterprise/_includes/components/InstallOpenShiftManifests.js';
+import InstallOpenShiftManifests from '@site/calico-enterprise/_includes/components/InstallOpenShiftManifests';
```
ronanc-tigera: Could you group all imports together? Thanks 🙏

ronanc-tigera (approved):

calico-enterprise/getting-started/install-on-clusters/openshift/installation.mdx
ctauchen: Nice spot, thanks! Missed this one.

PR #1618, opened 2024-08-28
DOCS-2271: CC install overhaul (system requirements and pre-install checks)

ctauchen:
This pull request is the second part of a set of revisions to the Calico Cloud install book.

Major changes:
* ~moves all files to flat structure~ (See https://github.com/tigera/docs/pull/1621)
* rewrites system-requirments.mdx
* adds prepare-cluster.mdx, populates it with reorganized copy from system-requirements.mdx, aks.mdx, gke.mdx, rke2.mdx
* ~removes aks.mdx, gke.mdx, rke2.mdx. (Currently mixed up in git, shows as renamed only, will sort later)~ (See https://github.com/tigera/docs/pull/1621)

Also:
* Fixes markdown file links broken by earlier rearrangements
* Adds items to prose linter

Reviewers PTAL:
* https://deploy-preview-1618--calico-docs-preview-next.netlify.app/calico-cloud/next/get-started/system-requirements
* https://deploy-preview-1618--calico-docs-preview-next.netlify.app/calico-cloud/next/get-started/prepare-cluster

Items outstanding:
* Reconciler business
* clarify RKE2 `cni: none`

NEXT PART:
* Changes to install procedures.


calico-cloud/get-started/system-requirements.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+## Kubernetes distributions and CNIs
+
+Calico Cloud works with Kubernetes on self-provisioned infrastructure and on managed Kubernetes distributions.
+To use Calico Cloud for both networking and network policy, your cluster must have Calico Open Source installed before you connect to Calico Cloud.
+For most managed distributions, you can use the provider's CNI for networking and use Calico Cloud for network policy.
+
+| Distribution | CNI |
+| --- | --- |
+| Kubernetes on self-provisioned infrastructure | - Calico Open Source 3.26 and 3.27 |
```
ctauchen: Previously, we said you could do this with OSS 3.20+. It seems inconsistent to direct people to old and unsupported versions of our product. I suggest we stick with N-2 on the low side, which is what we support for documentation.

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -0,0 +1,47 @@
...
+## Kubernetes versions
+
+* Your Kubernetes distribution must be based on Kubernetes 1.25 to 1.29
+
+## Architectures
+
+* Calico Cloud can be installed on nodes based on x86-64 and ARM64 architectures.
+  :::note
+  For the GKE managed Kubernetes service, Calico Cloud supports only the x86-64 architecture.
+  :::
```
RoryDoherty: This is no longer needed, we've resolved the issue with GKE ARM clusters now

calico-cloud/get-started/prepare-cluster.mdx
```diff
@@ -0,0 +1,209 @@
...
+```bash
+kubectl get <type> -n <namespace> <resource-name> -o yaml | grep ' addonmanager.kubernetes.io/mode:'
+```
+
+
+## Optional: Verify settings for your managed Kubernetes service
+
+Although {{prodname}} checks that your cluster meets [System requirements](system-requirements.mdx), verifying now can save you time troubleshooting.
+
+### Verify settings for EKS
```
ctauchen: @tmjd This is new. Surely the same checks would apply to EKS as to the other platforms? For IAM permissions, probably some config verification? Anything else to add here?

We've had customers think that EKS is not supported because this was missing.

apardeshi-tigera:
@ctauchen 
Comparing to the existing instruction on system requirements, I think these are much more clearer. 
Some of the things I wanted to touch on:
1. This line "Your cluster must already have a CNI installed before you can connect to Calico Cloud." talks about the CNI requirement. You think  moving this to CNI specific section makes sense? Reason being CNI is one of the set of requirements.
2. For _Distribution - CNI_ table, can we make it more explicit to convey that user  does not actually need both the CNIs. And that user only needs to have 1 based on the use case. I know we specify it in the paragraph above but can we also incorporate that message within the table?

calico-cloud/get-started/system-requirements.mdx
ctauchen: Thanks, I've removed the note.

ctauchen:
> @ctauchen Comparing to the existing instruction on system requirements, I think these are much more clearer. Some of the things I wanted to touch on:
> 
> 1. This line "Your cluster must already have a CNI installed before you can connect to Calico Cloud." talks about the CNI requirement. You think  moving this to CNI specific section makes sense? Reason being CNI is one of the set of requirements.
> 2. For _Distribution - CNI_ table, can we make it more explicit to convey that user  does not actually need both the CNIs. And that user only needs to have 1 based on the use case. I know we specify it in the paragraph above but can we also incorporate that message within the table?

Thanks for taking a look at this, @apardeshi-tigera. 

1) I wanted to make that clear from the very start because this differs from the expected pattern from OSS (and from any CNI I can think of, really). Technically you're right. But I think it's worth making this prominent even at the expense of some redundancy.
2) Thanks, I can see that was not clear. Changed the column head to "Supported CNIs", which gets that idea across.

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -1,62 +1,50 @@
...
-- `tigera-operator` deployment in the `tigera-operator` namespace (Installation on AKS may still work if the mode is Reconcile, see [AKS](aks.mdx#verify-addonmanager-is-compatible) for an additional check.)
-- `calico-node` daemonset in the `kube-system` namespace
+## Kubernetes versions
 
-If the following command finds addonmanager on either of the resources, then Addon-manager is being used. Find a different cluster to use.
+Your Kubernetes distribution must be based on one of the following Kubernetes versions:
+* Kubernetes 1.29
+* Kubernetes 1.28
+* Kubernetes 1.27
+* Kubernetes 1.26
```
sujeet-kr:
```suggestion
```

sujeet-kr: Since our k8s support is n-2, this is due to the k8s libraries that we use. We usually use an n-1 version of those libraries in our product and per k8s official support (of +- 1) can only call support for n-2.

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -1,62 +1,50 @@
...
-Pods running in your Kubernetes cluster must allow outbound traffic to the following endpoints:
+## Browser support for the Manager UI web console
 
-- `https://installer.calicocloud.io:443/*`
-- `https://www.calicocloud.io:443/api/*`
-- `https://client-auth.calicocloud.io:443/*`
-- TCP to `<management-cluster>.calicocloud.io:9000`
+To access the Manager UI web console, you can use latest two versions of the following web browsers:
+* Chrome
+* Firefox
```
sujeet-kr: IIUC we do not test in Firefox at this time but there is work in progress to get this capability added to our automated tests.

calico-cloud/get-started/system-requirements.mdx
tmjd: We weren't directing people to use old versions but saying if they had those they could be connected. Stating those old versions was in the spirit of allowing upgrades for clusters that had old installations. We don't want to turn away  users because their clusters are running old versions of Calico.

It is a difference of expecting people to set up new clusters to connect to Calico Cloud vs connecting clusters they already have, I feel like you're writing the docs from the Calico or Enterprise point of setting up a new cluster but Calico Cloud is around being able to connect existing clusters.

calico-cloud/get-started/system-requirements.mdx
tmjd: These versions with the v20.0.0 release need to be updated to 1.27-1.30

ctauchen: N-2 would put us supporting 1.29-1.31, no? https://kubernetes.io/releases/
@tmjd Does that sound right?

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -1,62 +1,50 @@
...
+For most managed distributions, you can use the provider's CNI for networking and use Calico Cloud for network policy.
 
-### Step 2: Verify cluster is not managed by a Kubernetes reconciler
+| Distribution | Supported CNIs |
+| --- | --- |
+| Kubernetes on self-provisioned infrastructure | - Calico Open Source 3.26 and 3.27 |
+| Amazon Elastic Kubernetes Service | - Calico Open Source 3.26 and 3.27<br/> - Amazon VPC CNI |
+| Azure Kubernetes Service | - Calico Open Source 3.26 and 3.27<br/> - Azure CNI |
+| Google Kubernetes Engine | - Calico Open Source 3.26 and 3.27<br/> - GKE CNI |
+| Rancher Kubernetes Engine 2 | - Calico Open Source 3.26 and 3.27 |
```
tmjd: I believe each of AKS, GKE, and RKE2 have configurations that cannot be upgraded.
I think it is incomplete to leave the caveats to the "Prepare your cluster for Calico Cloud" section.

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -1,62 +1,50 @@
...
-You won't be able to use {{prodname}} if other elements of your infrastructure, such as firewalls or security groups, restrict access to these endpoints.
-
-### Step 4: Verify platform-specific requirements
-
-If you using one of the following platforms, review further requirements. Otherwise, you are ready to connect your cluster in Manager UI: go to Managed clusters, and click **Connect**.
-
-- [AKS](aks.mdx)
-- [GKE](gke.mdx)
-- [RKE2](rke2.mdx)
+* [Prepare your cluster for Calico Cloud](prepare-cluster.mdx)
```
tmjd, line 148: You've put many of the requirements under "Prepare your cluster for Calico Cloud", the title suggests if you've made it that far you can connect your cluster you might just need to make some changes which is not the case.

I think all of the items that are now in this new section that were previously in "Verify platform-specific requirements" are all things that in the majority of the clusters cannot be modified.

calico-cloud/get-started/prepare-cluster.mdx
tmjd: Something like that is reasonable  but I don't think the link you've found is around permissions in an EKS cluster but permissions in the AWS console accessing EKS features.

calico-cloud/get-started/system-requirements.mdx
tmjd: No that does not sound right. Enterprise v3.20.0-EP1 supports a max of k8s 1.30 so that is the maximum version CC will support at this time (with v20.0.0)

calico-cloud/get-started/system-requirements.mdx
ctauchen: > configurations that cannot be upgraded

What do you mean here, exactly? That you can't upgrade CC with configurations intact?

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -1,62 +1,101 @@
...
-- `https://installer.calicocloud.io:443/*`
-- `https://www.calicocloud.io:443/api/*`
-- `https://client-auth.calicocloud.io:443/*`
-- TCP to `<management-cluster>.calicocloud.io:9000`
+To access the Manager UI web console, you can use latest two versions of the following web browsers:
+* Chrome
+* Safari
 
-For each node, Docker must be able to pull images from the following endpoints:
+## Provision about reconcilers //TODO
```
ctauchen: @tmjd From my research, it looks like addon-manager isn't really used at all anymore. Does that sound right to you? 

I think we should abstract this idea and provide realistic examples of reconcilers that people might actually be using. Calico might be managed in a Helm chart, for example.

If the reconcilers are generally things that can be altered, like a Helm chart, then we might want to move this to Prepare your cluster and provide guidance as follows:

If you've got something that's reconciling Calico resources, you need to get them to stop.

Would be good to know: what reconcilers we're talking about, that we can usually turn it off and connect successfully to CC, and any situations we know about where there's no way around this (such as addon-manager seemed to be).

calico-cloud/get-started/system-requirements.mdx
ctauchen: Got it, I've divvied them up based on whether they can be fixed (Prepare your cluster) or not (System requirements).

calico-cloud/get-started/system-requirements.mdx
```diff
@@ -1,62 +1,101 @@
...
+  ```bash
+  gcloud container clusters describe <cluster-name> --flatten addonsConfig.networkPolicyConfig.disabled
+  ```
+* Your cluster's Dataplane V2 is set to `null`.
+  To verify, run the following command:
+  ```bash
+  gcloud container clusters describe <cluster-name> --flatten networkConfig.datapathProvider
+  ```
+
+### Rancher Kubernetes Engine 2 //TODO
```
ctauchen: FYI @tmjd

apardeshi-tigera:
@ctauchen For this requirement in GKE: 
"Your cluster's Dataplane V2 is set to null. To verify, run the following command:"
I found that the response can also be Legacy data path. And that will still work. Legacy data path simply means that cluster is using V1 dataplane which Calico.

ctauchen:
> @ctauchen For this requirement in GKE: "Your cluster's Dataplane V2 is set to null. To verify, run the following command:" I found that the response can also be Legacy data path. And that will still work. Legacy data path simply means that cluster is using V1 dataplane which Calico.

Thanks, I noticed that too while reading the GKE docs, but I assumed there was some other reason for the null requirement. @tmjd any ideas here?

tmjd:
> And that will still work.

@apardeshi-tigera were you able to connect a cluster like that to Calico Cloud? or upgrade it to Enterprise?

> Legacy data path simply means that cluster is using V1 dataplane which Calico.

I think that is not a supported configuration because you can't update Calico components on a cluster with Calico managed by GKE.

calico-cloud/get-started/system-requirements.mdx
tmjd: I mean you can't install CC depending on the options or the CNI being used in the cluster.
When I said "upgraded" I meant the cluster can't have CC installed onto it.

calico-cloud/get-started/system-requirements.mdx
tmjd: >it looks like addon-manager isn't really used at all anymore. Does that sound right to you?

That does not sound right. AKS definitely still uses addon-manager. I am looking at a cluster I just created on AKS (with manged Calico) and I still see the labels that Addon-manager uses.

>If the reconcilers are generally things that can be altered, 

It depends on who is managing the reconciler generally, like if GKE or AKS is managing the reconciler then the user isn't going to be able to disable management. In RKE2 I'm not sure how easy it is to make RKE2 stop managing Calico if it originally installed it, I don't know if it is an easy switch to flip.

> Would be good to know: what reconcilers we're talking about

We're talking about any that can manage Kubernetes resources, the user can't have any that are managing Calico.

ctauchen:
FYI: Merging organizational changes, left unresolved items as-is until after the release.

apardeshi-tigera:
> > And that will still work.
> 
> @apardeshi-tigera were you able to connect a cluster like that to Calico Cloud? or upgrade it to Enterprise?
> 
> > Legacy data path simply means that cluster is using V1 dataplane which Calico.
> 
> I think that is not a supported configuration because you can't update Calico components on a cluster with Calico managed by GKE.

@tmjd Yeah I was able to connect the cluster. I did not try upgrading it though.

PR #1684, opened 2024-09-27
[RS-1991] Add enable/disable detectors doc

Dean-Coakley:
Product Version(s):

Calico Cloud 20.1.0 (not yet released) 

Issue:

[RS-1991](https://tigera.atlassian.net/browse/RS-1991)

Link to docs preview:

https://deploy-preview-1684--calico-docs-preview-next.netlify.app/calico-cloud/next/threat/container-threat-detection

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed 

[RS-1991]: https://tigera.atlassian.net/browse/RS-1991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
```diff
@@ -64,12 +66,15 @@ so we can bundle the BTF data to precisely match the version of the kernel runni
...
+- [Disable detectors](#disable-detectors)
+  - [Disable detector via UI](#disable-detector-via-ui)
+  - [Disable detector via RuntimeSecurity Custom Resource](#disable-detector-via-runtimesecurity-custom-resource)
 
 ### Enable Container Threat Detection
 
 Container threat detection is disabled by default.
 
-To enable Container threat detection on your managed cluster, go to the **Threat Defence** section in the {{prodname}} UI, and select **Enable Container Threat Detection**.
+To enable Container threat detection on your managed cluster, go to the **Threat Defense** section in the {{prodname}} UI, and select **Enable Container Threat Detection**.
```
Dean-Coakley, line 76: (We use `en_us` everywhere else other than this instance)

Dean-Coakley:
Will promote to "Ready for review" after @gantony reviews content.

calico-cloud/threat/container-threat-detection.mdx
```diff
@@ -85,11 +90,79 @@ EOF
...
+
+If duplicate entries are present. The last entry will take precedence.
+
+:::
+
+- Edit the RuntimeSecurity CR:
+
+    To patch entire config:
+    ```
+    kubectl patch -n tigera-runtime-security runtimesecurities.operator.tigera.io default --type='merge' -p='{"spec":{"detectorsConfig": [{"id": "persistence-user_modification", "disabled": false}]}}'
```
nelljerram: This is what I was talking about in stand up.  I don't think this doc makes sense, if we don't provide any guidance for how the user can infer a detector ID like "persistence-user_modification".

nelljerram (changes requested):

calico-cloud/threat/container-threat-detection.mdx
Dean-Coakley: I think @gantony's suggestion of in the future generating a yaml/kubectl command should be the path forward for sure.

Are you saying the first step is not very practical/reasonable? https://github.com/tigera/docs/pull/1684/files#diff-52ad6c1409b45aadb04bb1ff6b823504bd80596585c38707317ea5d86441bae6R112-R128
If you're saying it cannot work, I'm not sure that I follow.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
```diff
@@ -85,11 +90,80 @@ EOF
 
 ### Monitor the Security Events page for malicious programs
 
-If a malicious or suspicious program is run within the cluster, it will be reported on the Security Events page of the
+If a malicious or suspicious program is run within the cluster, it will be reported on the **Security Events** page of the
 {{prodname}} UI.
 
 ![malware-alert-example](/img/calico-cloud/malware-alert-example.png)
 
+### Disable detectors
```
gantony: Would change this to `Detectors settings` to match the name of the tab, and reduce the emphasis on disabling detectors.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
```diff
@@ -23,15 +23,17 @@ Get alerts when security threats, such as malware and other suspicious processes
 As part of these threat detection capabilities, {{prodname}} maintains a database of malware file
 hashes. This database consists of SHA256, SHA1, and MD5 hashes of executable file contents that are
 known to be malicious. Whenever a program is launched in a {{prodname}} cluster, malware
-detection generates an alert in the Security Events dashboard if the program's hash matches one that is known
+detection generates an alert in the **Security Events Dashboard** if the program's hash matches one that is known
 to be malicious.
 
 Our threat detection engine also monitors activity within the containers running in your clusters to detect suspicious behavior and generate corresponding alerts. The threat detection engine monitors the following types of suspicious activity within containers:
```
gantony, line 29: Not sure how to change this (and whether now is the right time), but it would be good to emphasis that these "types of suspicious activity" map to MITRE tactics https://attack.mitre.org/tactics/enterprise/

calico-cloud/threat/container-threat-detection.mdx
nelljerram: Sorry, I somehow missed the first step before.

But I still don't see how this makes sense overall.  The section is "Disable detector via RuntimeSecurity Custom Resource" and the first step is "No, disable it in the UI instead".  So what value are we actually trying to provide here?  Perhaps we should just delete this section for now?

Perhaps the point is "You've disabled some detectors in one cluster.  Now how do you use the RS resource to deploy the same to other clusters?"  That would make sense, but in that case we should update the narrative to be clearer about that being the use case.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
```diff
@@ -85,11 +90,80 @@ EOF
...
 
-If a malicious or suspicious program is run within the cluster, it will be reported on the Security Events page of the
+If a malicious or suspicious program is run within the cluster, it will be reported on the **Security Events** page of the
 {{prodname}} UI.
 
 ![malware-alert-example](/img/calico-cloud/malware-alert-example.png)
 
+### Disable detectors
+
+All detectors are enabled by default. You may want to disable detectors that have a large performance impact and/or produce excessive false positive alerts.
```
gantony: Up to this point, a user may have no idea of what a detector is, and we want to emphasis more the "showcasing capabilities" of the feature more than disabling individual detectors.

With this in mind, I would suggest we rephrase to something like:

```
Threat detection capabilities are provided by detectors that protect against specific [MITRE techniques](https://attack.mitre.org/techniques/enterprise/). 
Detectors are grouped by [MITRE tactics](https://attack.mitre.org/tactics/enterprise/) they protect against. Additional information for each detector can be found by clicking on a detector. 
All detectors are enabled by default, with the option of disabling specific detectors if needed.

<maybe INSERT SCREENSHOT> 
```

At this stage I don't think there is much to add about `Disable detector via UI`. It's fairly self-explanatory. I've lumped it in the first section. If we feel more detail is needed, I think we should balance it with info about "view detector details". 

I would only add one sub-section for applying a configuration to multiple clusters. Maybe something along the lines of:
```
#### Configure detectors via RuntimeSecurity Custom Resource

Detector settings configured through the UI are stored in the cluster's RuntimeSecurity Custom Resource. When configuring multiple clusters, the same configuration can be applied to all clusters.

Get the reference `detectorsConfig`:

    ```
    kubectl get runtimesecurities.operator.tigera.io default -o jsonpath="{.spec.detectorsConfig}" > detectorsConfig.json
    ```

Alternatively, can enable Container Threat Detection with the required configuration in one command:
    ```
    <kubectl apply command with settings>
    ```

```

calico-cloud/threat/container-threat-detection.mdx
Dean-Coakley: My thought process was: 
The first time you do it, it's not reasonable that you'd use kubectl. So you'd disable your desired detectors via UI and then use these docs as a guide to figure out and test the commands. To make it repeatable for the rest of your clusters. - A process that we'd aim to improve in the near future.

I think it's better to leave this less-than-ideal solution documented for the time being but I'd also be ok with removing if @gantony agrees.

gantony (changes requested):
I think we need to give a bit more background about what a detector is, that the main use case it to showcase them and eventually, we may disable some if needed.

Then would limit the kubectl to applying config to multiple clusters

calico-cloud/threat/container-threat-detection.mdx
gantony: Yes, I'd be in favor of providing less details about disabling detectors and only describing kubectl commands in the context of applying the same config on multiple cluster. Tried to capture this in https://github.com/tigera/docs/pull/1684#discussion_r1781042629

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
gantony: I initially had `#### Configure detectors via RuntimeSecurity Custom Resource` as `#### Configure detectors on multiple clusters`. Still not sure which one I prefer :sweat_smile:. The later is much more use case orientated, which might be a better fit for the docs.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
Dean-Coakley: I have some unstaged changes for a following PR to update the mitre tactics at the bottom of https://docs.tigera.io/calico-cloud/threat/security-event-management#ui-help . I think it's nice that each entry references back to mitre docs on that page too.

Maybe it makes sense to reference that page from here instead as a single source of truth?

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
Dean-Coakley: Doesn't make much sense in the context of:
```
## How to

...
- [Detectors settings](#detectors-settings)
```

Going with `Update detectors settings`

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
gantony: I'm not sure. I would ignore that comment for now. I meant to revisit before pushing my review. If we mention tactics and techniques as proposed in https://github.com/tigera/docs/pull/1684#discussion_r1781042629 then I think we can leave as it for now.

I've noticed a small issue too, the tactics/categories are not order per attack phase as we do in the security events dasboard. I'll log a bug separately. I think we should rework that section of the docs only when that bug is resolved.
image

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
gantony: For the example, reading `detectorsConfig.json` from file could be a bit messy and distracting. We could just use and env var for simplicity and let the user figure out how to adapt it to their situation.

`detectorsConfig=$(kubectl get runtimesecurities.operator.tigera.io default -o jsonpath="{.spec.detectorsConfig}"\)`

calico-cloud/threat/container-threat-detection.mdx
```diff
@@ -85,11 +89,58 @@ EOF
...
+
+Detector settings configured through the UI are stored in the cluster's RuntimeSecurity Custom Resource. When configuring multiple clusters, the same configuration can be applied to all clusters.
+
+Get the reference `detectorsConfig`:
+
+```
+export detectorConfig="$(kubectl get runtimesecurities.operator.tigera.io default -o jsonpath="{.spec.detectorsConfig}")"
+```
+
+Alternatively, can enable Container Threat Detection with the required configuration in one command:
```
nelljerram: I don't think "Alternatively" is correct here.  The next line is the second step of the process, not an alternative.

Also, don't need the `export`.  (And it looks odd to use it when not needed.)

calico-cloud/threat/container-threat-detection.mdx
```diff
@@ -85,11 +89,58 @@ EOF
...
+
+```
+kubectl patch -n tigera-runtime-security runtimesecurities.operator.tigera.io default --type='merge' -p="{\"spec\":{\"detectorsConfig\": $detectorsConfig}}"
+```
+
+:::note
+
+If duplicate detector IDs are present in the RuntimeSecurity CR. The last entry will take precedence.
```
nelljerram: Full stop should be a comma, i.e. these should be a single sentence.

nelljerram (approved):
Thanks, just some very minor tweaks from me now.

Although I've approved, I think you should wait for Antony's approval too, as he is much closer to the detail here.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
```diff
@@ -85,11 +89,58 @@ EOF
...
+detectorConfig="$(kubectl get runtimesecurities.operator.tigera.io default -o jsonpath="{.spec.detectorsConfig}")"
+```
+
+Container Threat Detection can be enabled with the desired configuration in a single command:
+
+```
+kubectl patch -n tigera-runtime-security runtimesecurities.operator.tigera.io default --type='merge' -p="{\"spec\":{\"detectorsConfig\": $detectorsConfig}}"
+```
+
+:::note
```
gantony: I’m not convinced the 2 notes are needed now that we don’t encourage editing the CR, but just patching it.
Might be better to move them to the CR reference page added in #1685.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
```diff
@@ -85,11 +89,58 @@ EOF
...
+  id: persistence-user_modification
+```
+
+```
+detectorsConfig:
+- id: persistence-user_modification
+```
+
+:::
+
```
gantony: For the screenshot, it could be good to scroll up a tiny bit to see the 2 tabs in the page.
I like the size of the window, like we’re using a smaller screen resolution, looks good 👍

gantony (approved):
Looks good Dean, thanks for iterating on this.
Approving with 2 small comments, would be good to get feedback from @ctauchen too before merging.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
Dean-Coakley: Good point. Updated.

calico-cloud_versioned_docs/version-20-1/threat/container-threat-detection.mdx
Dean-Coakley: Removed for now. Feel free to add them back in once that lands.

ctauchen:
Closing here, commits pulled in as #1693 with edits.

PR #1685, opened 2024-09-30
[RS-2091] Add resource definition for runtime security resource

LorcanMcVeigh:
Add a resource definition page for the Runtime Security Resource for enabling Container Threat Detection

Product Version(s):

This resource only exists in the Calico Cloud

Issue:

[jira ticket](https://tigera.atlassian.net/browse/RS-2091)

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-cloud/reference/resources/runtimesecurity.mdx
+++ b/calico-cloud/reference/resources/runtimesecurity.mdx
@@ -46,8 +46,7 @@ spec:
 | Field                    | Description                                                                                  | Accepted Values  | Schema                                                | Default |
 | ------------------------ | -------------------------------------------------------------------------------------------- | ---------------- | ----------------------------------------------------- | ------- |
 | detectorConfig           | Configuration that allows particular threat detectors to be disabled                         |                  | [DetectorConfig](#detectorconfig)                     |         |
-| runtimeExceptionList     | List of entries of processes that are allowed to run that won't generate an event            |                  | [runtimeExceptionList](#runtimeexceptionlist)         | Enabled |
-| runtimereporterdaemonset | Daemonset spec to override the default daemonset spec for runtime reporter                   |                  | [RuntimeReporterDaemonSet](#runtimereporterdaemonset) |         |          
+| runtimeExceptionList     | List of entries of processes that are allowed to run that won't generate an event            |                  | [runtimeExceptionList](#runtimeexceptionlist)         | Enabled |        
 
 ### DetectorConfig 
 
@@ -71,21 +70,3 @@ generation of runtime reports.
 | processInvocation | The exact name or regex of the process to which a user wants to negate the generation of runtime logs                  |                 | string  |
 | pod               | The exact name or regex of the pod(s) to which a user wants to negate the generation of runtime logs                   |                 | string  |
 | namespace         | The exact name or regex of the namespace(s) for which a user wants to negate the generation of runtime logs            |                 | string  |
-
-### RuntimeReporterDaemonSetPodSpec 
-
-If any of the fields below are specified, they will override the default runtime reporter DaemonSet values for these fields. Otherwise they will use the default values
-
-| Field                           | Description                                                                               | Accepted Values | Schema                                                                                                         | Default |
-| ------------------------------- | ----------------------------------------------------------------------------------------- | --------------- | -------------------------------------------------------------------------------------------------------------- | ------- |
-| spec.template.spec.containers   | Containers is a list of runtime reporter DaemonSet containers.                            |                 | [RuntimeReporterDaemonSetContainer](#runtimereporterdaemonsetcontainer)                                        |         |                                      |         |
-| spec.template.spec.affinity     | Affinity is a group of affinity scheduling rules for the runtime reporter DaemonSet pods. |                 | [corev1.Affinity](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#affinity-v1-core)       |         |
-| spec.template.spec.nodeSelector | NodeSelector is the runtime reporter pod's scheduling constraints.                        |                 | Map                                                                                                            |         |
-| spec.template.spec.tolerations  | Tolerations is the runtime reporter pod's tolerations.                                    |                 | [corev1.Tolerations](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#toleration-v1-core)  |         |
-
-### RuntimeReporterDaemonSetContainer
-
-| Field    | Description                                                                        | Accepted Values | Schema                      |
-| -------- | ---------------------------------------------------------------------------------- |---------------- | --------------------------  |
-| name     | Name of the runtime reporter daemonset container                                   |                 | [RuntimeReportDaemonSetSpec](#runtimereporterdaemonsetspec)  |
-| resource | Customization of limits and requests for compute resources such as cpu and memory. |                 | [corev1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#resourcerequirements-v1-core) |
--- a/calico-cloud/reference/resources/runtimesecurity.mdx
+++ b/calico-cloud/reference/resources/runtimesecurity.mdx
@@ -72,7 +72,7 @@ generation of runtime reports.
 | pod               | The exact name or regex expression of the pod(s) to which a user wants to negate the generation of runtime logs        |                 | string  |
 | namespace         | The exact name or regex expression of the namespace(s) to which a user wants to negate the generation of runtime logs  |                 | string  |
 
-### RuntimeReproterDaemonset
+### RuntimeReporterDaemonset
 
 | Field    | Description                                                  | Accepted Values | Schema                      |
 | -------- | ------------------------------------------------------------ |---------------- | --------------------------  |
--- a/calico-cloud/reference/resources/runtimesecurity.mdx
+++ b/calico-cloud/reference/resources/runtimesecurity.mdx
@@ -72,34 +72,16 @@ generation of runtime reports.
 | pod               | The exact name or regex of the pod(s) to which a user wants to negate the generation of runtime logs                   |                 | string  |
 | namespace         | The exact name or regex of the namespace(s) for which a user wants to negate the generation of runtime logs            |                 | string  |
 
-### RuntimeReporterDaemonset
-
-| Field    | Description                                                  | Accepted Values | Schema                      |
-| -------- | ------------------------------------------------------------ |---------------- | --------------------------  |
-| spec     | Spec is the specification of the runtime reporter DaemonSet. |                 | [RuntimeReportDaemonSetSpec](#runtimereporterdaemonsetspec)  |
-
-### RuntimeReportDaemonSetSpec
-
-| Field        | Description                                                                  | Accepted Values | Schema                                   |
-| ------------ | ---------------------------------------------------------------------------- | --------------- | ---------------------------------------- |
-| template     | Template describes the runtime reporter DaemonSet pod that will be created.  |                 | [RuntimeReporterDaemonSetPodTemplateSpec](#runtimereporterdaemonsetpodtemplatespec)  |
-
-### RuntimeReporterDaemonSetPodTemplateSpec
-
-| Field    | Description                                        | Accepted Values | Schema                           |
-| -------- | -------------------------------------------------- | --------------- | -------------------------------- |
-| spec     | Spec is the runtime reporter DaemonSet's PodSpec.  |                 | [RuntimeReporterDaemonSetPodSpec](#runtimereporterdaemonsetpodspec)  |
-
 ### RuntimeReporterDaemonSetPodSpec 
 
 If any of the fields below are specified, they will override the default runtime reporter DaemonSet values for these fields. Otherwise they will use the default values
 
-| Field        | Description                                                                               | Accepted Values | Schema                                                                                                         | Default |
-| --------     | ----------------------------------------------------------------------------------------- | --------------- | -------------------------------------------------------------------------------------------------------------- | ------- |
-| containers   | Containers is a list of runtime reporter DaemonSet containers.                            |                 | [RuntimeReporterDaemonSetContainer](#runtimereporterdaemonsetcontainer)                                        |         |                                      |         |
-| affinity     | Affinity is a group of affinity scheduling rules for the runtime reporter DaemonSet pods. |                 | [corev1.Affinity](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#affinity-v1-core)       |         |
-| nodeSelector | NodeSelector is the runtime reporter pod's scheduling constraints.                        |                 | Map                                                                                                            |         |
-| tolerations  | Tolerations is the runtime reporter pod's tolerations.                                    |                 | [corev1.Tolerations](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#toleration-v1-core)  |         |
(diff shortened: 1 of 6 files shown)


calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+The **RuntimeSecurity** custom resource (CR) is used to enable and configure Container Threat Detection in a Calico Cloud managed cluster.
+
+### Resource Definition
+
+```yaml
+apiVersion: operator.tigera.io/v1
+kind: RuntimeSecurity
+metadata:
+  name: defaultmy-runtimesecurity-config
```
nelljerram: Hm, doesn't it have to be named "default" ?

nelljerram: Yes it does, just checked.

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+spec:
+  detectorConfig:
+    - id: 190
```
nelljerram: I think these ids are bogus, aren't they?  I expect to see some kind of name.

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+      disabled: false
+    - id: 330
+      disbaled: True
```
nelljerram: sp: disabled
True should be true

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+```yaml
+apiVersion: operator.tigera.io/v1
+kind: RuntimeSecurity
+metadata:
+  name: defaultmy-runtimesecurity-config
+spec:
+  detectorConfig:
+    - id: 190
+      disabled: false
```
nelljerram: I think we do support `disabled: false`, but I don't think we should advertise it, because it has no purpose.

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+      disbaled: True
+  runtimeExceptionList:
+    - matching: regex
+      processInvocation: "/bin/ls*"
+      pod: "not-evil-pod"
+      namespace: "default"
+    - matching: exact
+      pod: "nginx"
+      namespace: default
+    - matching:
```
nelljerram: need a value here.

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+      namespace: "company-operations"
+```
+
+## Runtime Security Definition 
+
+### Metadata
+
+| Field  | Description                                   | Accepted Values                           | Schema |
+| ------ | --------------------------------------------- | ----------------------------------------- | ------ |
+| name   | The name of the runtime security resource.    | Lower-case alphanumeric with optional `-` | string |
```
nelljerram: Must always be "default"

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+| labels | A set of labels to apply to this resource.    |                                           | map    |
+
+
+### Spec
+
+| Field                    | Description                                                                                  | Accepted Values  | Schema                                                | Default |
+| ------------------------ | -------------------------------------------------------------------------------------------- | ---------------- | ----------------------------------------------------- | ------- |
+| detectorConfig           | Configuration parameters for a detector in a Calico Cloud managed cluster.                   |                  | [DetectorConfig](#detectorconfig)                     |         |
```
nelljerram: Can we say "Configuration that allows particular threat detectors to be disabled".

nelljerram: (I understand we might add more function here in the future.  When we do that, we can update the doc accordingly.  For now, the only thing this does is disabling, so let's say that.)

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+| runtimeExceptionList     | List of entries of suspicious processes that are allowed to run that wont generate an event  |                  | [runtimeExceptionList](#runtimeexceptionlist)         | Enabled |
```
nelljerram: sp: won't

nelljerram: I think should remove "suspicious" here.

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+| runtimereporterdaemonset | Daemonset spec to override the default daemonset spec for runtime reporter                   |                  | [RuntimeReporterDaemonSet](#runtimereporterdaemonset) |         |
```
nelljerram: Oh dear, I hate this, and it's really problematic for documentation.  Firstly I don't think this is true, i.e. it can't be used to override the whole spec.  I believe we only use it for overriding resource limits and requests.  Secondly, I'm not sure we intend it for users at all, so I wonder if we should document the field at all.

nelljerram: What about the RuntimeReporterConfig field?  This one is fully implemented, but problematic to document, unless we want to expose all that in our docs.

I think a reasonable solution would be to omit both RuntimeReporterConfig and runtimereporterdaemonset from this doc, and add a note somewhere saying "There are other fields in the RuntimeSecurity CR, but these are not yet designed for customer use".

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+| ------------------------ | -------------------------------------------------------------------------------------------- | ---------------- | ----------------------------------------------------- | ------- |
+| detectorConfig           | Configuration parameters for a detector in a Calico Cloud managed cluster.                   |                  | [DetectorConfig](#detectorconfig)                     |         |
+| runtimeExceptionList     | List of entries of suspicious processes that are allowed to run that wont generate an event  |                  | [runtimeExceptionList](#runtimeexceptionlist)         | Enabled |
+| runtimereporterdaemonset | Daemonset spec to override the default daemonset spec for runtime reporter                   |                  | [RuntimeReporterDaemonSet](#runtimereporterdaemonset) |         |          
+
+### DetectorConfig 
+
+The `DetectorCongfig` by default is not present but can hold a list of configurations entries for a detector in the Calico Cloud Managed cluster.
```
nelljerram: sp: Congfig

nelljerram: "DetectorConfig can be used to disable particular threat detectors."

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+| Field    | Description                                                       | Accepted Values | Schema  |
+| -------- | ----------------------------------------------------------------- | --------------- | ------- |
+| id       | The ID of the detector this entry applies too                     |                 | string  |
+| disabled | Boolean represents weather the detector should be disabled or not | True, False     | boolean |
+
+
+### RuntimeExceptionList
+
+This `RuntimeExceptionList` holds a list of entries that contains a list of supported fields by which a user can use to negate the
```
nelljerram: ... by which a user can negate the ...

nelljerram: ... entries that contain ...  (i.e. no "s" on "contain")

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+
+### RuntimeExceptionList
+
+This `RuntimeExceptionList` holds a list of entries that contains a list of supported fields by which a user can use to negate the 
+generation of runtime reports.
+
+| Field             | Description                                                                                                            | Accepted Values | Schema  |
+| --------          | ---------------------------------------------------------------------------------------------------------------------- | --------------- | ------- |
+| matching          | Whether the entries are exact matches to fields or considered a regex expression                                       | Exact, Regex    | string  |
```
nelljerram: "a regex expression" -> "as regular expressions (regexes)"

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+| processInvocation | The exact name or regex expression of the process to which a user wants to negate the generation of runtime logs       |                 | string  |
+| pod               | The exact name or regex expression of the pod(s) to which a user wants to negate the generation of runtime logs        |                 | string  |
+| namespace         | The exact name or regex expression of the namespace(s) to which a user wants to negate the generation of runtime logs  |                 | string  |
```
nelljerram: In all of these, delete "expression".  (Same as how we don't say "CD disc")

nelljerram: "to which" -> "for which"

calico-cloud/reference/resources/runtimesecurity.mdx
```diff
@@ -0,0 +1,109 @@
...
+| tolerations  | Tolerations is the runtime reporter pod's tolerations.                                    |                 | [corev1.Tolerations](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#toleration-v1-core)  |         |
+
+### RuntimeReporterDaemonSetContainer
+
+| Field    | Description                                                                        | Accepted Values | Schema                      |
+| -------- | ---------------------------------------------------------------------------------- |---------------- | --------------------------  |
+| name     | Name of the runtime reporter daemonset container                                   |                 | [RuntimeReportDaemonSetSpec](#runtimereporterdaemonsetspec)  |
+| resource | Customization of limits and requests for compute resources such as cpu and memory. |                 | [corev1.ResourceRequirements](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#resourcerequirements-v1-core) |
```
nelljerram: Not your fault, but I think this is a horrible presentation.  If we are going to document RuntimeReporterDaemonSet at all, I think we must do so by presenting a flat list of all of the fields that it allows to be overwritten, i.e. something like
```
- Spec.Template.Spec.Containers[*].Resources
- Spec.Template.Spec.Affinity
- ...
```
I've tried reading the relevant code to work out what that complete flat list would be, but the code is so horrible that I cannot.  I think our choices are:
1. Don't document this field at all (as already suggested in an earlier comment).
2. Go back to whoever developed this work and ask them to provide that flat list.

nelljerram (changes requested):

calico-cloud/reference/resources/runtimesecurity.mdx
LorcanMcVeigh: my thinking was there are lots of processes that don't generate an event so we wanted to specify that it is meant to suspicious process. but happy to remove it all the same :slightly_smiling_face:

calico-cloud/reference/resources/runtimesecurity.mdx
LorcanMcVeigh: Yeah I purposely left out RuntimeReporterConfig, The only reason the RuntimeReporterDaemonset was left in was because other components that have this field document it as well. But I'm more than happy to remove it

calico-cloud/reference/resources/runtimesecurity.mdx
nelljerram: Indeed.  I think it might be one of those cases where we _could_ spell out exactly what we really mean - which is "processes that our threat detection engine would normally consider to be suspicious, and so would generate Security Events, but which we want to prevent from generating Security Events when they occur with particular invocations, or in particular pods or namespaces" - but where the full description that it's better to have something shorter even if it's not precisely accurate, so long as the reader can infer the meaning.  But I'm not sure.  Perhaps we should give the longer precise definition in a form like I've written here?

calico-cloud/reference/resources/runtimesecurity.mdx
gantony: I would prefer to go with option 1.
We can always iterate later as needed.

For option 2, @jaderhs, do you know if other components that also support pod affinity etc have had their docs updated? If so we should follow the same format. If not, that’s a strong case for option 1 (we can always log a task to address that separately).

peterkellydev (approved):
From a content perspective of what we want to document, LGTM

calico-cloud/reference/resources/runtimesecurity.mdx
LorcanMcVeigh: ImageAssurance & Compliance benchmarking has documented these fields
- https://docs.tigera.io/calico-cloud/reference/installation/ia-api#image-assurance.operator.tigera.io/v1.CrawdadDaemonSet
- https://docs.tigera.io/calico-cloud/reference/installation/api#operator.tigera.io/v1.ComplianceBenchmarkerDaemonSet

But I am more than happy to remove references to these fields from the Docs altogether

calico-cloud/threat/container-threat-detection.mdx
```diff
@@ -90,6 +90,24 @@ If a malicious or suspicious program is run within the cluster, it will be repor
 
 ![malware-alert-example](/img/calico-cloud/malware-alert-example.png)
 
+
+### Create Exceptions for runtime logs
+
+If there is a program running within the cluster that is performing a suspicious action but you want to except it from create a runtime log, and thus from triggering a Security Event. Then you can create an Excetpion in the Runtime Security CR:
```
nelljerram: "from create" -> "from creating"

nelljerram: "Excetpion"  -> "Exception"

calico-cloud/threat/container-threat-detection.mdx
```diff
@@ -90,6 +90,24 @@ If a malicious or suspicious program is run within the cluster, it will be repor
...
+If there is a program running within the cluster that is performing a suspicious action but you want to except it from create a runtime log, and thus from triggering a Security Event. Then you can create an Excetpion in the Runtime Security CR: 
+
+```bash
+kubectl apply -f - <<EOF
+apiVersion: operator.tigera.io/v1
+kind: RuntimeSecurity
+metadata:
+  name: default
+spec:
+  runtimeExceptionList:
```
nelljerram: Surely we should provide an actual example here?

nelljerram (changes requested):
A few last things

calico-cloud/threat/container-threat-detection.mdx
LorcanMcVeigh: this was a mistake. I must've been on the wrong branch working on https://github.com/tigera/docs/pull/1692 and added this by mistake. Reverting this

nelljerram (approved):

ctauchen:
@LorcanMcVeigh Just want to check we have this in the right products and versions.

I figure this at least needs to go in for the latest CC release (calico-cloud/version-20-1). 

Does this also belong in CE versions?

nelljerram:
@ctauchen No, this is definitely CC only.

LorcanMcVeigh:
yup CC only @ctauchen

LorcanMcVeigh:
@ctauchen updated the version docs as well and sidebars. Can you take a look at this one as well when you get a chance 
please?

PR #1699, opened 2024-10-03
Add tiered policies to Calico OSS

mazdakn:
Doc changes related to:
- Supporting tiered policy in Calico OSS v3.29: https://github.com/projectcalico/calico/pull/9085
- Allowing users to specify tier's default action, i.e. `defaultAction`: https://github.com/projectcalico/calico/pull/9232

Product Version(s): Calico v3.29, Calico Enterprise v3.20.2

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
+++ b/calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
@@ -64,7 +64,7 @@ The default tier is where:
 
 ## AdminNetworkPolicy tier:
 
-The AdminNetworkPolicy tier where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000.
+The AdminNetworkPolicy tier is where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000.
 
 ## System tiers
 
@@ -92,7 +92,7 @@ Now you can reorder tiers by dragging and moving them.
 
 ## Tier order
 
-Tiers are ordered based on their orders (from left to right), starting from the lowest order (the highest priority) tiers.
+Tiers are sorted by their orders (from left to right), starting from the lowest order (the highest priority) tiers.
 
 ![tier order](/img/calico-cloud/tier-order.png)
 
--- a/calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
+++ b/calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
@@ -64,11 +64,7 @@ The default tier is where:
 
 ## AdminNetworkPolicy tier:
 
-The AdminNetworkPolicy tier is created during installation and has the order of 1,000.
-
-The AdminNetworkPolicy tier is where:
-
-- You manage all Kubernetes admin network policies
+The AdminNetworkPolicy tier where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000.
 
 ## System tiers
 
@@ -96,7 +92,7 @@ Now you can reorder tiers by dragging and moving them.
 
 ## Tier order
 
-Tiers are ordered from left to right, starting with the highest priority (also called highest precedence) tiers.
+Tiers are ordered based on their orders (from left to right), starting from the lowest order (the highest priority) tiers.
 
 ![tier order](/img/calico-cloud/tier-order.png)
 
--- a/calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
+++ b/calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
@@ -52,7 +52,7 @@ kubectl apply -f security.yaml
 
 ## The default tier:
 
-The default tier is created during installation and has the order of 1,000,000.
+The AdminNetworkPolicy tier where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
 
 ![default tier](/img/calico-cloud/default-tier.png)
 
@@ -100,13 +100,13 @@ In the example above, tier priorities are as follows:
 
 - **security tier** - is higher priority than platform tier
 - **platform tier** - is higher priority than default tier
-- **default tier** - is the lowest priority than the rest
+- **default tier** - is the lowest priority
 
 The tier you put as the highest priority (after system tiers), depends on your environment. In compliance-driven environments, the security tier may be the highest priority (as shown above). There is no one-size-fits-all order.
 
 ## Policy processing
 
-Policies are processed in sequential order from top to bottom.
+Policies are processed in sequential order from lowest order (highest priority) to highest order (lowest priority), which is from top to bottom in the following diagram.
 
 ![policy processing](/img/calico-cloud/policy-processing.png)
 
@@ -140,19 +140,51 @@ The following diagrams show the relationship between all of the elements that af
 ### Implicit default deny
 
 As shown in the following diagram, at the end of each tier is an implicit default deny. This is a safeguard that helps mitigate against unsecured policy.
-To override the implicit default deny, and to allow traffic evaluation to continue, you can change tier's default action to **Pass**.
+Because of this safeguard, you must either explicitly update tier's default action to **Pass** or apply a Pass action rule when you want traffic evaluation to continue.
+In the following example, the Pass action in a policy ensures that traffic evaluation continues, and overrides the implicit default deny.
 
-![implicit deny](/img/calico-cloud/implicit-deny.svg)
+![implicit-deny](/img/calico-cloud/implicit-deny.svg)
 
+Let’s look at a Dev/Ops global network policy in a high precedence tier (Platform). The policy denies ingress and egress traffic to workloads that match selector, `env == "stage"`. To ensure that policies continue to evaluate traffic after this policy, the policy adds a Pass action for both ingress and egress.
+
+**Pass action rule example**
+
+```yaml
+apiVersion: projectcalico.org/v3
+kind: GlobalNetworkPolicy
+metadata:
+  name: devops.stage-env
+spec:
+  tier: devops
+  order: 255
+  selector: env == "stage"
+  ingress:
+    - action: Deny
+      source:
+        selector: env != "stage"
+    - action: Pass
+  egress:
+    - action: Deny
+      destination:
+        selector: env != "stage"
+    - action: Pass
+  types:
+    - Ingress
+    - Egress
+```
+
+The other option, is to update tier's default action to **Pass** to override the implicit default deny.
+
+**Pass default action tier example**
 
 ```yaml
 apiVersion: projectcalico.org/v3
 kind: Tier
 metadata:
-  name: security
+  name: devops
 spec:
   order: 300
-  defaultAction: Pass
+  defautlAction: Pass
 ```
 
 ### Policy endpoint matching across tiers
--- a/calico-cloud/reference/resources/tier.mdx
+++ b/calico-cloud/reference/resources/tier.mdx
@@ -30,7 +30,7 @@ Policies in each Tier are then processed in order.
 
 If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped by default. This behaviour can be changed by setting the `defaultAction` to `Pass` for a tier.
 
-If the last Tier applying to the endpoint `Pass`es the packet, that endpoint's [Profiles](profile.mdx) are evaluated.
+If the last Tier applying to the endpoint `Pass`es the packet, that endpoint's [Profiles] (profile.mdx) are evaluated.
 
 ## Sample YAML
 
--- a/calico-cloud/reference/resources/tier.mdx
+++ b/calico-cloud/reference/resources/tier.mdx
@@ -28,7 +28,7 @@ Policies in each Tier are then processed in order.
 - If a [NetworkPolicy](networkpolicy.mdx) or [GlobalNetworkPolicy](globalnetworkpolicy.mdx) in the Tier `Allow`s or `Deny`s the packet, then evaluation is done: the packet is handled accordingly.
 - If a [NetworkPolicy](networkpolicy.mdx) or [GlobalNetworkPolicy](globalnetworkpolicy.mdx) in the Tier `Pass`es the packet, the next Tier containing a Policy that applies to the endpoint processes the packet.
 
-If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped by default. This behaviour can be changed by setting the `defaultAction` to `Pass` for a tier.
+If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped. This behaviour can be changed by setting the `defaultAction` of a tier to `Pass`.
 
 If the last Tier applying to the endpoint `Pass`es the packet, that endpoint's [Profiles] (profile.mdx) are evaluated.
 
--- a/calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
+++ b/calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
@@ -64,7 +64,7 @@ The default tier is where:
 
 ## AdminNetworkPolicy tier:
 
-The AdminNetworkPolicy tier where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
+The AdminNetworkPolicy tier is where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
 
 ## System tiers
 
@@ -90,7 +90,7 @@ Now you can reorder tiers by dragging and moving them.
 
 ## Tier order
 
-Tiers are ordered based on their orders (from left to right), starting from the lowest order (the highest priority) tiers.
+Tiers are sorted by their orders (from left to right), starting from the lowest order (the highest priority) tiers.
 
 ![tier-order](/img/calico-enterprise/tier-order.png)
 
--- a/calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
+++ b/calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
@@ -64,11 +64,7 @@ The default tier is where:
 
 ## AdminNetworkPolicy tier:
 
-The AdminNetworkPolicy tier is created during installation and has the order of 1,000.
-
-The AdminNetworkPolicy tier is where:
-
-- You manage all Kubernetes admin network policies
+The AdminNetworkPolicy tier where is all Kubernetes admin network policies reside. It is automatically created during installation and has the order of 1,000.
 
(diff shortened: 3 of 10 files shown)


calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -1,8 +1,8 @@
 ---
-description: Learn about policies, tiers, and policy evaluation.
+description: Understand how tiered policy works and supports microsegmentation.
 ---
 
-# Policy tiers tutorial
+# Get started with policy tiers
```
mazdakn, line 5: The above changes happened by syncing enterprise version with cloud version. It make sense to sync these, unless there is reason behind the diff.

calico/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+**Tiers** are a hierarchical construct used to group policies and enforce higher precedence policies that cannot be circumvented by other teams. As you will learn in this tutorial, tiers have built-in features that support workload microsegmentation.
+
+All {{prodname}} and Kubernetes network policies reside in tiers. You can start "thinking in tiers" by grouping your teams and types of policies within each group. For example, we recommend these three tiers (platform, security, and application).
+
+![policy-types](/img/calico-enterprise/policy-types.png)
+
+Next, you can determine the priority of policies in tiers (from top to bottom). In the following example, the platform and security tiers use {{prodname}} global network policies that apply to all pods, while developer teams can safely manage pods within namespaces using Kubernetes network policy for their applications and microservices.
+
+![policy-tiers](/img/calico-enterprise/policy-tiers.png)
```
mazdakn, line 25: @ctauchen It looks the enterprise and cloud versions, each keep a copy of the pictures/diagrams.
should we keep a copy the diagrams for Calico? Isn't it better to not keep multiple copies and use a shared one?

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -62,6 +62,14 @@ The default tier is where:
...
 - Recommended policies are placed when you use the **Recommend a policy** feature
 
+## AdminNetworkPolicy tier:
+
+The AdminNetworkPolicy tier is created during installation and has the order of 1,000.
+
+The AdminNetworkPolicy tier is where:
+
+- You manage all Kubernetes admin network policies
+
```
tomastigera:
```suggestion
The AdminNetworkPolicy tier where is all Kubernetes admin network policies reside. It is automatically created during installation and has the order of 1,000.

```

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -94,7 +102,7 @@ In the example above, tier priorities are as follows:
 
 - **security tier** - is higher priority than platform tier
 - **platform tier** - is higher priority than default tier
-- **default tier** - is always the last tier and cannot be reordered
+- **default tier** - is the lowest priority than the rest
```
tomastigera:
```suggestion
- **default tier** - is the lowest priority.
```

Previously, is could not be reordered, now it has an order. Can it be reordered? If yes, it would be worth to say that "previously it could not be reordered. It can be now, but we advise against that" or something similar?

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -133,36 +141,20 @@ The following diagrams show the relationship between all of the elements that af
 
 ### Implicit default deny
 
-As shown in the following diagram, at the end of each tier is an implicit default deny. This is a safeguard that helps mitigate against unsecured policy. Because of this safeguard, you must explicitly apply the **Pass** action rule when you want traffic evaluation to continue. In the following example, the Pass action in a policy ensures that traffic evaluation continues, and overrides the implicit default deny.
+As shown in the following diagram, at the end of each tier is an implicit default deny. This is a safeguard that helps mitigate against unsecured policy.
+To override the implicit default deny, and to allow traffic evaluation to continue, you can change tier's default action to **Pass**.
```
tomastigera: I think some of the original text should remain in some form:

```
 Because of this safeguard, you must explicitly apply the **Pass** action rule when you want traffic evaluation to continue. In the following example, the Pass action in a policy ensures that traffic evaluation continues, and overrides the implicit default deny.
 ```
 
 This is still somewhat true as this is one way how to get from tier to another. Now we are adding another method, which may not be as safe. Btw. the ANP tier has pass as a default action? I would think so.

File: calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx [hunk 6]
```diff
@@ -133,36 +141,20 @@ The following diagrams show the relationship between all of the elements that af
...
 
 ![implicit-deny](/img/calico-enterprise/implicit-deny.svg)
 
-Let’s look at a Dev/Ops global network policy in a high precedence tier (Platform). The policy denies ingress and egress traffic to workloads that match selector, `env == "stage"`. To ensure that policies continue to evaluate traffic after this policy, the policy adds a Pass action for both ingress and egress.
```
tomastigera: It feels like this should remain

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -133,36 +141,20 @@ The following diagrams show the relationship between all of the elements that af
...
-  egress:
-    - action: Deny
-      destination:
-        selector: env != "stage"
-    - action: Pass
-  types:
-    - Ingress
-    - Egress
+  order: 300
+  defaultAction: Pass
```
tomastigera: And this should be in addition to the pass rule. We did not remove the pass rule, right?

calico-enterprise/reference/resources/tier.mdx
```diff
@@ -28,7 +28,7 @@ Policies in each Tier are then processed in order.
 - If a [NetworkPolicy](networkpolicy.mdx) or [GlobalNetworkPolicy](globalnetworkpolicy.mdx) in the Tier `Allow`s or `Deny`s the packet, then evaluation is done: the packet is handled accordingly.
 - If a [NetworkPolicy](networkpolicy.mdx) or [GlobalNetworkPolicy](globalnetworkpolicy.mdx) in the Tier `Pass`es the packet, the next Tier containing a Policy that applies to the endpoint processes the packet.
 
-If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped.
+If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped by default. This behaviour can be changed by setting the `defaultAction` to `Pass` for a tier.
```
tomastigera:
```suggestion
If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped by. This behaviour can be changed by setting the `defaultAction` to `Pass` for a tier.
```

Perhaps chance to clarify `If the Tier applies to the endpoint` as I was confused what it mean myself. It means that that tier contains a rule that applies to an endpoint? How can it apply to an endpoint and not to take an action? Example?

calico/network-policy/policy-tiers/rbac-tiered-policies.mdx
```diff
@@ -0,0 +1,507 @@
+---
+description: Configure RBAC to control access to policies and tiers.
+---
+
+# Configure RBAC for tiered policies
```
tomastigera, line 5: I suppose this is just a plain copy from EE, isn't it?

How does the RBAC play with ANP tier? Is there some RBAC created by default or do kubernetes handle it on their own? Idk how it works, just double checking here!

calico/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+The AdminNetworkPolicy tier is created during installation and has the order of 1,000.
+
+The AdminNetworkPolicy tier is where:
+
+- You manage all Kubernetes admin network policies
+
+## Tier order
+
+Tiers are ordered from left to right, starting with the highest priority (also called highest precedence) tiers.
```
tomastigera: left to right is a UI concept. We do not have anything like in OSS and we cannot reference a UI image

calico/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+In the example above, tier priorities are as follows:
+
+- **security tier** - is higher priority than platform tier
+- **platform tier** - is higher priority than default tier
+- **default tier** - is the lowest priority than the rest
+
+## Policy processing
+
+Policies are processed in sequential order from top to bottom.
```
tomastigera: similarly, there is no top to bottom without UI

calico/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -0,0 +1,164 @@
...
+  order: 300
+```
+
+```bash
+kubectl apply -f security.yaml
+```
+
+## The default tier:
+
+The default tier is created during installation and has the order of 1,000,000.
```
tomastigera:
```suggestion
The default tier is created during installation and has the order of 1,000,000. The desire is to evaluate policies in the default tier as last.
```

I meantioned it before, but worth spelling it out explicitly whether the order can by changed.

tomastigera (changes requested):

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: No, with the `defaultAction` added, there is no need to add rules with Pass action just to prevent hitting the implicit deny. I think the previous method looked like a hack, and this version is cleaner.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
tomastigera: I think there should be a link to kANP documentation!

calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
ctauchen: I agree that it makes sense, but only if you're happy to stand over the changes. I wasn't involved in the original documentation, so I can't explain the diffs. Could be simply that someone forgot to carry over updates, could be that some were intentional. Or a confusing mix of the two.

calico/network-policy/policy-tiers/tiered-policy.mdx
ctauchen: Unless we're changing the diagrams, it's fine to reference an existing one. The only reason we maintain duplicates for CE and CC is that we migrated everything over from separate repositories. Rationalizing our images hasn't been a priority.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
tomastigera: I understand that we have a better approach, but I mean we have two way now. At the least, We should add that the previous method still works, but is not recommended. There are people who will upgrade and will not change their rules and may get confused if something does not work, that they are doing it wrong. WDYT?

calico-cloud/reference/resources/tier.mdx
```diff
@@ -28,11 +28,9 @@ Policies in each Tier are then processed in order.
 - If a [NetworkPolicy](networkpolicy.mdx) or [GlobalNetworkPolicy](globalnetworkpolicy.mdx) in the Tier `Allow`s or `Deny`s the packet, then evaluation is done: the packet is handled accordingly.
 - If a [NetworkPolicy](networkpolicy.mdx) or [GlobalNetworkPolicy](globalnetworkpolicy.mdx) in the Tier `Pass`es the packet, the next Tier containing a Policy that applies to the endpoint processes the packet.
 
-If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped.
+If the Tier applies to the endpoint, but takes no action on the packet the packet is dropped by default. This behaviour can be changed by setting the `defaultAction` to `Pass` for a tier.
 
-<!--TODO-XREF-CC
-If the last Tier applying to the endpoint `Pass`es the packet, that endpoint's [Profiles](/reference/resources/profile) are evaluated.
--->
+If the last Tier applying to the endpoint `Pass`es the packet, that endpoint's [Profiles](profile.mdx) are evaluated.
```
ctauchen: This xref breaks the build. An example of a file from the CE docs that doesn't exist in CC.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: No, the default tier **must** have the order of 1M. It cannot be reordered. Previously, it had an order of `nil`, and now it is `1M` and it cannot change. Also any  tier with `nil` order will be treated as `1M`.
We will add a release note to warn customers about the change.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: Yes, ANP tier has `Pass` default action.

Kept the original text, and added the option with the default action.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: Fixed.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: It makes sense to document both options. Adding a pass rule is more specific than having default pass action.

calico-enterprise/reference/resources/tier.mdx
mazdakn: This page is about documenting the resource itself. I believe it should be short and to the point. There is more description about this topic in the tiered policy page: https://docs.tigera.io/calico-enterprise/latest/network-policy/policy-tiers/tiered-policy#default-endpoint-behavior

calico-cloud/network-policy/policy-tiers/tiered-policy.mdx
tomastigera: There are so many discrepancies across the docs, that I would not worry about it :man_shrugging: At least it fixes this discrepancy.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -62,6 +62,10 @@ The default tier is where:
 - Network and global network policies are placed when you upgrade from Project Calico to {{prodname}}
 - Recommended policies are placed when you use the **Recommend a policy** feature
 
+## AdminNetworkPolicy tier:
+
+The AdminNetworkPolicy tier where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
```
tomastigera:
```suggestion
The AdminNetworkPolicy tier is where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
```

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -86,21 +90,21 @@ Now you can reorder tiers by dragging and moving them.
 
 ## Tier order
 
-Tiers are ordered from left to right, starting with the highest priority (also called highest precedence) tiers.
+Tiers are ordered based on their orders (from left to right), starting from the lowest order (the highest priority) tiers.
```
tomastigera:
```suggestion
Tiers are sorted by their orders (from left to right), starting from the lowest order (the highest priority) tiers.
```

calico/network-policy/policy-tiers/tiered-policy.mdx
```diff
@@ -0,0 +1,181 @@
...
+
+The default tier is where:
+
+- You manage all Kubernetes network policies
+- All network and global network policies without an explicit tier are placed.
+- Network and global network policies are placed when you upgrade from Project Calico without tier support to {{prodname}} release with tier support.
+
+## AdminNetworkPolicy tier:
+
+The AdminNetworkPolicy tier where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
```
tomastigera:
```suggestion
The AdminNetworkPolicy tier is where all [Kubernetes admin network policies](https://network-policy-api.sigs.k8s.io/reference/examples/) reside. It is automatically created during installation and has the order of 1,000. This value is fixed, and cannot be changed.
```

tomastigera (commented):
a few tiny typos

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: Fixed.

calico-enterprise/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: Fixed.

calico/network-policy/policy-tiers/tiered-policy.mdx
mazdakn: Fixed.

ctauchen:
@tmjd There are some changes here to the CC tiers docs. They appear to be removing innocuous diffs between the CE and CC versions. But might be a good idea for you do take a quick pass at the changes here to see if anything is wrong for CC.

calico/network-policy/policy-tiers/rbac-tiered-policies.mdx
```diff
@@ -0,0 +1,507 @@
...
+
+Create an Admin user with full access to the {{prodname}} Manager (as well as everything else in the cluster) using the following command. See the Kubernetes documentation to identify users based on your chosen [authentication method](https://kubernetes.io/docs/reference/access-authn-authz/authentication/), and how to use the [RBAC resources](https://kubernetes.io/docs/reference/access-authn-authz/rbac/).
+
+```bash
+kubectl create clusterrolebinding permissive-binding \
+    --clusterrole=cluster-admin \
+    --user=<USER\>
+```
+
+### Create minimum permissions for all non-Admin users
```
ctauchen: This section discusses the UI. Not sure how much needs removing. What do you think?

calico/network-policy/policy-tiers/rbac-tiered-policies.mdx
ctauchen: FYI @mazdakn

calico/network-policy/policy-tiers/rbac-tiered-policies.mdx
mazdakn: Right. Removed it.

mazdakn: I believe we can remove the whole section.

calico/network-policy/policy-tiers/rbac-tiered-policies.mdx
mazdakn: Yes, this is just a copy from EE. 
ANP design states the API is for cluster administrators. This is not something we enforce though. We provide Tier's RBAC for users to configure the access, so in the end it's up to users to setup the access properly.

tmjd:
@ctauchen I don't see anything that should be different for Calico Cloud.

ctauchen:
LGTM

PR #1725, opened 2024-10-17
fix explaining DNS policy modes

tomastigera:
Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/network-policy/domain-based-policy.mdx
+++ b/calico-enterprise/network-policy/domain-based-policy.mdx
@@ -219,24 +219,19 @@ spec:
         selector: color == 'red'
 ```
 
-:::note
+### Tuning Performance and Latency
 
-Newly initialized pods sometimes fail to connect to domains that are allowed by a DNS policy (staged or enforced).
+{{prodname}} supports different DNS policy modes with different peformance and latency implications. The
+default mode for iptables/nftables dataplane is `DelayDeniedPacket`, eBPF dataplane defaults to
+`Inline` on kernels 5.17 (or RedHat 5.14) and newer and to `NoDelay` on older kernels. Windows
+dataplane only support `NodeDelay` mode.
+
+With `NoDelay` mode, newly pods sometimes fail to connect to domains that are allowed by a DNS policy.
 In these cases, the pod tries to connect to a domain before {{prodname}} has finished processing the DNS policy that allows the connection.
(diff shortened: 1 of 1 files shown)


ctauchen (commented):
At this stage, I'm just wondering where the best place is for this information.

It appears that you're explaining the what the default or supported DNS policy modes are for the different dataplanes. You provide the values for each dataplane, and then point people to the reference documentation.

I wonder whether it would be more useful to explain what these differences are and why someone should be interested in them? In other words, to provide some degree of information for users so they don't have to go to reference to get a rough idea of what this means.

The title suggests that there's something that can be done for performance and latency. I suspect these policy modes have something to do with that, but the text doesn't make the connection. Perhaps this could be more explicit?

tomastigera:
> At this stage, I'm just wondering where the best place is for this information.

There will be a whole blog on this, perhaps we can reference it in the docs when it goes out https://docs.google.com/document/d/19TMYGQPsN2xA1jveZmoI8rLvsVL9o8GG8NlyQRtHp5A/edit?usp=sharing

> 
> It appears that you're explaining the what the default or supported DNS policy modes are for the different dataplanes. You provide the values for each dataplane, and then point people to the reference documentation.

yes, that is kind of what it did before, but in an incomplete and misleading way.

> 
> I wonder whether it would be more useful to explain what these differences are and why someone should be interested in them? In other words, to provide some degree of information for users so they don't have to go to reference to get a rough idea of what this means.

though about that, but I definitely do not want to maintain the same information in even more places


> 
> The title suggests that there's something that can be done for performance and latency. I suspect these policy modes have something to do with that, but the text doesn't make the connection. Perhaps this could be more explicit?

Calico Enterprise supports different DNS policy modes with different peformance and latency implications.
...
For more information, see DNSPolicyMode and BPFDNSPolicyMode.

I don't know how more explicit this can be. If you are not happy with the defaults, go and dig deeper, that is the message.

tomastigera:
@ctauchen any progress on this or are we going to stay in "I am wondering state" ? I provided the information, but I am in no position to make calls on where this should end up. It is still marked as a draft because I am not going to update other places unless I have an ack on this. But feel free to take the information and place it where ever you see fit. Currently the docs are wrong.

ctauchen:
Reopened and merged here: #1854

PR #1767, opened 2024-11-08
DOCS-2385: add interface verification step for BPF dataplane configuration

ivansharamok:
In some corner cases cluster nodes may have custom interface names which may not get matched by the `bpfDataIfacePattern` regex. Need to clarify what users need to change to have their nodes BPF dataplane ready.

Product Version(s):
CE v3.19+

Issue:
[DOCS-2385](https://tigera.atlassian.net/browse/DOCS-2385)

Link to docs preview:

SME review:

DOCS review:

Additional information:
This needs to be added to all versions of Calico docs.

Merge checklist:
- [ ] Test have passed 

[DOCS-2385]: https://tigera.atlassian.net/browse/DOCS-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/calico-enterprise_versioned_docs/version-3.19-2/operations/ebpf/enabling-ebpf.mdx
+++ b/calico-enterprise_versioned_docs/version-3.19-2/operations/ebpf/enabling-ebpf.mdx
@@ -185,6 +185,17 @@ kubectl patch felixconfiguration default --patch='{"spec": {"bpfKubeProxyIptable
 
 If both `kube-proxy` and `BPFKubeProxyIptablesCleanupEnabled` is enabled then `kube-proxy` will write its iptables rules and Felix will try to clean them up resulting in iptables flapping between the two.
 
+### Verify node interface naming pattern
+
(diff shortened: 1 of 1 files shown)


ctauchen:
Given that this is an edge case, I wonder whether this wouldn't fit better in the [ebpf troubleshooting page](https://docs.tigera.io/calico/latest/operations/ebpf/troubleshoot-ebpf)?

ivansharamok:
this is not necessarily an edge-case scenario. Interface names on the nodes differ based on the environment. For instance, the type of a VM provided by cloud providers can have different interface names, VMWare sets its own naming pattern, any virtualization solution can use its own naming pattern, on-prem bare-metals can have their own naming patterns. This is a step that everyone should do to make sure their cluster is configured correctly before they switch to Calico BPF mode.

ivansharamok:
adjusted the message in the proposed section to verify the node's interface naming pattern according to the feedback received from the engg team.

ctauchen:
LGTM, will follow up with changes to later versions.

PR #1777, opened 2024-11-20
Documentation for Log actions

frozenprocess:
Product Version(s):

Issue:
Log actions doesn’t have a clear example.
Ebpf log actions are now supported.


calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
+---
+description: Debug your policies with Log rules.
+---
+# Use Log action to debug policies
+
+## Big picture
+Calico Log rules are powerful resources that can be implemented in an environment to log the matching traffic and allowing you to determine how your policies are behaving.
```
tomastigera: >  Log rules are powerful resources that can be implemented in an environment

what do you mean by that? It is enough to say that you can use it to see how other rules are behaving.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs groupId='log-rules'>
+<TabItem label="eBPF" value="eBPF">
+    - Calico 3.29+
+    - Linux Kernel 5.16+
+
+    eBPF policy logs depend on advanced kernel features introduced in Linux Kernel 5.16 and later. If you encounter a scenario where logs only display the traffic verdict (e.g., ALLOWED or DENIED) without detailed information about IP packets, it indicates that your system is running on an older kernel version lacking the necessary capabilities.
```
tomastigera:
```suggestion
    eBPF policy logs depend on eBPF printing capabilities introduced in Linux Kernel 5.16 although they may have been backported to your distros kernel. If you encounter a scenario where logs only display the traffic verdict (e.g., ALLOWED or DENIED) without detailed information about IP packets, it indicates that your system is running on an older kernel version lacking the necessary capabilities.
```

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+## Concepts
+A Calico policy with the action `Log` serves as a diagnostic tool that captures and logs network traffic information based on the rules and criteria specified. By leveraging the `Log` action, administrators gain insight into how traffic is evaluated by their policies, enabling them to debug, refine, and validate their network security posture.
+
+
+:::caution
+
+## Key Behavior Differences: eBPF vs. iptables log actions
+
+Please note that Log action in eBPF behaves differently from iptables.
```
tomastigera:
```suggestion
Please note that Log action behaviour in different dataplanes behaves differently based on the the available features.
```

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+### eBPF 
+
+In the eBPF dataplane, the Log action outputs the final verdict applied to the traffic. This means that the log entry reflects the ultimate decision—whether the traffic is allowed or denied—after evaluating the full set of applicable rules and policies.
```
tomastigera:
```suggestion
BPF dataplane logs the final verdict applied to the packet once it matched ANY log rule. This means that the log entry reflects the ultimate decision—whether the traffic is allowed or denied—after evaluating the full set of applicable rules and policies.
```

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+
+<Tabs groupId='log-rules'>
+<TabItem label="eBPF" value="eBPF">
+eBPF policy logs are sent to the trace pipe and can be viewed by using the following command:
+```bash
+kubectl exec -n calico-system -it ds/calico-node -- bpftool prog tracelog
+```
+
+An eBPF log starts with the pid name and ID
```
tomastigera: pid and process name - may not be related to the process responsible for the traffic - be clear about it if you want to meantion it.

More importantly explain `cali527b0801ecb-E` or `eth0-X`

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+    eBPF policy logs depend on advanced kernel features introduced in Linux Kernel 5.16 and later. If you encounter a scenario where logs only display the traffic verdict (e.g., ALLOWED or DENIED) without detailed information about IP packets, it indicates that your system is running on an older kernel version lacking the necessary capabilities.
+
+</TabItem>
+
+<TabItem label="iptables/ipvs" value="iptables">
+    - Rsyslogd
+    
+    :::note
+    
+    Keep in mind that `logs` action doesn't output anything if you are running a `Kind` cluster given that `rsyslogd` is not installed in `Kind` created clusters.
```
fasaxc: I don't think this is right, logs can go to the kernel log, syslog, or to the system journal.  Recent version of nftables might even send them to the nftables trace pipe, which needs the nftrace tool to see them (not sure if logs go there or if it's only TRACE actions).

Logs don't work inside containers because it's disabled by default, not because syslog isn't running.  That's a security feature since there's only one log and it would be bad if containerised apps could spam your host's logs.

calico/network-policy/policy-rules/log-rules.mdx
fasaxc:
```suggestion
The precise format of the eBPF trace logs depends on your kernel and sysctl settings.  However, a typical log format includes the active process and PID (`curl-5288` in the example), the CPU number (`[000]`), various flags, a timestamp and then the log message.  Note that, for incoming packets, the reported process may not be the process that eventually receives the packet; it is typically a kernel IRQ process, or another process that happened to trigger a poll of the network interface.
```

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+curl-5288    [000] ..s2.  3055.982021: bpf_trace_printk: cali527b0801ecb-E: policy ALLOWED proto 6 src 172.16.193.131:56042 dest 69.147.80.12:80
+
+<...>-4950    [001] ..s2.  2960.748464: bpf_trace_printk: cali527b0801ecb-E: policy ALLOWED proto 1 src 172.16.193.131:0 dest 8.8.8.8:8
+
+<...>-5288    [000] ..s2.  3055.954466: bpf_trace_printk: cali527b0801ecb-E: policy ALLOWED proto 17 src 172.16.193.131:36325 dest 10.43.0.10:53
+```
+
+</TabItem>
+<TabItem label="iptables/ipvs" value="iptables">
+Policy logs are sent to the syslog, this requires your host installation to have the rsyslog package.
```
fasaxc: Still not sure that it's syslog

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,184 @@
...
+  - Egress
+  egress:
+  - action: Log
+  ingress:
+  - action: Log
+```
+
+### Logging all traffic in a cluster
+
+By default, Calico policies apply only to namespaced resources (e.g., Pods, ServiceAccounts). To achieve total protection, enable hostEndpoint support to include non-namespaced resources and host nodes.
```
fasaxc: We should caution that BPF log actions are very slow and should only be used for debugging.  As soon as a program logs it seems to add a  huge overhead.

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: I'm going to replace it with
```
In a containerized Kubernetes environment, such as Kind, receiving syslog messages is not permittied. This limitation is by the design to prevent a noisy neighbor scenario since both containers and the host share the same Kernel .
```

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Noted.

calico/network-policy/policy-rules/log-rules.mdx
tomastigera: but you can enable it! It is only disabled by default.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,192 @@
+---
+description: Debug your policies with Log rules.
+---
+# Use Log action to debug policies
+
+## Big picture
+$[prodname] Log rules are powerful resources that can be implemented in an environment to log the matching traffic and allowing you to determine how your policies are behaving.
```
tomastigera:
```suggestion
$[prodname] Log rules are powerful resources  to log the matching traffic and allowing you to determine how your policies are behaving.
```

tomastigera (commented):
ohhh I fogot to click to submit my earlier feedback :doh:

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,192 @@
...
+</TabItem>
+</Tabs>
+
+
+## How to
+
+:::caution
+For a complete and efficient observability, consider exploring the use of [Flow Logs](/calico-enterprise/latest/visibility/visualize-traffic) available in Calico Enterprise and Calico Cloud.
+
+Using log rules may not be the most effective way to achieve comprehensive observability. While network observability can be achieved by utilizing log actions, this approach may introduce performance overhead in busy clusters.
```
tomastigera:
```suggestion
Using log rules is not be the most effective way to achieve comprehensive observability. While network observability can be achieved by utilizing log actions, this approach may introduce significant performance overhead.
```

fasaxc (approved):
LGTM.  Please can you cross-link it with the bpf troubleshooting page/next steps links on the BPF install page

frozenprocess:
Hey, any updates on this?

ctauchen:
@frozenprocess Yes, I've spent some time with this but was planning to get notes to you Monday when you're back from holiday ...

frozenprocess:
@ctauchen if changes are not that severe let's merge this and I'll raise another PR to address them.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
+---
+description: Debug your policies with Log rules.
+---
+# Use Log action to debug policies
+
+## Big picture
```
ctauchen, line 6: We don't use the old template anymore. Based on what I'm seeing, you should aim for a structure like this:

```
# title

quick summary about this stuff

## About log actions

Here you say what they are, how they work, and who should use them.

## Prerequisites

## Do this procedure

## Do that procedure

## Additional resources

File: calico/network-policy/policy-rules/log-rules.mdx [hunk 12]
```diff
@@ -0,0 +1,206 @@
...
+$[prodname] Log rules are powerful resources to log the matching traffic and allowing you to determine how your policies are behaving.
+
+## Value
+$[prodname] has a unique `Log` action that provides traffic observability and logging which is missing in the standard Kubernetes Network Policy. This unique action can be used by security teams and admins to troubleshoot their policies and make sure that their cluster security posture is doing what its expected to do.
+
+
+## Requirements
```
ctauchen, line 13: Let's scrap the tab structure. Something more direct, like this:

```
## Prerequisites
* Your cluster is not a containerized Kubernetes environment, such as Kind.
* If you're running $[prodname] with the eBPF dataplane:
   * You installed $[prodname] 3.29 or later.
   * Your cluster node or VM host is running on a Linux distribution based on the Linux kernel 5.16 or later.
 ```

File: calico/network-policy/policy-rules/log-rules.mdx [hunk 13]
```diff
@@ -0,0 +1,206 @@
...
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs groupId='log-rules'>
+<TabItem label="eBPF" value="eBPF">
+    - $[prodname] 3.29 and above
+    - Linux Kernel 5.16 and above
+
+     eBPF policy logs depend on eBPF printing capabilities introduced in Linux Kernel 5.16 although they may have been backported to your distros kernel. If you encounter a scenario where logs only display the traffic verdict (e.g., ALLOWED or DENIED) without detailed information about IP packets, it indicates that your system is running on an older kernel version lacking the necessary capabilities.
```
ctauchen, line 23: Interesting, but unnecessary. We can with just the version number.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+    In a containerized Kubernetes environment, such as Kind, sending `syslog` messages to the Kernel is not permitted. This limitation is by the design to prevent a noisy neighbor scenario since both containers and the host share the same Kernel.
+
+    :::
+
+    The $[prodname] Log action creates a LOG rule in iptables. Any logs matching this rule are recorded by the kernel's logging service, usually through syslog.
+
+    The following suffix, is an example of iptables LOG ruled programmed by $[prodname]:
+    ```bash
+    -j LOG --log-prefix "calico-packet: " --log-level 5
+    ```
```
ctauchen, line 40: Again, interesting, but not necessary for the user's main goal here.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+</TabItem>
+</Tabs>
+
+## Concepts
+A $[prodname] policy with the action `Log` serves as a diagnostic tool that captures and logs network traffic information based on the rules and criteria specified. By leveraging the `Log` action, administrators gain insight into how traffic is evaluated by their policies, enabling them to debug, refine, and validate their network security posture.
```
ctauchen, line 46: Renamed, moved up as indicated above. 

For the 'about' section, I think it would be good to back up a bit and explain the basics of what's going on here. Specifically, explain how 'Log' works with 'Allow' and 'Deny', where it goes in the yaml, how it's part of `.spec.ingress` or `.spec.egress`.

I think it's important to explain that Allow or Deny rules give no direct feedback. You can test whether the rules work as you like, but you can't see what's actually happening. 

Adding a log rule with the same configuration will show you detailed information about the actual traffic.

ctauchen: Would there also be some use in explaining how to use log rules to test policy without implementing it? A DIY staged policy for open source?

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+
+:::caution
+
+## Key Behavior Differences: eBPF vs. iptables log actions
```
ctauchen, line 51: I don't think this needs a new heading or subsections. A separate paragraph, with a brief description, should be enough.

This is one that could benefit greatly from a diagram. That would make immediately clear how the eBPF logs are comprehensive, where the iptables logs just give you a particular point in the chain.

ctauchen: Consider explaining this with an example of traffic that gets evaluated by several policies. The log for iptables would be for the first rule only, right? And the eBPF log would show the final rule. Is that right?

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+</Tabs>
+
+
+## How to
+
+:::caution
+For a complete and efficient observability solution, consider exploring the use of [Flow Logs](/calico-enterprise/latest/visibility/visualize-traffic) available in Calico Enterprise and Calico Cloud.
+
+Using log rules is not be the most effective way to achieve comprehensive observability. While network observability can be achieved by utilizing log actions, this approach may introduce significant performance overhead.
+:::
```
ctauchen, line 128: You can add links like this to Additional resources at the bottom of the doc.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+    types:
+    - Ingress
+    - Egress
+    egress:
+    - action: Log
+    ingress:
+    - action: Log
+```
+
+### Logging in a namespace
```
ctauchen, line 165:
```suggestion
### Log all traffic for a namespace
```

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+To make the most of log actions, follow these guidelines:  
+
+    1. **Order Matters:** Assign log actions a high `order` value to ensure they are evaluated after all lower-priority policies.  
+    1. **Explicit Allow Rule:** Always pair a log action with an explicit `Allow` rule immediately after it. Otherwise, traffic may be denied due to the tier's implicit default action.  
+    1. **Comprehensive Protection:** Ensure both namespaced and non-namespaced resources are covered by enabling `hostEndpoint` for protecting your host in addition to Pods and workloads.  
+
+By strategically placing log actions at the bottom of your policy tiers, you can capture unhandled traffic and identify gaps in your policy configuration. Analyze the logged traffic to gather insights about source, destination, and protocol, and use this data to create targeted policies. Once implemented, validate by checking if the corresponding traffic is no longer logged, indicating the new policies are effective.  
+
+
+### Logging an unprotected cluster
```
ctauchen, line 142:
```suggestion
### Log all traffic for an unprotected cluster 
```

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+  types:
+  - Ingress
+  - Egress
+  egress:
+  - action: Log
+  ingress:
+  - action: Log
+```
+
+### Logging all traffic in a cluster
```
ctauchen, line 188: Log traffic for non-namespaced resources, hosts, and VMs

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+By default, $[prodname] policies apply only to namespaced resources (e.g., Pods, ServiceAccounts). To achieve total protection, enable hostEndpoint support to include non-namespaced resources and host nodes.
+
+To log all traffic involving host endpoints, enable auto-creation of host endpoints:
+```bash
+kubectl patch kubecontrollersconfiguration default --type=merge --patch='{"spec": {"controllers": {"node": {"hostEndpoint": {"autoCreate": "Enabled"}}}}}'
+```
+
+Please ensure you disable policy logging by removing the Log policies once you have finalized your environment security, as it can impact cluster performance.
```
ctauchen, line 197: This seems very important. You should make the full sequence clear from the beginning. 

I wonder whether we should have a section somewhere about lifecycle for log actions. Is it best to set and forget? Only run for a set period? What about the time range, data allowances, and so on? Even if this isn't the place for all that detail, it may be worth mentioning and linking from Additional resources.

ctauchen (commented):
Thanks for this! We badly needed information about log rules. Here's a few things to get you started.

The main things for this review are high-level structural items. If you have any questions, let me know. I've commented only on the vNext log-rules.mdx.

Editing prose in Github is made far simpler when the code is written with only one sentence per line. Usually doable with a quick regex.

calico/network-policy/policy-rules/log-rules.mdx
```diff
@@ -0,0 +1,206 @@
...
+### Iptables
+
+In the standard Linux dataplane where `iptables` like backends are used to implement cluster security, the Log action logs the traffic at the point of evaluation, before any subsequent actions (such as allowing or denying traffic) are applied. This provides granular visibility into the matching traffic for specific rules or policies.
+:::
+
+
+<Tabs groupId='log-rules'>
+<TabItem label="eBPF" value="eBPF">
+
+eBPF policy logs are sent to the trace pipe and can be viewed by using the following command:
```
ctauchen, line 68: Seems like this might be good to be broken out as a procedure, no?

View your logs, find your logs, retrieve logs, etc.

If there are different commands per dataplane, then duplicate the procedure section:

Find your logs (eBPF)
Find your logs (iptables)

ctauchen: Related to this, all the information about the logs themselves should have its own section up with the other conceptual information. Log format reference, or something like that.

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: I'm a bit confused are we talking about Kernel or Calico version number here?

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Is there a template file or an example that we should follow which I missed?

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Add the link to the "Additional resources" and keep this one too?

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Nice catch thanks!

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Nice catch thanks!

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Are we talking about removing tabs from the whole document or just this section?

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Love it!

calico/network-policy/policy-rules/log-rules.mdx
ctauchen: Sorry, grabbed too many lines. Just meant line 23, which explains the prerequisite on line 21.

calico/network-policy/policy-rules/log-rules.mdx
ctauchen: From the whole doc.

calico/network-policy/policy-rules/log-rules.mdx
ctauchen: There isn't, but there should be. I really should publicize this better.

calico/network-policy/policy-rules/log-rules.mdx
ctauchen: Just additional resources.

frozenprocess: So the idea here is if someone reads "flowlogs," which is a Calico enterprise thing, they can immediately click on it and go to the correspondent page to learn more.
Removing the hyperlink from that sentence and moving it only to the additional resources would make what we are talking about less visible since there will be no indication of a link at the bottom of the page for that footnote when the user is reading the sentence and when our user gets there they might not even remember what flowlogs was ...

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: Lines 23 and 21 are two different contexts. Line 21 is what you should run, and line 23 is if you are running a version and it's not working (Keep in mind distros can have different builds) and what is the cause.
In line 23 we say what is the function that has been invoked in the Kernel that makes everything possible.

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: I believe tabs are essential given that logs are different in each dataplane and we are trying to make eBPF more visible in the docs.

calico/network-policy/policy-rules/log-rules.mdx
frozenprocess: As we discussed in the 1on1, a new user should not try to enable the host endpoint right at the bat since it may break their cluster if they don't know what they are doing. Moving this step any higher would create a lot of support cases like I enabled logs and it broke my cluster.
In my opinion, the flow of discovery should be as it is written on the page.

ctauchen: I'm not suggesting a change to the actual workflow. What I'm saying is that it would be better for users if you first explain what the workflow is, and then show them how to do it.

In particular, I'm interested in the idea that you suggest here at the end. It seems that, at least in some cases, log rules are best used on a temporary basis. You set them, figure something else, and then remove them. That's an idea that would be good to present up at the top.

calico/network-policy/policy-rules/log-rules.mdx
ctauchen: I see what you mean. You're right that we need to clearly present the differences that occur with the two datap lanes. But I don't see that there's a good reason to use tabs in this case. The information can be mixed (as in the case of the prerequisites) or organized with simpler methods. Like paragraphs and headings.

calico/network-policy/policy-rules/log-rules.mdx
ctauchen: We can add these sorts of things to additional resources, where they are available for folks to check out if they're interested. But without unnecessarily distracting them from their main job: understanding log actions and how to use them.

I see that I wasn't entirely clear in my earlier note. It's not just the link. This admonition needs to go away. Best to condense to 1-2 sentences and add to additional resources.

ctauchen:
Reopening in #1827

PR #1813, opened 2024-12-13
Docs for Gateway API

nelljerram:
Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,325 @@
...
+kind: GatewayAPI
+metadata:
+  name: tigera-secure
+EOF
+```
+
+Shortly after this you will see that Gateway API resources are now available:
+
+```bash
+kubectl api-resources | grep gateway.networking.k8s.io
```
rene-dekker, line 78: Should we be relying on the new `tigerastatus` instead?

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,325 @@
...
+
+And also that there is a GatewayClass resource corresponding to the Envoy Gateway implementation included in {{prodname}}:
+
+```bash
+kubectl get gatewayclass -o yaml
+```
+
+### Use the Gateway API
+
+Now you can use the Gateway API in any way that the API supports.  When configuring a Gateway resource, be sure to specify `gatewayClassName: tigera-gateway-class`.
```
rene-dekker, line 111: Should we be calling it `tigera-gateway`? `-class` is redundant.

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,325 @@
...
+    imagePullPolicy: Always
+    name: c1
+  terminationGracePeriodSeconds: 0
+EOF
+```
+
+- Once that pod is running, install `curl` in it:
+
+```bash
+kubectl exec -it test-client -- apt-get -y update
```
rene-dekker: Should we combine the lines and remote `-it`?

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,325 @@
...
+
+## Value
+
+{{prodname}} includes support for the Kubernetes Gateway API, which allows advanced routing to services in a cluster, including weighted or blue-green load balancing.
+
+## Concepts
+
+### Gateway API
+
+The Gateway API is an official Kubernetes API for advanced routing to services in a cluster.  To read about its use cases, structure and design, please see [the official docs](https://gateway-api.sigs.k8s.io/).
```
rene-dekker: Should we specify which (minimum) gateway api version(s) we  support? I imagine that there may be conflicts with api versions, for example the grant is at v1beta1 now.

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,325 @@
...
+
+- Curl from the test client pod to a URL that should be handled by the echo server in namespace `ns1`, via the Gateway:
+
+```bash
+curl http://<cluster IP>/ns1/subpath?query=demo | jq
+```
+
+- The output confirms that the request was handled by the echo server in namespace `ns1`:
+
+```
```
rene-dekker: Is this supposed to be empty? Maybe if we expect an empty output, we can put it in text, rather than an empty code block? (I'm not even sure something will be rendered, the deploy preview didn't build unfortunately.)

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,325 @@
...
+
+- Curl from the test client pod to a URL that should be handled by the echo server in namespace `ns2`, via the Gateway:
+
+```bash
+curl http://<cluster IP>/ns2/subpath?query=demo | jq
+```
+
+- The output confirms that the request was handled by the echo server in namespace `ns2`:
+
+```
```
rene-dekker: Is this supposed to be empty as well?

calico-enterprise/networking/gateway-api.mdx
nelljerram: Thanks, I've added a table to show this.

calico-enterprise/networking/gateway-api.mdx
nelljerram: Sorry, this was work-in-progress.  It's filled in now.

calico-enterprise/networking/gateway-api.mdx
nelljerram: Sorry, this was work-in-progress. It's filled in now.

calico-enterprise/networking/gateway-api.mdx
nelljerram: I don't see a need to do that.  I agree it's slightly annoying that everyone puts `-it` everywhere even when not needed, but it does no harm.  (And I remember cases where `-i` is surprisingly needed.  Not sure if this could be one of those.)

calico-enterprise/networking/gateway-api.mdx
nelljerram: We could do.  But in this case I think the `-class` is helpful, on balance.  There are so many things that are called "gateway": the API, the controller, the Gateway objects, and the GatewayClass.  So here I think it's helpful to be more explicit.

calico-enterprise/networking/gateway-api.mdx
nelljerram: At this point I'm more interested in showing that the CRDs have been installed.  But certainly we could mention the "gatewayapi" tigerastatus object as well - would you like me to add that?

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,418 @@
...
+
+## Concepts
+
+### Gateway API
+
+The Gateway API is an official Kubernetes API for advanced routing to services in a cluster.  To read about its use cases, structure and design, please see [the official docs](https://gateway-api.sigs.k8s.io/).  {{prodname}} provides the following resources and versions of the Gateway API.
+
+| Resource         | Versions          |
+| ---------------- | ----------------- |
+| BackendLBPolicy  | v1alpha2          |
```
peterkellydev, line 29: I was going to ask if we should state a specific version of the gateway API we support and not document alpha and beta versions, so the list would just be 

Gateway, GatewayClass, HTTPRoute and GRPCRoute to start. Which are v1. 

But I guess we support all of the beta and alpha too via EnvoyGateway so there's no reason not to document that really, I can't think of any

peterkellydev (approved):
LGTM. 

This naming and sidebar etc is fine while PM still decides on the naming. 

I can see the sidebar being "Ingress Gateway" potentially (and positioned directly above / below "Egress Gateway" but we are waiting on PM to decide on what this feature is going to be called.

calico-enterprise/networking/gateway-api.mdx
rene-dekker: I think `tigerastatus` is a more  useful thing to use. If it succeeds, then we know CRDs are installed. If not, it provides you useful troubleshooting information.

calico-enterprise/networking/gateway-api.mdx
rene-dekker: We don't add suffixes to any resource we create in the operator, like storage classes, deployments, services, ... I think we should be consistent in naming things.

calico-enterprise/networking/gateway-api.mdx
rene-dekker: I'm confident that these don't need `it`.

rene-dekker: I think we should use https://hub.docker.com/r/alpine/curl image for this page to avoid  apt update and installing curl.

-- update -- 
Never mind, I see you need jq

calico-enterprise/networking/gateway-api.mdx
```diff
@@ -0,0 +1,418 @@
...
+    terminationGracePeriodSeconds: 0
+  EOF
+  ```
+
+- Once that pod is running, install `curl` in it:
+
+  ```bash
+  kubectl exec -it test-client -- apt-get -y update
+  kubectl exec -it test-client -- apt-get -y install curl
+  ```
```
gantony, line 221: Could use an image that already has curl installed. I typically use `wbitt/network-multitool` for that kind of stuff (not sure if there are better alternatives).

ctauchen:
Merging this now with plans to make improvements as discussed in coming weeks. Thanks for this, @nelljerram!

PR #1837, opened 2025-01-14
Add example for Role, RoleBinding and Secret to configure BGP passwords

### LINKED ISSUE #6893: Documentation: Cluster Role lacks Permissions for BGP Peer Password secretKeyRef
## Expected Behavior
Calico node can use configured password for bgp peer. 

## Current Behavior
BGP peer can't be configured:
```
E1022 16:11:52.490874     137 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.0/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets "redacted" is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot list resource "secrets" in API group "" in the namespace "kube-system"
W1022 16:11:53.018088     137 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.0/tools/cache/reflector.go:167: failed to list *v1.Secret: secrets "<redacted>" is forbidden: User "system:serviceaccount:kube-system:calico-node" cannot list resource "secrets" in API group "" in the namespace "kube-system"
```

## Possible Solution
Missing permission should be added to `https://github.com/projectcalico/calico/blob/master/charts/calico/templates/calico-node-rbac.yaml` (and other relevant locations)

we have added following permission as a quickfix:
```yaml
  - verbs:
      - list
      - get
      - watch
    apiGroups:
      - ''
    resources:
      - secrets
```
(although maybe not ideal & restricted to least priviledges)

## Steps to Reproduce (for bugs)
- Create BGP Peer according to documentation and using a password

paketb0te:
Product Version(s):
Calico

Issue:
fixes projectcalico/calico#6893

Link to docs preview:
[CLICK](https://deploy-preview-1837--calico-docs-preview-next.netlify.app/calico/next/reference/resources/bgppeer#keyref)

SME review:

DOCS review:

Additional information:

The Docs (in the reource definition for BGP peers) show that BGP Passwords must be stored in a Secret and then be referenced via the `KeyRef` (name of the secret and the key within that secret).

BUT it does not mention that this requires additional privileges for the `calico-node` ServiceAccount to access this secret.

It obviously makes sense if you think about it, but IMO it would make a lot of sense to mention that somewhere - see also @caseydavenport comments in projectcalico/calico#6893

I _think_ the included example might be a bit over the top, maybe there is a way to make it show only partially by default and expand on click or something?

I am not exactly familiar with mdx unfortunately.

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico/reference/resources/bgppeer.mdx
+++ b/calico/reference/resources/bgppeer.mdx
@@ -91,7 +91,7 @@ secret must be in the same namespace as the $[nodecontainer] pod.
 
 :::warning
 
-Calico must be able to read the referenced secret!
+$[prodname] must be able to read the referenced secret!
 
 This means that the `calico-node` ServiceAccount must have permissions to `list, get, watch` the secret referenced in the KeyRef.
 
@@ -105,7 +105,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: bgp-passwords-reader
-  namespace: kube-system
+  namespace: $NAMESPACE_IN_WHICH_CALICO_IS_RUNNING
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
@@ -117,7 +117,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: calico-read-bgp-passwords
-  namespace: kube-system
+  namespace: $NAMESPACE_IN_WHICH_CALICO_IS_RUNNING
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
@@ -131,7 +131,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: "bgp-passwords"
-  namespace: kube-system
+  namespace: $NAMESPACE_IN_WHICH_CALICO_IS_RUNNING
 data:
   peer_a_pw: "base64-encoded Password for Peer A"
   peer_b_pw: "base64-encoded Password for Peer B"
--- a/calico/reference/resources/bgppeer.mdx
+++ b/calico/reference/resources/bgppeer.mdx
@@ -89,6 +89,78 @@ secret must be in the same namespace as the $[nodecontainer] pod.
 | name  | The name of the secret    | string |
 | key   | The key within the secret | string |
 
+:::warning
+
+Calico must be able to read the referenced secret!
+
+This means that the `calico-node` ServiceAccount must have permissions to `list, get, watch` the secret referenced in the KeyRef.
+
+In practice, this can be done by creating a Role (which allows to `list, get, watch` the secret) and a RoleBinding (which grants the Role's permission to the `calico-node` ServiceAccount).
+
+Example:
+
+```yaml
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: bgp-passwords-reader
+  namespace: kube-system
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["bgp-passwords"]
+    verbs: ["list", "watch", "get"]
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: calico-read-bgp-passwords
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: bgp-passwords-reader
+subjects:
+  - kind: ServiceAccount
+    name: calico-node
+
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "bgp-passwords"
+  namespace: kube-system
+data:
+  peer_a_pw: "base64-encoded Password for Peer A"
+  peer_b_pw: "base64-encoded Password for Peer B"
+
+---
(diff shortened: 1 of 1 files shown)


ctauchen:
Thanks for this contribution, @paketb0te! I'll let @caseydavenport vet the technical side before I follow up with a docs review.

ctauchen:
@caseydavenport I expect this new information would be universal. Am I right thinking we should eventually pick these changes to all products / versions?

calico/reference/resources/bgppeer.mdx
```diff
@@ -89,6 +89,78 @@ secret must be in the same namespace as the $[nodecontainer] pod.
...
+
+Example:
+
+```yaml
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: bgp-passwords-reader
+  namespace: kube-system
```
caseydavenport: I think we want to default to calico-system (most new installs use the calico-system namespace).

Perhaps this is a use-case for Tabs for "Operator" and "Manifest" installs? We have those elsewhere in the docs.

paketb0te: Sure, I don't mind which one is used as example :)

Maybe we could even make it extra-explicit by using something like `$NAMESPACE_IN_WHICH_CALICO_IS_RUNNING`

That would 
a) make it non-copy-paste-able, and thus 
b) ensure that people look up the namespace calico is actually running in.

I like these Tabs for the different installation methods because they help keep the docs a bit clearer, but I also think they are better used in the "explanatory" sections of the docs and could be confusing in the Resource Definition Reference.

caseydavenport: I'd be happy with that approach :) 

I'll leave it up to @ctauchen as to what the preferred style guideline would be in this case.

paketb0te: Updated the namespace for the keyref example as discussed

calico/reference/resources/bgppeer.mdx
```diff
@@ -89,6 +89,78 @@ secret must be in the same namespace as the $[nodecontainer] pod.
...
+
+Example:
+
+```yaml
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: bgp-passwords-reader
+  namespace: $NAMESPACE_IN_WHICH_CALICO_IS_RUNNING
```
ctauchen:
```suggestion
  namespace: kube-system # Or calico-system, if you installed Calico using the manifest method.
```
Let's go with something like this to keep the actual names. And some explanation about why they might be different.

calico/reference/resources/bgppeer.mdx
```diff
@@ -89,6 +89,78 @@ secret must be in the same namespace as the $[nodecontainer] pod.
...
+    resources: ["secrets"]
+    resourceNames: ["bgp-passwords"]
+    verbs: ["list", "watch", "get"]
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: calico-read-bgp-passwords
+  namespace: $NAMESPACE_IN_WHICH_CALICO_IS_RUNNING
```
ctauchen: See above.

calico/reference/resources/bgppeer.mdx
```diff
@@ -89,6 +89,78 @@ secret must be in the same namespace as the $[nodecontainer] pod.
...
+subjects:
+  - kind: ServiceAccount
+    name: calico-node
+
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "bgp-passwords"
+  namespace: $NAMESPACE_IN_WHICH_CALICO_IS_RUNNING
```
ctauchen: See above.

calico/reference/resources/bgppeer.mdx
```diff
@@ -89,6 +89,78 @@ secret must be in the same namespace as the $[nodecontainer] pod.
 | name  | The name of the secret    | string |
 | key   | The key within the secret | string |
 
+:::warning
+
+$[prodname] must be able to read the referenced secret!
```
ctauchen:
```suggestion
$[prodname] must be able to read the referenced secret.
```

ctauchen (commented):
A few suggestions.

calico/reference/resources/bgppeer.mdx
caseydavenport: I am good with this, but two things:

- It should default to calico-system
- and should say "kube-system if installed using manifests"

paketb0te: @ctauchen do you update your suggestion as per @caseydavenport 's update, and I then commit that directly in Github?

Or should I just add your suggestions locally and then push (which I think will remove the comments here)?

Sorry I am not familiar with the workflow here, working mostly on gitlab... :stuck_out_tongue_closed_eyes:

ctauchen: Thanks @caseydavenport. @paketb0te , make the changes as Casey suggests and push. Comments will be just fine.

paketb0te: updated :)

ctauchen:
Many thanks, @paketb0te . LGTM.

PR #1844, opened 2025-01-17
add deprecation notice to FIPS in OSS (enterprise page is gone)

aaaaaaaalex:
[CORE-10911](https://tigera.atlassian.net/browse/CORE-10911)

Product Version(s):
OSS next

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed 

[CORE-10911]: https://tigera.atlassian.net/browse/CORE-10911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/calico/operations/fips.mdx
+++ b/calico/operations/fips.mdx
@@ -1,5 +1,5 @@
 ---
-description: Deprecated - Run Calico using FIPS validated cryptography.
+description: Run Calico using FIPS validated cryptography.
 ---
 
 # FIPS mode
--- a/calico/operations/fips.mdx
+++ b/calico/operations/fips.mdx
@@ -4,8 +4,10 @@ description: Run Calico using FIPS validated cryptography.
 
 # FIPS mode
 
-:::warning
-$[prodname] FIPS is now deprecated.
+:::info deprecation notice
+
+FIPS mode is deprecated and will be removed in a future release.
+
 :::
 
 ## Big picture
--- a/calico/operations/fips.mdx
+++ b/calico/operations/fips.mdx
@@ -1,8 +1,12 @@
 ---
-description: Run Calico using FIPS validated cryptography.
+description: Deprecated - Run Calico using FIPS validated cryptography.
 ---
 
 # FIPS mode
+
+:::warning
+$[prodname] FIPS is now deprecated.
+:::
 
 ## Big picture
 


ctauchen:
Thanks @aaaaaaaalex. I know we deprecated and removed FIPS support in CE and CC. Last I remember we were keeping this in OSS because it might work, even if we don't support it. My memory may be off though. 

Does this change stem from code changes that will kill this function entirely?

aaaaaaaalex:
@ctauchen , while I haven't ripped out any FIPS code, I have been working on a FIPS-adjacent customer issue and was advised to add this notice.

> Last I remember we were keeping this in OSS because it might work, even if we don't support it.

I'll CC @fasaxc and @matthewdupre here - I think we were of the belief that this FIPS documentation was left in by accident. Since I haven't killed off any FIPS functionality, I would imagine your reasoning still holds true. Do we want to carry on with this change now?

ctauchen:
Ok, thanks for that context. If we do want this gone, we might go ahead and add the deprecation notice to 3.29 with scheduled removal in 3.30.

matthewdupre:
We left the docs so far because both:
1. It was hard to find the time to remove (this note is the most important thing to put in asap, but we should also have an upgrade note telling people they'll need to switch to using the regular images).
2. We didn't have any particular reason to believe it was going to stop working.

We now know that it is going to stop "working" (since the certification will expire and we don't intend to renew), and are doing work in the area so it's good to bundle in the cleanup, so neither reason really applies any more.

rene-dekker:
Some internal slack discussion can be found here on why we removed it from EE: https://tigera.slack.com/archives/C027Z8Z9RM0/p1704985376286959
Basically, Nist put out a requirement that TLS 1.3 is required since Jan2024, but our cryptographic module was not validated with 1.3.

calico/operations/fips.mdx
```diff
@@ -1,9 +1,13 @@
 ---
-description: Run Calico using FIPS validated cryptography.
+description: Deprecated - Run Calico using FIPS validated cryptography.
```
ctauchen:
```suggestion
description: Run Calico using FIPS validated cryptography.
```
No need here.

calico/operations/fips.mdx
```diff
@@ -1,9 +1,13 @@
...
 ---
 
 # FIPS mode
 
+:::warning
+$[prodname] FIPS is now deprecated.
+:::
```
ctauchen:
```suggestion
:::info deprecation notice

FIPS mode is deprecated and will be removed in a future release.

:::
```

ctauchen: To bring this in line with similar notices for feature pages.

ctauchen (commented):
Changes to bring this in line with similar deprecation notices.

PR #1858, opened 2025-01-23
[RS-2261] Add version info for new Gateway API images

nelljerram:
Product Version(s):
Calico Enterprise

Issue:
https://tigera.atlassian.net/browse/RS-2261

### DIFF
--- a/calico-enterprise/reference/component-versions.mdx
+++ b/calico-enterprise/reference/component-versions.mdx
@@ -9,3 +9,10 @@ import ComponentVersions from '@site/calico-enterprise/_includes/components/Comp
 This page lists the specific component versions that go into each release of $[prodname].
 
 <ComponentVersions />
+
+Where components are built from third party sources, the following table indicates the source and upstream version.
+
+| Component       | Source                                  | Version  |
+| --------------- | --------------------------------------- | -------- |
+| envoy-gateway   | https://github.com/envoyproxy/gateway   | v1.1.2   |
+| envoy-ratelimit | https://github.com/envoyproxy/ratelimit | 26f28d78 |
--- a/calico-enterprise/releases.json
+++ b/calico-enterprise/releases.json
@@ -211,8 +211,16 @@
         "image": "tigera/envoy",
         "version": "master"
       },
+      "envoy-gateway": {
+        "image": "tigera/envoy-gateway",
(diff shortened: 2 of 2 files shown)


ctauchen:
@nelljerram Thanks for this. Did you consider just adding these images to releases.json with the appropriate version numbers? That would continue the existing paradigm and save us considerable maintenance. Those version image version numbers are automated, where this new separation would require hand-coding for each release.

The only new information type is the link to the github source. If that's required, then we'll find a way to make this work. If it isn't, I think we should consider going with the default.

CC @rene-dekker

nelljerram:
> @nelljerram Thanks for this. Did you consider just adding these images to releases.json with the appropriate version numbers?

Yes, it has occurred to me.  The problem is that I don't know what other effects that might have.  I'm also not sure if you mean overwriting the value in the existing `version` field, or adding a new `upstream_version` field.

> That would continue the existing paradigm and save us considerable maintenance. Those version image version numbers are automated, where this new separation would require hand-coding for each release.

Maybe, but I need to be sure what we mean first.

> The only new information type is the link to the github source. If that's required, then we'll find a way to make this work. If it isn't, I think we should consider going with the default.

What do you mean by "going with the default" ?

ctauchen:
By 'default', I mean the system we currently have in place. We have components listed in `releases.json` with their version numbers. This system includes both internal and third-party components.

You've added some new third-party components. You could simply make sure this gets integrated into the current (default) system. From what I can tell, that would work fine.

The only thing the default setup doesn't do is add the links. If those are required, we can't use the default setup.

nelljerram:
@ctauchen and I discussed this PR yesterday, and agreed to close it, and leave the possibility of documenting our upstream versions until a clearer need comes along.  The factors involved in this conclusion were:
- It was suggested that the upstream version could go in the existing table instead of the Calient version (as is already the case for elastic), but we established that that would be wrong for the Envoy images, because the same underlying data is used for our private registry docs, and someone setting up a private registry should definitely use the Calient version for our Envoy images.
- The suggestion of adding all the other existing third-party cases into the new table is non-trivial, as it might not be straightforward to identify the upstream version in each case.
- I believe it was only ever a nice idea and not a hard requirement.
- So on balance, better to leave this for now.

PR #1931, opened 2025-02-28
Update support & compatibility page for 3.19.5

sabags:
Product Version(s):
Calico Enterprise 3.19.5

Issue:
Update support & compatibility page

Link to docs preview:
N/A

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise_versioned_docs/version-3.19-2/getting-started/compatibility.mdx
+++ b/calico-enterprise_versioned_docs/version-3.19-2/getting-started/compatibility.mdx
@@ -53,7 +53,7 @@ Kubernetes version support aligns with [upstream Kubernetes](#kubernetes-kubeadm
 
 | $[prodname] version | kOps and Kubernetes versions | $[prodname] support                                                                   |
 | -------------------- | ---------------------------- | -------------------------------------------------------------------------------------- |
-| 3.19                 | 1.27 - 1.29                  | - $[prodname] CNI with network policy<br />- AWS CNI with $[prodname] network policy |
+| 3.19                 | 1.28 - 1.30                  | - $[prodname] CNI with network policy<br />- AWS CNI with $[prodname] network policy |
 | 3.18                 | 1.26 - 1.28                  | - $[prodname] CNI with network policy<br />- AWS CNI with $[prodname] network policy |
 | 3.17                 | 1.25 - 1.26                  | - $[prodname] CNI with network policy<br />- AWS CNI with $[prodname] network policy |
 
@@ -85,7 +85,7 @@ Kubernetes version support aligns with [upstream Kubernetes](#kubernetes-kubeadm
 
 | $[prodname] version | RKE version | $[prodname] support                 | Kubernetes versions |
 | -------------------- | ----------- | ------------------------------------ | ------------------- |
-| 3.19                 | 1.5         | $[prodname] CNI with network policy | 1.28                |
+| 3.19                 | 1.6         | $[prodname] CNI with network policy | 1.30                |
 | 3.18                 | 1.4         | $[prodname] CNI with network policy | 1.26                |
 | 3.17                 | 1.4         | $[prodname] CNI with network policy | 1.25                |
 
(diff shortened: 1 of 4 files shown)


sabags:
@sujeet-kr : can you take a look ?

ctauchen (commented):
Do you really mean to change supported dependences BETWEEN patch updates to 3.19? That's unusual. And a very unwelcome change for affected customers.

Also, we don't document these items at the level of specific patches. By making this change, we overwrite the previous supported version for earlier patches. A customer who remains on 3.19.2, say will read this document and assume that they must be on these versions of K8s. 

If you intend to proceed with these changes, we should probably schedule a short meeting to discuss. 

@sabags @sujeet-kr

calico-enterprise_versioned_docs/version-3.19-2/getting-started/compatibility.mdx
```diff
@@ -85,7 +85,7 @@ Kubernetes version support aligns with [upstream Kubernetes](#kubernetes-kubeadm
 
 | $[prodname] version | RKE version | $[prodname] support                 | Kubernetes versions |
 | -------------------- | ----------- | ------------------------------------ | ------------------- |
-| 3.19                 | 1.5         | $[prodname] CNI with network policy | 1.28                |
+| 3.19                 | 1.6         | $[prodname] CNI with network policy | 1.30                |
```
sabags, line 88: I think this should be something like this

```suggestion
| 3.19                 | 1.5, 1.6         | $[prodname] CNI with network policy | 1.28 - 1.30                |
```

calico-enterprise_versioned_docs/version-3.19-2/getting-started/compatibility.mdx
```diff
@@ -53,7 +53,7 @@ Kubernetes version support aligns with [upstream Kubernetes](#kubernetes-kubeadm
 
 | $[prodname] version | kOps and Kubernetes versions | $[prodname] support                                                                   |
 | -------------------- | ---------------------------- | -------------------------------------------------------------------------------------- |
-| 3.19                 | 1.27 - 1.29                  | - $[prodname] CNI with network policy<br />- AWS CNI with $[prodname] network policy |
+| 3.19                 | 1.28 - 1.30                  | - $[prodname] CNI with network policy<br />- AWS CNI with $[prodname] network policy |
```
sabags, line 56: I don't think we ever support N-3 K8s versions so I don't think the base supported version can be 1.27 here. 

wdyt @sujeet-kr ?

sabags:
> Do you really mean to change supported dependences BETWEEN patch updates to 3.19? That's unusual. And a very unwelcome change for affected customers.
> 
> Also, we don't document these items at the level of specific patches. By making this change, we overwrite the previous supported version for earlier patches. A customer who remains on 3.19.2, say will read this document and assume that they must be on these versions of K8s.
> 
> If you intend to proceed with these changes, we should probably schedule a short meeting to discuss.
> 

One of the main reasons for this patch is a customer ask to see if 3.19 works with RKE 1.30. So we tested 3.19 with RKE 1.30 and as a result we also want to call out support for k8s 1.30 in the docs. 

I have added some suggestions to the changes I made. Do you think if we do that way it will work @ctauchen  ?

calico-enterprise_versioned_docs/version-3.19-2/getting-started/compatibility.mdx
sujeet-kr: I think this should be fine. Anyhow our support matrix for 3.19-GA+ is k8s 1.28 to 1.30, so we are in-line.

calico-enterprise_versioned_docs/version-3.19-2/getting-started/compatibility.mdx
sujeet-kr: Ideally it should only be 1.5, but since RKE does not support k8s 1.30 with 1.5 we needed to move to 1.6. Think it should be ok to keep it at only 1.6.

sujeet-kr (approved):
LGTM

ctauchen:
>One of the main reasons for this patch is a customer ask to see if 3.19 works with RKE 1.30. So we tested 3.19 with RKE 1.30 and as a result we also want to call out support for k8s 1.30 in the docs.

I don't mind ADDING a new version. But REMOVING the oldest version could be a problem for users. Also, unless we specified, a reader would reasonably assume that 1.30 works with ALL 3.19.x versions. Is that correct?

@sabags

PR #1939, opened 2025-03-05
[RS-2374] Add docs for Gateway API in Calico OSS

nelljerram:
This is the same as #1813, except using GatewayAPI resource name "default" instead of "tigera-secure".

### DIFF
--- a/calico-enterprise/networking/gateway-api.mdx
+++ b/calico-enterprise/networking/gateway-api.mdx
@@ -2,7 +2,7 @@
 description: Enable support for the Kubernetes Gateway API.
 ---
 
-# Gateway API
+# Configure an ingress gateway
 
 ## Big picture
 
--- a/calico/networking/gateway-api.mdx
+++ b/calico/networking/gateway-api.mdx
@@ -1,0 +1,418 @@
+---
+description: Enable support for the Kubernetes Gateway API.
+---
+
+# Gateway API
+
+## Big picture
+
+Enable support for the Kubernetes Gateway API.
+
+:::note
+
+This feature is tech preview. Tech preview features may be subject to significant changes before they become GA.
+
+:::
+
+## Value
+
+$[prodname] includes support for the Kubernetes Gateway API, which allows advanced routing to services in a cluster, including weighted or blue-green load balancing.
+
+## Concepts
+
+### Gateway API
+
+The Gateway API is an official Kubernetes API for advanced routing to services in a cluster.  To read about its use cases, structure and design, please see [the official docs](https://gateway-api.sigs.k8s.io/).  $[prodname] provides the following resources and versions of the Gateway API.
+
+| Resource         | Versions          |
+| ---------------- | ----------------- |
+| BackendLBPolicy  | v1alpha2          |
+| BackendTLSPolicy | v1alpha3          |
+| GatewayClass     | v1, v1beta1       |
+| Gateway          | v1, v1beta1       |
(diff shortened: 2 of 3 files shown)


nelljerram:
@ctauchen Do you know if I need to take action to get an equivalent of #1858 for OSS?  I'm hoping that might just happen automatically at OSS release time.

nelljerram:
@ctauchen Also I was expecting that the 3.21 Enterprise docs would already include the GatewayAPI resource in the Installation reference doc [here](https://docs.tigera.io/calico-enterprise/3.21/reference/installation/api), and was going to ask about a similar inclusion in the OSS doc [here](https://docs.tigera.io/calico/latest/reference/installation/api).  But in fact it's missing in the 3.21 Enterprise docs.  Do you know what we need to do to remedy that, and to make sure that the same material is included for OSS 3.30 ?

ctauchen:
>Do you know if I need to take action to get an equivalent of https://github.com/tigera/docs/pull/1858 for OSS? I'm hoping that might just happen automatically at OSS release time.

No, there is no automated mechanism for OSS, which generally has much fewer images. You'll need to add these manually.

 >Also I was expecting that the 3.21 Enterprise docs would already include the GatewayAPI resource in the Installation reference doc [here](https://docs.tigera.io/calico-enterprise/3.21/reference/installation/api), and was going to ask about a similar inclusion in the OSS doc [here](https://docs.tigera.io/calico/latest/reference/installation/api). But in fact it's missing in the 3.21 Enterprise docs. Do you know what we need to do to remedy that, and to make sure that the same material is included for OSS 3.30 ?
 
Those docs are generated automatically, and I know that they successfully pulled in some 3.21 changes before the release. So I'm not sure why yours didn't get picked up. I'll try running it again and put up a PR with changes. But if that doesn't work, we'll need to talk to someone who understands exactly where those changes come from. (It's all a black box to me).

ctauchen:
For this PR, LGTM, and happy to merge when you're ready.

nelljerram:
Thanks @ctauchen .  If it's OK with you, I'll look first at whether I can easily add the changes like in #1858 (but for OSS)...

nelljerram:
OK, my mistake here; it appears that no further changes are needed, because:

- In fact we agreed to close #1858 without merging it.
- It appears there is already a process for incorporating a correct list of images into pages like https://docs.tigera.io/calico-enterprise/3.21/getting-started/install-on-clusters/private-registry/private-registry-regular, without needing a prior update to https://github.com/tigera/docs/blob/main/calico-enterprise/releases.json, and I assume the same will be true for [OSS](https://docs.tigera.io/calico/3.28/operations/image-options/alternate-registry) as well.

nelljerram:
@ctauchen I'm happy now for you to merge this whenever you think appropriate.

ctauchen:
@nelljerram Does that mean we do not expect to see changes related to gateway api in the installation api reference?

nelljerram:
> @nelljerram Does that mean we do not expect to see changes related to gateway api in the installation api reference?

I may not be understanding your exact point.  But the installation API reference _should_ definitely include the GatewayAPI resource, both for Enterprise 3.21 and OSS 3.30.
- The Enterprise 3.21 docs already exist, and should include GatewayAPI, but currently do not; that is a bug that we should aim to fix asap.  
- The OSS 3.30 docs don't exist yet, but when they are created, they should also include GatewayAPI.

PR #1971, opened 2025-03-26
[CORE-10980] Add documentation for QoS Controls

coutinhop:
Add documentation for QoS Controls and remove Kubernetes bandwidth plugin ("traffic shaping")

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico/networking/configuring/qos-controls.mdx
+++ b/calico/networking/configuring/qos-controls.mdx
@@ -14,7 +14,7 @@ With QoS controls, $[prodname] enforces network resource limits (bandwidth, pack
 
 ## Limitations
 
-- This feature is currently supported on the iptables and nftables modes only, eBPF mode does not support it at the moment.
+- This feature is currently supported on the **iptables** and **nftables** modes only, **eBPF** mode does not support it at the moment.
 
 ## Concepts
 
--- a/calico/networking/configuring/qos-controls.mdx
+++ b/calico/networking/configuring/qos-controls.mdx
@@ -1,0 +1,120 @@
+---
+description: Configure QoS (Quality of Service) Controls to limit ingress and/or egress bandwidth, packet rate and number of connections of Calico workloads.
+---
+
+# Configure QoS Controls
+
+## Big picture
+
+Configure QoS (Quality of Service) Controls to limit ingress and/or egress bandwidth, packet rate and number of connections used by Calico workloads in order to prevent them from overusing or overloading network resources.
+
+## Value
+
+With QoS controls, $[prodname] enforces network resource limits (bandwidth, packet rate, connections) for Kubernetes pods and OpenStack VMs, ensuring fair resource allocation and preventing performance degradation for other workloads.
+
+## Limitations
+
+- This feature is currently supported on the iptables and nftables modes only, eBPF mode does not support it at the moment.
+
+## Concepts
+
+### Quality of Service Controls
+
+**Quality of Service** (**QoS**), in the field of computer networking, refers to traffic prioritization and resource reservation control mechanisms. It may either mean guaranteeing minimum levels of networking resources, or limiting the maximum levels of usage of networking resources in order to prevent a small number of users from excessive resource consumption, stopping others from using their fair share. For practical purposes, the latter will be considered as $[prodname] implements QoS Controls by limiting maximum levels of network resources usage.
+
+$[prodname] implements three types of QoS Controls, which can be configured using annotations on the Kubernetes pods for either ingress (traffic coming into the pod), egress (traffic going out of the pod), or both:
+1. Bandwidth: limiting the bit rate of traffic into/out of the pod
+2. Packet rate: limiting the number of packets per second that can be sent or received by the pod
+3. Number of established connections: limiting the absolute number of connections that the pod may initiate or accept
+
+### Annotations
+
+On Kubernetes, $[prodname] uses [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) on pods for the configuration of QoS controls.
+
+The values of the QoS controls annotations may contain suffixes, such as `k`, `M`, `G`, `T`, `P`, or `E` to represent large numbers. Binary suffixes are also supported, for example `1Ki` represents "1024".
+
+For example, to limit egress bandwidth to 100Mbps, with a maximum burst of 200Mb:
+
+```yaml noValidation
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+  labels:
+    app: my-app
+spec:
+  replicas: 3
+  template:
+    metadata:
+      annotations:
+        qos.projectcalico.org/egressBandwidth: "100M"
+        qos.projectcalico.org/egressBurst: "200M"
+    spec:
+        (...)
+```
+
+## How to
+
+- [Configure bandwidth limits](#configure-bandwidth-limits)
+- [Configure packet rate limits](#configure-packet-rate-limits)
+- [Configure established connection limits](#configure-established-connection-limits)
+- [See all available annotations](#all-available-annotations)
+
+### Configure bandwidth limits
+
+Use the annotations below to configure bandwidth and burst limits for traffic going into or coming out of the pod. Limiting the bandwidth enforces that, on average, that will be the maximum bit rate used. Limiting the burst enforces that at any given instant, the maximum burst traffic will not go over the burst limit. Burst configuration is optional, if burst is not explicitly configured with an annotation, a default value of `4Gib` is used.
+
+| Annotation                                    | Description                                                                                                                                                                                               |
+|-----------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `qos.projectcalico.org/ingressBandwidth`      | Specifies the ingress bandwidth limit (in bits per second).<br/>Min value: `1k`<br/>Max value: `1P`.                                                                                                      |
+| `qos.projectcalico.org/egressBandwidth`       | Specifies the egress bandwidth limit (in bits per second).<br/>Min value: `1k`<br/>Max value: `1P`.                                                                                                       |
+| `qos.projectcalico.org/ingressBurst`          | Specifies the ingress burst limit (in bits).<br/>Optional, defaults to `4Gi`.<br/>Min value: 2 * value of `ingressBandwidth`<br/>Max value: `4Gi`                                                         |
+| `qos.projectcalico.org/egressBurst`           | Specifies the egress burst limit (in bits).<br/>Optional, defaults to `4Gi`.<br/>Min value: 2 * value of `egressBandwidth`<br/>Max value: `4Gi`                                                           |
+
+If the [Kubernetes bandwidth plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/) annotations `kubernetes.io/ingress-bandwidth` or `kubernetes.io/egress-bandwidth` are configured and the `qos.projectcalico.org` annotations are not, they are honored and $[prodname] will configure bandwidth limits based on them . The $[prodname] specific annotations take precedence, so if both are configured, the Kubernetes annotation will be ignored. This makes upgrading from a $[prodname] version that uses the bandwidth plugin to a version with native bandwith limiting seamless, as the behavior will remain consistent across versions (with $[prodname] replacing the functionality previously provided by the Kubernetes bandwidth plugin).
+
+One essential difference between the bandwidth plugin and $[prodname]'s implementations is that with $[prodname], the bandwidth limits take effect immediately, whereas with the bandwidth plugin the pods need to be recreated for the limits to take effect.
+
+| Annotation                                    | Description                                                                                                                                                                                               |
+|-----------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `kubernetes.io/ingress-bandwidth`             | Kubernetes bandwidth plugin annotation for the ingress bandwidth limit (in bits per second).<br/>Ignored if `qos.projectcalico.org/ingressBandwidth` is present.<br/>Min value: `1k`<br/>Max value: `1P`  |
+| `kubernetes.io/egress-bandwidth`              | Kubernetes bandwidth plugin annotation for the egress bandwidth limit (in bits per second).<br/>Ignored if `qos.projectcalico.org/egressBandwidth` is present.<br/>Min value: `1k`<br/>Max value: `1P`    |
+
+### Configure packet rate limits
+
+Use the annotations below to configure packet rate limits for traffic going into or coming out of the pod. Limiting the packet rate enforces that, on average, that will be the maximum rate that packets will be sent and/or received by the pod.
+
+| Annotation                                    | Description                                                                                                                                                                                               |
+|-----------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `qos.projectcalico.org/ingressPacketRate`     | Specifies the maximum rate of ingress packets (in packets per second).<br/>Min value: `10`<br/>Max value: `1T`                                                                                            |
+| `qos.projectcalico.org/egressPacketRate`      | Specifies the maximum rate of egress packets (in packets per second).<br/>Min value: `10`<br/>Max value: `1T`                                                                                             |
+
+### Configure established connection limits
+
+Use the annotations below to configure limits for the number of established connections into or out of a pod. Limiting the number of connections enforces that the pod cannot accept or initiate more connections that the specified limits.
+
+| Annotation                                    | Description                                                                                                                                                                                               |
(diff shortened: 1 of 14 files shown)


calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+### Configure packet rate limits
+
+To configure packet rate limits,  use the `qos.projectcalico.org/ingressPacketRate` annotation for ingress, or `qos.projectcalico.org/egressPacketRate` for egress. Packet rate values are in packets per second and constrained from a minimum of `10` to a maximum of `1T` (10<sup>12</sup>) packets per second.
+
+### Configure established connection limits
+
+To configure limits for the number of established connections,  use the `qos.projectcalico.org/ingressMaxConnections` annotation for ingress, or `qos.projectcalico.org/egressMaxConnections` for egress. The values are in absolute numbers of simultaneous established connections and constrained from a minimum of `1` to a maximum of `100G` (100 * 10<sup>9</sup>) connections.
```
ctauchen: >a minimum of `10` to a maximum of `1T` (10<sup>12</sup>) packets per second

>a minimum of `1` to a maximum of `100G` (100 * 10<sup>9</sup>) connections

I've never encountered this practice of using SI prefixes to designate numbers, rather than modifying units (like TB or GHz.

Can we just use regular numbers? 

Do you expect/require that this style is used as values for your annotations?

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+### All available annotations
+
+Below is a table listing all available QoS controls annotations.
+
+| Annotation                                    | Description                                                                                 |
```
ctauchen: Given the number question earlier, and the fact that there may be very large numbers involved, it might be good to define the value type and defaults in this table.

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+## Concepts
+
+### Quality of Service Controls
+
+**Quality of Service**, or **QoS** for short, in the field of computer networking, refers to traffic prioritization and resource reservation control mechanisms. It may either mean guaranteeing minimum levels of networking resources, or limiting the maximum levels of usage of networking resources in order to prevent it from being overloaded by a small number of actors. For practical purposes, the latter will be considered as $[prodname] implements QoS Controls in this manner.
+
+### Annotations
+
+On Kubernetes, $[prodname] uses annotations on pods for the configuration of QoS controls. Pods can be directly annotated with `kubectl annotate`, created with the annotatios in their metadata or, if they're part of a deployment or daemonset, have the annotations in the metadata of their template spec. For example:
```
ctauchen:
```suggestion
On Kubernetes, $[prodname] uses annotations on pods for the configuration of QoS controls. Pods can be directly annotated with `kubectl annotate`, created with the annotations in their metadata or, if they're part of a deployment or daemonset, have the annotations in the metadata of their template spec. For example:
```

ctauchen: The second sentence would be clearer with this structure:

```
You can annotate pods in any of these ways:
* Do this
* Do this other thing
* Do this.

File: calico/networking/configuring/qos-controls.mdx [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+
+# Configure QoS Controls
+
+## Big picture
+
+Configure QoS (Quality of Service) Controls to limit ingress and/or egress bandwidth, packet rate and number of connections used by Calico workloads in order to prevent them from overusing or overloading network resources.
+
+## Value
+
+$[prodname] can configure limits as to how much network resources a workload (Kubernetes pod or OpenStack VM) can use for either ingress, egress or both: bandwidth, packet rate and number of established connections. This allows you to prevent workloads from overusing these network resources and thus from negatively affecting the quality of service of other workloads in the cluster.
```
ctauchen:
```suggestion
With QoS controls, [prodname] enforces network resource limits (bandwidth, packet rate, connections) for Kubernetes pods and OpenStack VMs, ensuring fair resource allocation and preventing performance degradation for other workloads.
```

If you like!

ctauchen (commented):
Thanks again for this, @coutinhop! I've got a few suggestions for you to consider.

calico/networking/configuring/qos-controls.mdx
coutinhop: This is because we use the k8s resource package to parse the quantites: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity

We do not require this style, but do expect it seeing as it is both more convenient for large numbers and already in use by the k8s bandwidth plugin...

Perhaps we could make this a bit clearer by explicitly stating we support parsing quantities as defined by the resource package, what do you think?

calico/networking/configuring/qos-controls.mdx
coutinhop: Good point! Types, defaults and min/max acceptable values, right? Will do

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+## Value
+
+With QoS controls, $[prodname] enforces network resource limits (bandwidth, packet rate, connections) for Kubernetes pods and OpenStack VMs, ensuring fair resource allocation and preventing performance degradation for other workloads.
+
+## Concepts
+
+### Quality of Service Controls
+
+**Quality of Service**, or **QoS** for short, in the field of computer networking, refers to traffic prioritization and resource reservation control mechanisms. It may either mean guaranteeing minimum levels of networking resources, or limiting the maximum levels of usage of networking resources in order to prevent it from being overloaded by a small number of actors. For practical purposes, the latter will be considered as $[prodname] implements QoS Controls in this manner.
```
tomastigera: Title says QoS, then Big Picture says `Configure QoS (Quality of Service)` and here you define QoS again in reverse :man_shrugging: Imo QoS is well established - if not, using it the Title is too early. But once you define it, you can just keep using QoS

```suggestion
**Quality of Service (QoS)** for short, in the field of computer networking, refers to traffic prioritization and resource reservation control mechanisms. It may either mean guaranteeing minimum levels of networking resources, or limiting the maximum levels of usage of networking resources in order to prevent it from being overloaded by a small number of actors. For practical purposes, the latter will be considered as $[prodname] implements QoS Controls in this manner.
```

tomastigera:
```suggestion
**Quality of Service**, or **QoS** for short, in the field of computer networking, refers to traffic prioritization and resource reservation control mechanisms. It may either mean guaranteeing minimum levels of networking resources, or limiting the maximum levels of usage of networking resources in order to prevent a small number of users from excessive resource consumption while preventing others from using their fair share. For practical purposes, the latter will be considered as $[prodname] implements QoS Controls in this manner.
```

> for practical purposes, the latter will be considered as $[prodname] implements QoS Controls in this manner.

not quite sure what this means. State explicitly what calico does.

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+### Annotations
+
+On Kubernetes, $[prodname] uses annotations on pods for the configuration of QoS controls. You can annotate pods in any of these ways:
+- pods can be directly annotated with `kubectl annotate`;
+- they can be created with the annotations in their metadata;
+- if they're part of a deployment or daemonset, they can have the annotations in the metadata of their template spec.
```
tomastigera: you start explaining annotations before you say what they are for in QoS. Imo, it is a k8s concept and a link to how annotations are used should suffice.

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+The values of the QoS controls annotations are parsed using the [Quantity type of the resource package](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity),
+so they may contain suffixes, such as `k`, `M` or `G` (`1k` represents "1000", `1Ki` represents "1024", and so on).
```
tomastigera: Again, too early. Does not seem like a user need to know what package parses the annotations, what is important is to say (at the right place) what is the syntax. What you so not say is whether it is in bytes or bits. My assumption would be bits per second since it is bandwidth.

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,91 @@
...
+  template:
+    metadata:
+      annotations:
+        qos.projectcalico.org/egressBandwidth: "100M"
+        qos.projectcalico.org/egressBurst: "200M"
+    spec:
+        (...)
+```
+
+## How to
```
tomastigera, line 56: perhaps here is a the time to say "calico allows you to configure these N things and we use annotations" and then explain each annotation in a subsection (as you do) later on?

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+### Configure packet rate limits
+
+To configure packet rate limits,  use the `qos.projectcalico.org/ingressPacketRate` annotation for ingress, or `qos.projectcalico.org/egressPacketRate` for egress. Packet rate values are in packets per second and constrained from a minimum of `10` to a maximum of `1T` (10<sup>12</sup>) packets per second.
+
+### Configure established connection limits
+
+To configure limits for the number of established connections,  use the `qos.projectcalico.org/ingressMaxConnections` annotation for ingress, or `qos.projectcalico.org/egressMaxConnections` for egress. The values are in absolute numbers of simultaneous established connections and constrained from a minimum of `1` to a maximum of `100G` (100 * 10<sup>9</sup>) connections.
```
tomastigera: In all these 3 sections explain the role of the knobs. What does it mean to limit established connections etc.

Specifically here using `maximum of 100G` is likely better to use `100.000.000.000` or `100 billion` or `100 * 10^9` as people do not tend to think about kilo-connections ;-)

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+### All available annotations
```
tomastigera, line 101: feels like a duplicate of what you are doing in the 3 sections above? What if each of the sections has a small paragraph explaining in a sentence or two what the knobs are for followed by list of knobs that belong to that section (subset of the table) ?

calico/networking/configuring/qos-controls.mdx
coutinhop: This is before the specific types of controls, so I wanted to only mention the quantity suffixes (but yes, for bandwidth it will be bits per second, for packet rate, packets per second and for connections an absolute number). I'll remove the mention to the package the parses and just explain the suffixes/syntax...

calico/networking/configuring/qos-controls.mdx
coutinhop: I just found this would be a way to have a more concrete example for the user, that hey can copy and adapt (I tend to function well that way so as a user, so I thought it could be good). Do you think I could keep it? I did expand the previous section to explain what the annotations are used for

calico/networking/configuring/qos-controls.mdx
coutinhop: I did that in the QoS concept, please let  me know how that looks!

calico/networking/configuring/qos-controls.mdx
coutinhop: Good point, I can see that these sections are getting too long, will work on that

calico/networking/configuring/qos-controls.mdx
ctauchen: I'm wondering whether it would be simplest to just use the correct unit. Could you do something like this? 

| Annotation                                  | Description                                                                                               |
|---------------------------------------------|-----------------------------------------------------------------------------------------------------------|
| `qos.projectcalico.org/ingressBandwidth`| Specifies the maximum incoming bandwidth allowed (in bits per second).<br/>Min value: 1 kilobit per second (1 kbps)<br/>Max value: 1 petabit per second (1 Pbps) |

ctauchen: That means you can read the table without any confusion, and you don't need a separate paragraph to explain things.

ctauchen: But if you say this is a widely accepted format, I won't press too hard. This just seems much easier to work with.

tomastigera: linux tools like iperf usually take 5G or 3k as the bit rate. It is a little weird to specify count as 3M, however I guess that connection limits will be rather small and user will use a precise number

calico/networking/configuring/qos-controls.mdx
coutinhop: @tomastigera sorry for the long delay, I finally did this. Please take a look!

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+$[prodname] implements three types of QoS Controls, which can be configured using annotations on the Kubernetes pods for either ingress (traffic coming into the pod), egress (traffic going out of the pod), or both:
+1. Bandwidth: limiting the bit rate of traffic into/out of the pod
+2. Packet rate: limiting the number of packets per second that can be sent or received by the pod
+3. Number of established connections: limiting the absolute number of connections that the pod may initiate or accept
+
+### Annotations
+
+On Kubernetes, $[prodname] uses annotations on pods for the configuration of QoS controls. You can annotate pods in any of these ways:
```
tomastigera:
```suggestion
On Kubernetes, $[prodname] uses [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) on pods for the configuration of QoS controls. You can annotate pods in any of these ways:
```

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,117 @@
...
+- pods can be directly annotated with `kubectl annotate`;
+- they can be created with the annotations in their metadata;
+- if they're part of a deployment or daemonset, they can have the annotations in the metadata of their template spec.
```
tomastigera: You cannot give an exhaustive list and I think it makes the docs a bit bloated. You already have the example below :+1: 

What I think you forget to mention (and has much bigger value) is that with calico, the annotation takes affect on existing pod while with the k8s plugin, it does not. I'd highlight it. WDYT?

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+The values of the QoS controls annotations may contain suffixes, such as `k`, `M`, `G`, `T`, `P`, or `E` to represent large numbers. Binary suffixes are also supported, for example `1Ki` represents "1024", and so on. For more details, see the [Quantity type of the resource package](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity).
```
tomastigera:
```suggestion
The values of the QoS controls annotations may contain suffixes, such as `k`, `M`, `G`, `T`, `P`, or `E` to represent large numbers. Binary suffixes are also supported, for example `1Ki` represents "1024".
```

I clicked on the link, it got me to a pretty complex section, confusing.

calico/networking/configuring/qos-controls.mdx
coutinhop: Thanks! Done.

calico/networking/configuring/qos-controls.mdx
coutinhop: Thanks! Done.

calico/networking/configuring/qos-controls.mdx
coutinhop: Agreed, I'll just remove the list then! As for mentioning the annotation taking effect, I think I'll do it in the bandwidth section, what do you think? Thanks for catching this btw!

calico/networking/configuring/qos-controls.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+If the [Kubernetes bandwidth plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/) annotations `kubernetes.io/ingress-bandwidth` or `kubernetes.io/egress-bandwidth` are configured and the `qos.projectcalico.org` annotations are not, they are honored and $[prodname] will configure bandwidth limits based on them . The $[prodname] specific annotations take precedence, so if both are configured, the Kubernetes annotation will be ignored. This makes upgrading from a $[prodname] version that uses the bandwidth plugin to a version with native bandwith limiting seamless, as the behavior will remain consistent across versions (with $[prodname] replacing the functionality previously provided by the Kubernetes bandwidth plugin).
+
+One essential difference between the bandwidth plugin and $[prodname]'s implementations is that with $[prodname], the bandwidth limits take effect immediately, whereas with the bandwidth plugin the pods need to be recreated for the limits to take effect.
```
coutinhop, line 76: @tomastigera PTAL

ctauchen:
LGTM with latest revisions. Thanks!

tomastigera (approved):
looks great, thanks!

ctauchen:
LGTM, thanks!

PR #1987, opened 2025-04-03
Update documentation for non-cluster host Typha scaling

hjiawei:
This change updates the non-cluster host installation documentation to support Typha scaling introduced in CE v3.21 EP2. It also clarifies the purpose of the NonClusterHost custom resource and endpoints defined in the spec.

Product Version(s):

Calico Enterprise v3.21 EP2.

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/getting-started/bare-metal/about.mdx
+++ b/calico-enterprise/getting-started/bare-metal/about.mdx
@@ -105,21 +105,13 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
 
 ### Set up your non-cluster host or VM
 
-1. Configure the Calico RHEL yum/dnf repositories
+1. Configure the Calico Enterprise repository.
 
-   - If your non-cluster host or VM has RHEL 8 installed, then run the following command:
+    ```bash
+    curl -sfL https://downloads.tigera.io/ee/rpms/v3.21/calico-enterprise.repo -o /etc/yum.repos.d/calico-enterprise.repo
+    ```
 
-      ```bash
-      curl https://downloads.tigera.io/ee/rpms/v3.21/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
-      ```
-
-   - If your non-cluster host or VM has RHEL 9 installed, then run the following command:
-
-      ```bash
-      curl https://downloads.tigera.io/ee/rpms/v3.21/calico_rhel9.repo -o /etc/yum.repos.d/calico_rhel9.repo
-      ```
-
-     Only Red Hat Enterprise Linux 8 and 9 x86-64 operating systems are supported in this version of $[prodname].
+    Only Red Hat Enterprise Linux 8 and 9 x86-64 operating systems are supported in this version of $[prodname].
 
 1. Install Calico node and fluent-bit log forwarder packages.
 
--- a/calico-enterprise/getting-started/bare-metal/about.mdx
+++ b/calico-enterprise/getting-started/bare-metal/about.mdx
@@ -40,7 +40,7 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
 
 ## How to
 
-### Set up your Kubernetes cluster to work with Accepted Values non-cluster host or VM
+### Set up your Kubernetes cluster to work with a non-cluster host or VM
 
 1. Create a `NonClusterHost` custom resource.
 
@@ -73,12 +73,6 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
 
 3. Create a `kubeconfig` file for your non-cluster host or VM similar to the following:
 
-    - Cluster:
-      - `server`: The URL of the Kubernetes API server.
-      - `certificate-authority-data`: Base64 encoded CA certificate to verify the Kubernetes API server's certificate.
-    - User:
-      - `token`: A bearer token associated with the `tigera-noncluster-host` service account obtained in the previous step.
-
     ```yaml
     apiVersion: v1
     kind: Config
@@ -86,8 +80,8 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
     preferences: {}
     clusters:
     - cluster:
-        certificate-authority-data: <ca-cert-to-verify-Kubernetes-api-server>
-        server: <URL-of-the-Kubernetes-api-server>
+        certificate-authority-data: <certificate-authority-data>
+        server: <server>
       name: noncluster-hosts
     contexts:
     - context:
@@ -97,8 +91,15 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
     users:
     - name: tigera-noncluster-host
       user:
-        token: <token-for-tigera-noncluster-host>
+        token: <token>
     ```
+
+    Replace the following:
+    - Cluster:
+      - `<server>`: The URL of the Kubernetes API server.
+      - `<certificate-authority-data>`: Base64 encoded CA certificate to verify the Kubernetes API server's certificate.
+    - User:
+      - `<token>`: A bearer token associated with the `tigera-noncluster-host` service account obtained in the previous step.
 
 4. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
 
--- a/calico-enterprise/getting-started/bare-metal/about.mdx
+++ b/calico-enterprise/getting-started/bare-metal/about.mdx
@@ -40,57 +40,71 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
 
 ## How to
 
-### Set up your Kubernetes cluster to work with a non-cluster host or VM
+### Set up your Kubernetes cluster to work with Accepted Valuesa non-cluster host or VM
 
-1. Apply the `NonClusterHost` custom resource. The `endpoint` field is required. It defines the remote endpoint for the non-cluster host log forwarder.
+1. Apply the `NonClusterHost` custom resource. This resource enables the cluster-side ingestion endpoint to receive logs from non-cluster hosts. Additionally, it provides a dedicated Typha deployment for managing communication between the non-cluster host agent and Typha.
 
-   ```bash
-   kubectl create -f - <<EOF
-   apiVersion: operator.tigera.io/v1
-   kind: NonClusterHost
-   metadata:
-     name: tigera-secure
-   spec:
-     endpoint: <endpoint for non-cluster host log forwarder>
-   EOF
-   ```
+    To ensure a secure connection, non-cluster host agents will establish an mTLS connection via the `typhaEndpoint` endpoint. As a result, the external ingress or load balancer must be configured to pass through TCP Layer 4 traffic.
 
-   Wait until the manager shows a status of Available, then proceed to the next step.
+    :::note
 
-1. Obtain the token for `tigera-noncluster-host` service account.
+    Starting from Calico Enterprise version v3.21.0 Early Preview 2, the `typhaEndpoint` field is required for non-cluster host agents to function correctly. Ensure this filed is properly configured to maintain seamless connectivity.
 
-   ```bash
-   kubectl get secret -n calico-system tigera-noncluster-host -o jsonpath='{.data.token}' | base64 --decode
-   ```
+    :::
 
-1. Use a text editor to create a kubeconfig file for non-cluster host or VM.
+    | Field         | Description                                                          | Accepted Values                                                          | Schema |
+    | ------------- | -------------------------------------------------------------------- | ------------------------------------------------------------------------ | ------ |
+    | endpoint      | Location of the log ingestion point for non-cluster hosts. Required. | Any HTTPS URL with a domain name and a port number                       | string |
+    | typhaEndpoint | Location of the Typha endpoint for non-cluster host agent and Typha communication. Required. | Any IP address or domain name with a port number | string |
 
-   ```yaml
-   apiVersion: v1
-   kind: Config
-   current-context: noncluster-hosts
-   preferences: {}
-   clusters:
-   - cluster:
-       certificate-authority-data: <your cluster certificate>
-       server: <your cluster server>
-     name: noncluster-hosts
-   contexts:
-   - context:
-       cluster: noncluster-hosts
-       user: tigera-noncluster-host
-     name: noncluster-hosts
-   users:
-   - name: tigera-noncluster-host
-     user:
-       token: <token from previous step>
-   ```
+    ```bash
+    kubectl create -f - <<EOF
+    apiVersion: operator.tigera.io/v1
+    kind: NonClusterHost
+    metadata:
+      name: tigera-secure
+    spec:
+      endpoint: https://ingress-or-load-balancer.example.com:443
+      typhaEndpoint: 5.6.7.8:5473
+    EOF
+    ```
 
-   Take the cluster information from an already existing kubeconfig.
+    Wait until the Tigera Manager and non-cluster Typha deployments reach the Available status before proceeding to the next step.
 
-1. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
+2. Obtain the token for `tigera-noncluster-host` service account.
 
-1. Follow [Configure access to the $[prodname] web console](../../operations/cnx/access-the-manager.mdx) page to configure the $[prodname] web console access.
+    ```bash
+    kubectl get secret -n calico-system tigera-noncluster-host -o jsonpath='{.data.token}' | base64 --decode
+    ```
+
+3. Use a text editor to create a kubeconfig file for non-cluster host or VM.
+
+    ```yaml
+    apiVersion: v1
+    kind: Config
+    current-context: noncluster-hosts
+    preferences: {}
+    clusters:
+    - cluster:
+        certificate-authority-data: <your cluster certificate>
+        server: <your cluster server>
+      name: noncluster-hosts
+    contexts:
+    - context:
+        cluster: noncluster-hosts
+        user: tigera-noncluster-host
+      name: noncluster-hosts
+    users:
+    - name: tigera-noncluster-host
+      user:
+        token: <token from previous step>
+    ```
+
+    Take the cluster information from the administrator kubeconfig.
+
+4. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
+
+5. Follow [Configure access to the $[prodname] web console](../../operations/cnx/access-the-manager.mdx) page to configure the $[prodname] web console access.
 
 ### Set up your non-cluster host or VM
 
@@ -98,15 +112,15 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
 
    - If your non-cluster host or VM has RHEL 8 installed, then run the following command:
 
-     ```bash
-     curl https://downloads.tigera.io/ee/rpms/v3.20/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
-     ```
+      ```bash
+      curl https://downloads.tigera.io/ee/rpms/v3.21/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
+      ```
 
    - If your non-cluster host or VM has RHEL 9 installed, then run the following command:
 
-     ```bash
-     curl https://downloads.tigera.io/ee/rpms/v3.20/calico_rhel9.repo -o /etc/yum.repos.d/calico_rhel9.repo
-     ```
+      ```bash
+      curl https://downloads.tigera.io/ee/rpms/v3.21/calico_rhel9.repo -o /etc/yum.repos.d/calico_rhel9.repo
+      ```
 
      Only Red Hat Enterprise Linux 8 and 9 x86-64 operating systems are supported in this version of $[prodname].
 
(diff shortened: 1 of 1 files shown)


calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
 
 1. Configure the Calico RHEL yum/dnf repositories
 
    - If your non-cluster host or VM has RHEL 8 installed, then run the following command:
 
-     ```bash
-     curl https://downloads.tigera.io/ee/rpms/v3.20/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
-     ```
+      ```bash
+      curl https://downloads.tigera.io/ee/rpms/v3.21/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
```
hjiawei: @danudey This hard-coded version is likely to be missed or forget in the future releases. Do we have a release process to update them? Also, it is more ideal to use a RHEL version-agnostic `calico.repo` file as we discussed before.

danudey: We should definitely replace this with a templated value asap, and I'll work on the version-agnostic calico.repo replacement soon.

@ctauchen Are you able to replace this  `v3.21` with a reference to whichever actual version this is for?

ctauchen: We could do that. Best would be if we changed the location to follow the pattern of all our other downloads: 

`filesUrl: 'https://downloads.tigera.io/ee/v3.21.0-1.0',`

So something like $[filesURL]/rpms/calico.repo --> https://downloads.tigera.io/ee/v3.21.0-1.0/rpms/calico.repo.

Keeping it specific to the patch version is cleaner.

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+    :::note
+
+    Starting from Calico Enterprise version v3.21.0 Early Preview 2, the `typhaEndpoint` field is required for non-cluster host agents to function correctly. Ensure this filed is properly configured to maintain seamless connectivity.
+
+    :::
+
+    | Field         | Description                                                          | Accepted Values                                                          | Schema |
+    | ------------- | -------------------------------------------------------------------- | ------------------------------------------------------------------------ | ------ |
+    | endpoint      | Location of the log ingestion point for non-cluster hosts. Required. | Any HTTPS URL with a domain name and a port number                       | string |
+    | typhaEndpoint | Location of the Typha endpoint for non-cluster host agent and Typha communication. Required. | Any IP address or domain name with a port number | string |
```
ctauchen:
```suggestion
    | typhaEndpoint | Required. Location of the Typha endpoint for non-cluster host agent and Typha communication. | Any IP address or domain name with a port number | string |
```

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
-     user:
-       token: <token from previous step>
-   ```
-
-   Take the cluster information from an already existing kubeconfig.
-
-1. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
-
-1. Follow [Configure access to the $[prodname] web console](../../operations/cnx/access-the-manager.mdx) page to configure the $[prodname] web console access.
+### Set up your Kubernetes cluster to work with Accepted Valuesa non-cluster host or VM
```
ctauchen: Not quite sure what your change is doing here. Can you clarify this?

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+
+    Wait until the Tigera Manager and non-cluster Typha deployments reach the Available status before proceeding to the next step.
+
+2. Obtain the token for `tigera-noncluster-host` service account.
+
+    ```bash
+    kubectl get secret -n calico-system tigera-noncluster-host -o jsonpath='{.data.token}' | base64 --decode
+    ```
+
+3. Use a text editor to create a kubeconfig file for non-cluster host or VM.
```
ctauchen:
```suggestion
3. Create a `kubeconfig` file for your non-cluster host or VM similar to the following:
```

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+
+    ```yaml
+    apiVersion: v1
+    kind: Config
+    current-context: noncluster-hosts
+    preferences: {}
+    clusters:
+    - cluster:
+        certificate-authority-data: <your cluster certificate>
```
ctauchen:
```suggestion
        certificate-authority-data: <cluster-certificate>
```

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+        server: <your cluster server>
```
ctauchen:
```suggestion
        server: <cluster-server-IP-address>
```
I'd specify exactly what this is, since 'server' can mean many things. Is this an IP address?

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+      name: noncluster-hosts
+    contexts:
+    - context:
+        cluster: noncluster-hosts
+        user: tigera-noncluster-host
+      name: noncluster-hosts
+    users:
+    - name: tigera-noncluster-host
+      user:
+        token: <token from previous step>
```
ctauchen:
```suggestion
        token: <token-for-tigera-noncluster-host>
```

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+    ```
+
+    Take the cluster information from the administrator kubeconfig.
```
ctauchen:
```suggestion
    Take the cluster information from the administrator `kubeconfig` file.
```
Questions I have after reading this:
What's an 'administrator kubeconfig file'?
What do you mean by 'cluster information'?

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+
+4. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
+
+5. Follow [Configure access to the $[prodname] web console](../../operations/cnx/access-the-manager.mdx) page to configure the $[prodname] web console access.
```
ctauchen: This isn't really part of your procedure. If you think it's important to guide users to this information, consider adding it in a 'Next steps' section.

calico-enterprise/getting-started/bare-metal/about.mdx
ctauchen: Probably a model like this, considering the package types and os versions also need to fit:

Download domain | product (os | ee | cc) | version ( v3.21.0-1.0 | v3.21.0-2.0 | v3.21.1 | etc) | package (rpm | deb ) | OS+version (rhel9 | rhel8 | etc) | calico.repo

ctauchen: So https://downloads.tigera.io/ee/v3.21.0-1.0/rpm/rhel9/calico.repo
https://downloads.tigera.io/ee/v3.21.0-1.0/rpm/rhel8/calico.repo

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
 1. Configure the Calico RHEL yum/dnf repositories
 
    - If your non-cluster host or VM has RHEL 8 installed, then run the following command:
 
-     ```bash
-     curl https://downloads.tigera.io/ee/rpms/v3.20/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
-     ```
+      ```bash
+      curl https://downloads.tigera.io/ee/rpms/v3.21/calico_rhel8.repo -o /etc/yum.repos.d/calico_rhel8.repo
+      ```
```
ctauchen: Thanks for tidying up the code block indentation!

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+    | typhaEndpoint | Location of the Typha endpoint for non-cluster host agent and Typha communication. Required. | Any IP address or domain name with a port number | string |
+
+    ```bash
+    kubectl create -f - <<EOF
+    apiVersion: operator.tigera.io/v1
+    kind: NonClusterHost
+    metadata:
+      name: tigera-secure
+    spec:
+      endpoint: https://ingress-or-load-balancer.example.com:443
```
ctauchen: Why change the value from `<endpoint for non-cluster host log forwarder>`?

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+      typhaEndpoint: 5.6.7.8:5473
```
ctauchen:
```suggestion
      typhaEndpoint: 5.6.7.8:5473
```
Do they need to specify this exact address? If not, we need to add a placeholder and explain how to find the right address.

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+
+1. Apply the `NonClusterHost` custom resource. This resource enables the cluster-side ingestion endpoint to receive logs from non-cluster hosts. Additionally, it provides a dedicated Typha deployment for managing communication between the non-cluster host agent and Typha.
+
+    To ensure a secure connection, non-cluster host agents will establish an mTLS connection via the `typhaEndpoint` endpoint. As a result, the external ingress or load balancer must be configured to pass through TCP Layer 4 traffic.
+
+    :::note
+
+    Starting from Calico Enterprise version v3.21.0 Early Preview 2, the `typhaEndpoint` field is required for non-cluster host agents to function correctly. Ensure this filed is properly configured to maintain seamless connectivity.
+
+    :::
```
ctauchen: >is required for non-cluster host agents to function correctly.

This qualifies the 'required' idea to make it sound like you may lose only some functionality if you don't specify this field. Is that what you mean? 

If not, better to leave it at just 'required'. We don't need to know why.

ctauchen: I think we can do without this note. Just make clear in this step that we're setting these two fields and that they're required.

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+
+    | Field         | Description                                                          | Accepted Values                                                          | Schema |
+    | ------------- | -------------------------------------------------------------------- | ------------------------------------------------------------------------ | ------ |
+    | endpoint      | Location of the log ingestion point for non-cluster hosts. Required. | Any HTTPS URL with a domain name and a port number                       | string |
```
ctauchen:
```suggestion
    | endpoint      | Required. Location of the log ingestion point for non-cluster hosts. | Any HTTPS URL with a domain name and a port number                       | string |
```

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,94 +40,108 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
-   Take the cluster information from an already existing kubeconfig.
-
-1. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
-
-1. Follow [Configure access to the $[prodname] web console](../../operations/cnx/access-the-manager.mdx) page to configure the $[prodname] web console access.
+### Set up your Kubernetes cluster to work with Accepted Valuesa non-cluster host or VM
+
+1. Apply the `NonClusterHost` custom resource. This resource enables the cluster-side ingestion endpoint to receive logs from non-cluster hosts. Additionally, it provides a dedicated Typha deployment for managing communication between the non-cluster host agent and Typha.
+
+    To ensure a secure connection, non-cluster host agents will establish an mTLS connection via the `typhaEndpoint` endpoint. As a result, the external ingress or load balancer must be configured to pass through TCP Layer 4 traffic.
```
ctauchen: If you think people need to know details about the connection, then consider moving that out of the step procedure to an introduction that explains what's going on. Otherwise, you can leave this out.

I'd go with something like this:

```
1. Create an apply a `NonClusterHost` custom resource.
   This resource enables the cluster-side ingestion endpoint to receive logs from non-cluster hosts.
    Additionally, it provides a dedicated Typha deployment for managing communication between the non-cluster host agent and Typha.
```

ctauchen (commented):
Thanks for this @hjiawei! Some comments here for you.

calico-enterprise/getting-started/bare-metal/about.mdx
danudey: Keeping it patch version specific is not preferred, as it means any upgrade will require changing the configuration file on every server before upgrading - not the standard practice for software.

The path should be simplified, which, as pointed out, we can do by combining the two files into one, but we still want to keep it organized by minor release rather than patch.

ctauchen: > not the standard practice for software.
Isn't that standard practice for our software? 

Perhaps there's something I'm missing. How would someone decide to downgrade from 3.19.4 to 3.19.3, for example? If we supply only minor versions, then I suppose we overwrite that RPM every time? It's not just downgrades. If someone wanted to reinstall, or install the same patch on a second cluster ...

hjiawei: I believe using this format `https://downloads.tigera.io/ee/rpms/calico.repo` (without specifying the CE version) is important to ensure a smooth upgrade to a newer CE release version. Once the initial setup is done on hosts, we should avoid requiring system administrators to manually SSH into the host and modify the repository location. 

For downgrade, a system administrator can run the `dnf downgrade` command. Every RPM package contains version information in their filename and metadata, e.g., `calico-node-3.20.1.rpm` or `calico-node-3.20.0~pre1.1^20240823gitc210c473`.

Here is [an example](https://repo.almalinux.org/almalinux/8/BaseOS/x86_64/os/Packages/) on how packages are typically organized. Note, only the major (8 or 9) RHEL version is in the URL, while RPM packages from different releases are saved within the same directory.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: I will make the change but almost all other places (for example, [felix configuration doc](https://docs.tigera.io/calico-enterprise/latest/reference/resources/felixconfig#felix-configuration-definition)) put "Required" at the end.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Reverted the change. It was an unintentional change.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Updated as suggested.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Updated as suggested.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: This is the URL of the Kubernetes API server. I will make that explicit.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Updated as suggested.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Rewrite to explain the placeholder fields in bullet form. "administrator kubeconfig file" is inaccurate and removed.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: These are really placeholders. Not sure what the best presentation form is in our docs. A customer needs to fill the `endpoint` and `typhaEndpoint` fields.

* For `endpoint`, it must be in the URL format: `https://some-domain-or-ip-addr:port`.
* For `typheEndpoint`, it must in this form: `some-domain-or-ip:port`.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Step 5 is removed.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Sure. This note is removed.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Updated as suggested. The details for `typhaEndpoint` endpoint is moved to the "Configure hosts to communicate with your Kubernetes cluster" section below. The layer 4 pass through requirement is a bit untypical to configure an ingress or LB so I want to mention that in our doc.

calico-enterprise/getting-started/bare-metal/about.mdx
ctauchen: Hmm. Okay, thanks. We should definitely move required to the front, because it's most likely to be seen there. I'll see if there's some way I can get those moved.

calico-enterprise/getting-started/bare-metal/about.mdx
ctauchen: I'm not sure I understand. If we don't version the link, how can we simultaneously provide different Calico Enterprise versions? 

As in, we have one user who wants to install 3.20, and another who wants to install 3.21. How do they both get what they're looking for?

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,109 +40,121 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
-     user:
-       token: <token from previous step>
-   ```
-
-   Take the cluster information from an already existing kubeconfig.
-
-1. Create a [`HostEndpoint` resource](../../reference/host-endpoints/overview.mdx) for each non-cluster host or VM. The `node` and `expectedIPs` fields are required to match the hostname and the expected interface IP addresses.
-
-1. Follow [Configure access to the $[prodname] web console](../../operations/cnx/access-the-manager.mdx) page to configure the $[prodname] web console access.
+### Set up your Kubernetes cluster to work with Accepted Values non-cluster host or VM
```
ctauchen: I don't think your revert worked correctly.

calico-enterprise/getting-started/bare-metal/about.mdx
```diff
@@ -40,109 +40,121 @@ To learn how to restrict traffic to/from hosts and VMs using Calico network poli
...
+    kubectl get secret -n calico-system tigera-noncluster-host -o jsonpath='{.data.token}' | base64 --decode
+    ```
+
+3. Create a `kubeconfig` file for your non-cluster host or VM similar to the following:
+
+    - Cluster:
+      - `server`: The URL of the Kubernetes API server.
+      - `certificate-authority-data`: Base64 encoded CA certificate to verify the Kubernetes API server's certificate.
+    - User:
+      - `token`: A bearer token associated with the `tigera-noncluster-host` service account obtained in the previous step.
```
ctauchen: The preferred format for explaining placeholders is as follows:

In the code block, format placeholders like this `<placeholder-text>`.
If you need additional explanation, add a list AFTER the code block:

```
Replace the following:
* `<placholder-1>`: The thing for your whatsit.
* `<placeholder-2>`: A list of ice cream flavors.
* `<placeholder-3>`: Something else.
```

Particularly if you have explanations, you can be much less verbose with the placeholder names.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: `dnf install calico-node-3.20.1-1`. You can specify RPM package versions during install. See dnf install [examples](https://dnf.readthedocs.io/en/latest/command_ref.html#install-examples).

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Move the explanations after the code block.

calico-enterprise/getting-started/bare-metal/about.mdx
hjiawei: Should be fix this time.

calico-enterprise/getting-started/bare-metal/about.mdx
ctauchen: Ah okay, thanks for the explanation. It looks then like we should have a generic file location for the .repo file, and then specify the version in the next step. 

If that's the case, perhaps we should be specifying the version in step 2? 

image

PR #1988, opened 2025-04-04
[EV-5315] Add enforced, pending policy and PolicyEvaluationMode desc

dimitri-nicolo:
Product Version(s):

Calico Cloud and Enterprise 3.21-1+

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


calico-cloud/reference/component-resources/node/felix/configuration.mdx
```diff
@@ -248,6 +248,7 @@ for 1022 endpoints on the host.
...
+| `FlowLogsPolicyEvaluationMode`          | `FELIX_FLOWLOGSPOLICYEVALUATIONMODE`          | `Continuous`                          | Defines how policies are evaluated and reflected in flow logs. OnNewConnection - In this mode, staged policies are only evaluated when new connections are made in the dataplane. Staged/active policy changes will not be reflected in the `pending_policies` field of flow logs for long lived connections. Continuous - Felix evaluates active flows on a regular basis to determine the rule traces in the flow logs. Any policy updates that impact a flow will be reflected in the pending_policies field, offering a near-real-time view of policy changes across flows.                    |
```
ctauchen:
```suggestion
| `FlowLogsPolicyEvaluationMode`          | `FELIX_FLOWLOGSPOLICYEVALUATIONMODE`          | `Continuous`                          | Defines how policies are evaluated and reflected in flow logs. `OnNewConnection`: In this mode, staged policies are only evaluated when new connections are made in the data plane. Staged/active policy changes will not be reflected in the `pending_policies` field of flow logs for long lived connections. `Continuous`: Felix evaluates active flows on a regular basis to determine the rule traces in the flow logs. Any policy updates that impact a flow will be reflected in the `pending_policies` field, offering a near-real-time view of policy changes across flows.                    |
```
Apply globally, please.

ctauchen (commented):
A few edits, otherwise LGTM. Thanks for this!

calico-cloud/reference/component-resources/node/felix/configuration.mdx
dimitri-nicolo: @ctauchen I've applied the changes globally, thank you.

Additionally, I've added the `flowLogsPolicyEvaluationMode` configuration description to the relevant `reference/resources/felixconfig.mdx` files

ctauchen:
@dimitri-nicolo Thanks. Last question about versioning. From the tickets I'm reading, it looks like this is for 3.21 ep2 and the CC equivalent. Is that right? 

Currently, you have this set up to be published in the 3.21 ep1 documentation too. If that's not what you intended, you can remove the changes from the versioned docs in both CE and CC.

dimitri-nicolo:
> @dimitri-nicolo Thanks. Last question about versioning. From the tickets I'm reading, it looks like this is for 3.21 ep2 and the CC equivalent. Is that right?
> 
> Currently, you have this set up to be published in the 3.21 ep1 documentation too. If that's not what you intended, you can remove the changes from the versioned docs in both CE and CC.

@ctauchen I recently picked the implementation changes into the `v3.21-1` branch as the enforced and pending policy sub-fields are going to be used in the free tier. That's the reason I included the docs change in those branches as well.

ctauchen:
Ok, I think we need to change this a bit to accommodate that.

I'm ready to merge the changes to what will be 3.21-2 and the CC equivalent (in the vNext directories). You should modify this PR to include changes only to those versions (we'll call this PR #1) .

Even if you merged changes to EP1 in the code, we won't see that released unless they do a special EP1 patch release, which is rare. I know there's a 3.21 EP1.1 in the mix, but I want to check that it's being released publicly before we commit docs. 
* If it's not public, then we'll just wait for the the EP2 docs.
* If it is going to be released publicly, we'll add this to the publication PR. I can manage this and get your review when the time comes.

Finally, we'll need to see the situation for the next CC release, 21.2.0, which is based on the (possibly internal) 3.21 ep1.1. I would assume that this release can pick up your changes, but it might be good to check. For this, you should submit another PR for the CC 21-1 version, and we'll merge this into the `free` branch, which I'm using to stage changes for the next CC branch.

Let me know if I've misunderstood something. It's managed to get complicated.

dimitri-nicolo:
> Ok, I think we need to change this a bit to accommodate that.
> 
> I'm ready to merge the changes to what will be 3.21-2 and the CC equivalent (in the vNext directories). You should modify this PR to include changes only to those versions (we'll call this PR #1) .
> 
> Even if you merged changes to EP1 in the code, we won't see that released unless they do a special EP1 patch release, which is rare. I know there's a 3.21 EP1.1 in the mix, but I want to check that it's being released publicly before we commit docs.
> 
> * If it's not public, then we'll just wait for the the EP2 docs.
> * If it is going to be released publicly, we'll add this to the publication PR. I can manage this and get your review when the time comes.
> 
> Finally, we'll need to see the situation for the next CC release, 21.2.0, which is based on the (possibly internal) 3.21 ep1.1. I would assume that this release can pick up your changes, but it might be good to check. For this, you should submit another PR for the CC 21-1 version, and we'll merge this into the `free` branch, which I'm using to stage changes for the next CC branch.
> 
> Let me know if I've misunderstood something. It's managed to get complicated.

Thank you @ctauchen, I've removed the versioned changes from this PR, and squashed the commits.

Here is the PR with the CC v3.21-1 changes only: https://github.com/tigera/docs/pull/1995

ctauchen:
Thanks, @dimitri-nicolo. LGTM.

PR #2004, opened 2025-04-16
[RS-2421] Add doc for configuring additional GatewayClasses

nelljerram:
FTR, it was eventually agreed for this PR to add doc about configuring additional GatewayClasses to the OSS docs, but not to Enterprise - to avoid any possible suggestion that customer-configured GatewayClasses would be supported.

There was an additional piece of extra doc talking about possible customizations via the GatewayAPI CR.  This is supported, and has therefore been retained for both OSS and Enterprise.


calico/networking/gateway-api.mdx
```diff
@@ -416,3 +418,60 @@ EOF
...
+
+To make use of these customization fields, use `kubectl edit gatewayapi default` to edit the YAML for the `GatewayAPI` resource, and add or modify the customization fields that you require.
+
+### Configure additional GatewayClasses
+
+You may have a need for a tweak or customization that our `GatewayAPI` resource does not yet support.  For example, you may want to create a Gateway with EnvoyProxy config different from that which the Tigera operator provisions, and where the change that you need is not supported by an existing customization field in the `GatewayAPI` resource.
+
+To do this:
+- Extract the YAML for the EnvoyProxy and GatewayClass objects that the Tigera operator provisions.
+- Change the EnvoyProxy YAML according to your needs, and create that modified EnvoyProxy object with a different name (than "envoy-proxy-config").
```
ctauchen:
```suggestion
- Change the EnvoyProxy YAML according to your needs, and rename the modified EnvoyProxy object (not `envoy-proxy-config.yaml`).
```
For clarity.

calico/networking/gateway-api.mdx
```diff
@@ -416,3 +418,60 @@ EOF
...
+- Change the GatewayClass YAML to reference the name of the modified EnvoyProxy object, and create that modified GatewayClass with a different name (than "tigera-gateway-class").
```
ctauchen: Change as above.

ctauchen: Also change "filename" throughout to ``filename``.

ctauchen (commented):
A few comments, otherwise LGTM. 

I presume this needs to be copied elsewhere?

calico/networking/gateway-api.mdx
nelljerram: I'm afraid this isn't a correct suggestion, because it really is creating a new object, not renaming an existing one.

calico/networking/gateway-api.mdx
nelljerram: First point - as in my previous response.

Second point - actually these are resource names, not file names.  But OK, I think your suggestion still makes sense.

electricjesus (approved):

ctauchen:
@nelljerram Considering Ash's comments, let's remove the changes from CE and CC and merge the OSS changes.

PR #2032, opened 2025-04-29
Add page for OSS flow logs and Whisker

ctauchen:
Product Version(s):

Issue:

Link to docs preview:

https://deploy-preview-2032--calico-docs-preview-next.netlify.app/calico/next/observability/view-flow-logs

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


ctauchen:
@mazdakn @caseydavenport PTAL

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+
+You can observe network traffic in your cluster by viewing flow logs in the Calico Whisker web console.
+
+## About flow logs in Calico Open Source
+
+Tracking network activity in your cluster is essential for understanding how your applications communicate.
+Calico helps you see your network activity by processing your network flows and recording them in a digestible format called a flow log.
+
+A flow log isn't a record of every single network connection that takes place in your cluster.
+Rather, it is an aggregated summary of a type of network flow that occurs in your cluster over a period of time.
```
caseydavenport: I think defining a flow log in terms of network flows is a bit confusing, since a network flow isn't a self-evident primitive itself, so it ends up feeling a bit self-referential. 

I'd probably start with something like this:

> A flow log is an aggregation of connection data that matches the same set of criteria (i.e., source, destination, policies) across your cluster over a period of time. 

That in itself is still a bit reductive, as the full set of identifying information for a flow is the source fields, destination fields, the network protocol, the set of policy traversed by the connections, the action taken by policy, and who reported this (the source or destination end).

The "period of time" is also variable in general, although it is fixed within the context of Whisker to 15s intervals.

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+Imagine, for example, that your cluster has 1000 individual network flows that take place within 1 minute.
```
caseydavenport:
```suggestion
Imagine, for example, that your cluster has 1000 individual network connections that take place within 1 minute.
```

Same thing here - I think it's confusing to define a flow log in terms of flows, because we haven't defined what a flow is.

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+But the corresponding flow logs for that minute might turn those 1000 entries into 15 flow logs, each representing a distinct type of flow (Pod A to Pod B, public internet to pod C), but without the duplicate flow type entries.
+This way, you can actually understand the overall network activity without being overwhelmed by the sheer volume of data.
+
+### Using flow logs
+
+Flow logs are useful for a variety of purposes, including:
+* **Network visibility.**
+  Understanding traffic patterns within the cluster, between namespaces, and to or from external services.
+* **Troubleshooting connectivity issues.**
+  If pods aren't communicating, flow logs can show if traffic is being denied by policy, if it's reaching the destination
```
caseydavenport: Unfinished sentence here?

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+* **Security.**
+  Detecting policy violations (unexpected deny logs), identifying anomalous traffic patterns (e.g., unusually high data transfer), and investigating potential security incidents.
+* **Policy validation.**
+  Confirming that network policies are working as intended (allowing expected traffic, denying forbidden traffic).
+  This is especially useful for testing out [staged network policies](../network-policy/staged-network-policies.mdx) before moving them to production.
+
+### Calico Whisker
+
+The Calico Whisker web console is a browser-based interface that shows you a stream of flow logs from your cluster in real time.
```
caseydavenport, line 40: "real time" is up for debate - the flows show up in the console after some delay as we need to aggregate data from multiple sources before displaying it. Maybe "near real time" ?

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+| `end_time`                        | date              | End time of log collection in UNIX timestamp format.                                                                                                                                                                                                                                                                                                 |
+| `action`                          | keyword           | - `allow`: Calico accepted the flow.<br />- `deny`: Calico denied the flow.                                                                                                                                                                                                                                                              |
+| `bytes_in`                        | long              | Number of incoming bytes since the last export.                                                                                                                                                                                                                                                                                                      |
```
caseydavenport:
```suggestion
| `bytes_in`                        | long              | Number of incoming bytes within the timeframe.                                                                                                                                                                                                                                                                                                   |
```

It might be equivalent? But this is more accurate - i.e., what is "since the last export" for the first flow log you receive?

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+| `bytes_out`                       | long              | Number of outgoing bytes since the last export.                                                                                                                                                                                                                                                                                                      |
+| `dest_name`                       | keyword           | Contains one of the following values:<br />- Name of the destination pod.<br />- Name of the pod that was aggregated or the endpoint is not a pod. Check <code>dest_name_aggr</code> for more information, such as the name of the pod if it was aggregated.                                                                                         |
```
caseydavenport: There's no `dest_name_aggr` in OSS Flows - `dest_name` is effectively `dest_name_aggr` semantically, though. 

Note that throughout here, we use "Pod", but in actuality this means a collection of pods that share a name prefix (e.g., a Deployment).

For a concrete example, `dest_name` might be `my-service-abcd-*` and the pods within that deployment would have names like `my-service-abcd-1` and `my-service-abcd-2`

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+| `dest_port`                       | long              | Destination port. Not applicable for ICMP packets.                                                                                                                                                                                                                                                                                                   |
+| `dest_labels`                     | array of keywords | Labels applied to the destination pod. A hyphen indicates aggregation.                                                                                                                                                                                                                                                                               |
```
caseydavenport: I'm not sure that Calico OSS ever emits a hyphen for this field

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+| `reporter`                        | keyword           | - `src`: flow came from the pod that initiated the connection.<br />- `dst`: flow came from the pod that received the initial connection.                                                                                                                                                                                                            |
```
caseydavenport:
```suggestion
| `reporter`                        | keyword           | - `src`: flow came from the source that initiated the connection.<br />- `dst`: flow came from the destination that received the initial connection.                                                                                                                                                                                                            |
```

Not all sources / destinations are pods. (e.g., host endpoints can appear in flow data)

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+| `protocol`                        | keyword           | Protocol.                                                                                                                                                                                                                                                                                                                                            |
+| `policies`                        | array of arrays   | List of policies that interacted with this flow. See [Format of the policies field](#format-of-the-policies-field).                                                                                                                                                                                                                                                |
```
caseydavenport: guessing `format-of-the-policies-field` is a WIP?

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+   ```bash
+   kubectl port-forward -n calico-system service/whisker 8081:8081
+   ```
+
+   ```bash title="Expected output"
+   Forwarding from 127.0.0.1:8081 -> 8081
+   Forwarding from [::1]:8081 -> 8081
+   ```
+1. To open Calico Whisker, open your browser and go to `localhost:8081`.
+   You won't see any flows at the beginning.
```
caseydavenport:
```suggestion
   You may not see any flows at the beginning.
```

I think in most cases, users _will_ see flows at the beginning (we made a change to backfill with some historical data on load)

calico/observability/view-flow-logs.mdx
```diff
@@ -3,4 +3,98 @@ description: View flow logs in the Calico Whisker web console.
...
+| `source_namespace`                | keyword           | Namespace of the source endpoint. A `-` means the endpoint is not namespaced.                                                                                                                                                                                                                                                                        |
+| `source_labels`                   | array of keywords | Labels applied to the source pod. A hyphen indicates aggregation.                                                                                                                                                                                                                                                                                    |
+
+## View flow logs in the web console
+
+You can quickly view flow logs by port forwarding the Calico Whisker web console to your workstation.
+For security reasons, Whisker does not automatically expose an endpoint.
```
caseydavenport: I think it's fine to say "By default" instead of "For security reasons" - it makes it sound like there is something intrinsically insecure about Whisker, which isn't true - just that we can't guess the right way to export this Service outside of the cluster, and operationalizing a UI with authentication is typically a bespoke process per-user.

calico/observability/view-flow-logs.mdx
ctauchen: Thanks, I've added that line and rewritten some of the other sections. In particular, I'm avoiding 'flow' on its own, preferring 'connection'.

calico/observability/view-flow-logs.mdx
ctauchen: Changed.

calico/observability/view-flow-logs.mdx
ctauchen: Thanks, fixed.

calico/observability/view-flow-logs.mdx
ctauchen: I'd say it's true to the spirit of 'real time', even if it's pulling in data every 15 seconds. You could argue the same no matter what the interval, even if it was 15 ms.

But if you think this will mislead or confuse readers, let me know. It seems okay to me.

calico/observability/view-flow-logs.mdx
ctauchen: Ok. I've modified the line. This and all the other data types are lifted from the CE docs. It's good to adjust where necessary.

ctauchen: Changed all four instances.

calico/observability/view-flow-logs.mdx
ctauchen: Removed.

calico/observability/view-flow-logs.mdx
ctauchen: Holdover from CE, now removed.

calico/observability/view-flow-logs.mdx
ctauchen: Changed.

calico/observability/view-flow-logs.mdx
ctauchen: Changed.

calico/observability/view-flow-logs.mdx
ctauchen: Changed.

calico/observability/view-flow-logs.mdx
ctauchen: Thanks for the clarification. I removed the `dest_name_aggr` line.

I think for now I'll stick with the current 'pod' usage. But I'll make a note to clarify this here and in the other flow logs documentation in CE and CC.

ctauchen (commented):
@caseydavenport Thanks for the review. I've taken on nearly all of your suggestions. I'll merge this now and follow up separately if you have any strong disagreements.

Thanks!

PR #2102, opened 2025-06-12
[RS-2577] Mention new GatewayAPI customization fields

nelljerram:


ctauchen:
LGTM. Do these need to be moved also to OSS and CC?

nelljerram:
Thanks @ctauchen .  At time of writing this was intended for Enterprise and Cloud only, but the requirements position has changed and the changes should now be for OSS as well.  So I will copy the changes across to OSS and CC.

nelljerram:
@ctauchen I've just remembered that I meant to tag @apardeshi-tigera also to take a look at these docs.  So let's hold off on merging until he has done that.

calico/networking/gateway-api.mdx
```diff
@@ -429,56 +428,14 @@ The `GatewayAPI` resource has fields that allow some aspects of Gateway deployme
...
+- `spec.gatewayDeployment.service.metadata.annotations` allows control over annotations to place on the Service that is provisioned for each Gateway; these can be used, for example, to configure the cloud-specific type and properties of the associated external load balancer.
+- `spec.gatewayDeployment.service.spec.*loadbalancer*` allows control over the corresponding `*loadbalancer*` fields in the Service that is provisioned for each Gateway; in some clouds these can also be used to configure the type and properties of the associated external load balancer.
+- `spec.gatewayClasses` allows the provisioning of multiple GatewayClasses, each with its own set of customizations, instead of the single GatewayClass that the Tigera operator provisions by default.
 
 For full details please see [the `GatewayAPI` reference documentation](../reference/installation/api.mdx#operator.tigera.io/v1.GatewayAPI).
 
 To make use of these customization fields, use `kubectl edit gatewayapi default` to edit the YAML for the `GatewayAPI` resource, and add or modify the customization fields that you require.
 
-### Configure additional GatewayClasses
```
ctauchen, line 437: @nelljerram Why are we taking this out? This section is already live in the 3.30 docs.

nelljerram: Because we now have an official way to provision additional GatewayClasses, using the `spec.gatewayClasses` field.

ctauchen: Of course -- let me check where we are with Ash and we'll get this sorted out.

electricjesus (approved):

calico/networking/gateway-api.mdx
nelljerram: Many thanks @ctauchen .

ctauchen:
@nelljerram I got the thumbs-up from Ash. If you can squash and sort your merge conflict, we can get this merged today.

nelljerram:
@ctauchen Conflict is done.  Is squashing essential?  Personally I dislike it because it destroys the history of the conversation.  But will if you insist.  (Also not a fan of squash-merging because that leaves a branch in my local repo that appears to be unmerged - but I feel I'm losing the battle on that one.)

apardeshi-tigera:
@nelljerram @ctauchen 
For the 'Customization' part can we update to adopt following structure:

Section title: Gateway Customization:
Overview: 
    - Explaining that all gateway customization is done through a single GatewayAPI resource 
    - what fields are controlled : pods, load balancers, controller, multiple gateway classes
Example for common customization scenarios/use cases: Include YAML for each and kubectl for applying and verifying.
     - Handle high traffic with multiple gateway pods and adequate resources. (replicas, memory)
     - Set up external load balancers with cloud-specific features 
     - multiple gateway classes

nelljerram:
@apardeshi-tigera The convention for section title is that it's a sentence beginning with an imperative verb, so I think we should stick to that.

For your "Overview", I think the only point that is not already covered is that it's done through a **single** GatewayAPI resource.  If that's right, I'll propose a change to incorporate that.

For the Examples - yes, great idea to add those; I'll propose that shortly.

nelljerram:
@apardeshi-tigera I've updated the PR per your comments now.  Can you take another look at the **_calico-docs-preview-next_** preview and LMK what you think?

apardeshi-tigera:
Looks good

PR #2126, opened 2025-06-24
DOCS-2624: Add K8s compatibility for old CC versions

ctauchen:
This adds historical k8s compatibility data for active CC versions.

DOCS-2624


ctauchen:
CC @sujeet-kr @tmjd

PR #2183, opened 2025-07-17
Rename our API pages to Client library pages

rene-dekker:
Rename our API pages to Client library pages

Tigera API and Calico API are a source of confusion to some users. By renaming the page and adding slightly more context, the user should have more context around the purpose and use cases for this page.


ctauchen:
LGTM, thanks!

PR #2205, opened 2025-07-31
[RS-2690] Doc OpenShift restriction of Gateway API CRDs

nelljerram:


ctauchen:
@nelljerram Thanks for this. A few questions: 
* Are these CRDs unrestricted for earlier versions? Like 4.18?
* Do we have an idea of what happens for upcoming versions, like 4.20?

nelljerram:
@ctauchen 
- Yes
- No

I have tried to avoid creating a maintenance burden here - i.e. to write something that has a chance of remaining true when future versions come out.  But perhaps that can be done in a better way; please LMK your thoughts.

calico-cloud/networking/gateway-api.mdx
```diff
@@ -31,6 +31,12 @@ The Gateway API is an official Kubernetes API for advanced routing to services i
 | TLSRoute         | v1alpha2          |
 | UDPRoute         | v1alpha2          |
 
+:::note
+
+Some versions of OpenShift (currently just 4.19) restrict the Gateway API CRDs that may be installed: only `GatewayClass`, `Gateway`, `GRPCRoute`, `HTTPRoute` and `ReferenceGrant`.  In clusters with those OpenShift versions the `BackendLBPolicy`, `BackendTLSPolicy`, `TCPRoute`, `TLSRoute` and `UDPRoute` APIs will not be available.
```
ctauchen: How about something like this?

```
:::note
OpenShift 4.19 introduced restrictions for which the Gateway API CRDs can be installed. This change means that certain APIs are unavailable, a departure from OpenShift 4.18 and earlier, which provided full access to all CRDs.

This table shows which Gateway API CRDs are available for each OpenShift version.

| Gateway API CRD | OpenShift 4.18 and lower | OpenShift 4.19 |
| :--- | :--- | :--- |
| `GatewayClass` | ✅ Available | ✅ Available |
| `Gateway` | ✅ Available | ✅ Available |
| `GRPCRoute` | ✅ Available | ✅ Available |
| `HTTPRoute` | ✅ Available | ✅ Available |
| `ReferenceGrant` | ✅ Available | ✅ Available |
| `BackendLBPolicy` | ✅ Available | ❌ Not available |
| `BackendTLSPolicy` | ✅ Available | ❌ Not available |
| `TCPRoute` | ✅ Available | ❌ Not available |
| `TLSRoute` | ✅ Available | ❌ Not available |
| `UDPRoute` | ✅ Available | ❌ Not available |
```

ctauchen: If there's an OpenShift docs page that lists these, best would be to point there. I haven't found one yet.

nelljerram: [This](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html-single/updating_clusters/index#nw-ingress-gateway-api-manage-succession_updating-cluster-prepare) is on this topic - WDYT?

nelljerram: > How about something like this?

For me the table is a bit overkill - I feel it will make the page as a whole heavier and give undeserved prominence to OpenShift's choices.  But I see a little merit in listing the resources separately from the narrative.  Perhaps a good compromise would be a bulleted form like:
- Available in OpenShift 4.19: `GatewayClass`, `Gateway`, `GRPCRoute`, `HTTPRoute` and `ReferenceGrant`.
- Not available in OpenShift 4.19: `BackendLBPolicy`, `BackendTLSPolicy`, `TCPRoute`, `TLSRoute` and `UDPRoute`.

As regards the other wording changes in the narrative, I don't feel that any of them are big wins - but happy to defer to you if you feel strongly that they are.

ctauchen: Fair enough, let's try something like that. 

I suggest we link to this OS Gateway API Overview: https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html-single/ingress_and_load_balancing/index#nw-ingress-gateway-api-overview_ingress-gateway-api

It lists the resources that are available, and will likely be updated to include expanded functionality in future versions.

@nelljerram

nelljerram: Thanks @ctauchen , I've pushed an updating commit.  Please take another look when you have time.

ctauchen:
LGTM, merging. Thanks for this!

PR #2227, opened 2025-08-15
New crd reference generation for Operator API

tmjd:
Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


calico-enterprise/reference/installation/_api.md
```diff
@@ -0,0 +1,5991 @@
+Packages
+- [operator.tigera.io/v1](#operatortigeraiov1)
+
+
+{/* value off */}
```
ctauchen:
```suggestion
{/* vale off */}
```
@tmjd

calico-enterprise/reference/installation/api.mdx
```diff
@@ -4,7 +4,7 @@ description: Installation API reference
 
 # Installation reference
 
-import API from '@site/calico-enterprise/reference/installation/_api.mdx';
+import API from '@site/calico/reference/installation/_api.md';
```
ctauchen: This should be importing from @site/calico-enterprise/, right?

calico-enterprise/reference/installation/_api.md
ctauchen: This syntax is for MDX, which may be thrown when you have this as a plain MD file. Might just rename MDX for that and for general consistency.

calico/reference/installation/_crd-ref-docs/config.yaml
```diff
@@ -0,0 +1,69 @@
+processor:
```
ctauchen, line 1: So it looks like you generate the API docs based on the Installation CRDs, presumably for a specific version of the Operator, which serves both OSS and CE. Here, you exclude CE-only items from the list. 

Have I got that right?

ctauchen: Is there a way to automate this? I'm concerned about maintenance in the long term.

tmjd: Yes that is right, it is doing the same thing the previous generation did. It builds from a specific version of the operator code and filters out CE-only items. This new mechanism to generate the reference documents has the same mechanism as the previous mechanism had where there was a list of resource types that should be omitted.

There is a way we could tag resources in the operator to designate them enterprise only and have those resources omitted for non-CE generation. We'd just need to make those changes in the operator code before we could do that.

ctauchen:
LGTM

PR #2256, opened 2025-09-03
DOCS-2696: Add procedure for Maglev load balancing

ctauchen:
This commit create a new docs page for enabling Maglev-style
load balancing for Kubernetes services.

DOCS-2696

image

Product Version(s):

* Calico 3.31
* CE 3.23.0-1.0?
* CC 23
* 
Issue:

https://tigera.atlassian.net/browse/DOCS-2696

Link to docs preview:

https://deploy-preview-2256--calico-docs-preview-next.netlify.app/calico/next/networking/configuring/add-maglev-load-balancing

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico/networking/configuring/add-maglev-load-balancing.mdx
+++ b/calico/networking/configuring/add-maglev-load-balancing.mdx
@@ -1,0 +1,41 @@
+---
+description: Add Maglev load balancing to a Kubernetes service.
+---
+
+# Add Maglev load balancing to a service
+
+This guide shows you how to enable Maglev-style load balancing to a service.
+Maglev's consistent-hashing load balancing algorithm minimizes the amount of dropped packets and improves performance.
+This is especially useful for maintaining sessions during a load balancer failover.
+
+## Prerequisites
+
+* Your cluster uses the eBPF data plane with [direct server return mode](https://docs.tigera.io/calico/latest/operations/ebpf/enabling-ebpf#try-out-direct-server-return-mode).
+* All your nodes are running on Linux.
+* You have the name and namespace of the service you want to use Maglev-style load balancing.
+
+## Enable consistent-hash load balancing for a service
+
+To enable consistent-hash load balancing for a service, annotate the service by running the following command:
+
+```bash
+kubectl -n <namespace> annotate service <service> 'lb.projectcalico.org/external-traffic-strategy: "maglev"'
+```
+
+Replace the following:
+* `<namespace>`: The namespace of the service.
+* `<service>`: The name of the service.
+
+To verify the configuration, check the logs for any `calico-node` pod for the string `Wrote Maglev` in reference to your service.
+
+```bash title='Example log query'
+kubectl -n calico-system logs calico-node-7kltn | grep --ignore-case 'Wrote Maglev'
+```
+
+```bash title='Example output'
+2025-10-28 15:30:53.091 [INFO][75] felix/syncer.go 893: Wrote Maglev service backends to NGINX service=default/svc:maglev-port
+```
+
+## Additional resources
+
+* [Enable the eBPF data plane](../../operations/ebpf/enabling-ebpf.mdx)
--- a/sidebars-calico.js
+++ b/sidebars-calico.js
@@ -272,6 +272,7 @@ module.exports = {
             'networking/configuring/pod-mac-address',
             'networking/configuring/node-local-dns-cache',
             'networking/configuring/qos-controls',
+            'networking/configuring/add-maglev-load-balancing',
           ],
         },
         {


calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+* You've enabled [direct server return mode](https://docs.tigera.io/calico/latest/operations/ebpf/enabling-ebpf#try-out-direct-server-return-mode).
+* You have the name and namespace of a Service of type `LoadBalancer`.
+* You have `kubectl` access to your cluster.
+
+## Enable consistent-hash load balancing for a service
+
+To enable consistent-hash load balancing for a service, annotate the service by running the following command:
+
+```bash
+kubectl -n <namespace> annotate service <service> <placeholder-annotation>
```
ctauchen: @aaaaaaaalex What's the annotation?

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+This guide shows you how to enable Maglev-style load balancing to a service.
+Maglev's consistent-hashing load balancing algorithm minimizes the amount of dropped packets and improves perfomance.
+This is especially useful for maintaining sessions during a load balancer failover.
+
+## Prerequisites
+
+* Your Kubernetes cluster uses $[prodname] for networking.
+* All your nodes are running on Linux
+* $[prodname] is configured to use the eBPF data plane.
+* You've enabled [direct server return mode](https://docs.tigera.io/calico/latest/operations/ebpf/enabling-ebpf#try-out-direct-server-return-mode).
```
ctauchen: Based on what I'm reading, I think this should be required. It seems that degraded performance, particularly with UDP, makes it not worth implementing without DSR. 

In a very literal sense, you can implement this LB without DSR. Which is why it's referred to as a recommendation in some places.
But the purpose of this feature isn't 'implementing Maglev'; it's about adding the performance improvments that Maglev helps you to achieve. The non-DSR use case works with the 'implementing Maglev' use case, but not necessarily with the 'full performance improvement' use case.

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
+---
+description: Add Maglev load balancing to a Kubernetes service.
+---
+
+# Add Maglev load balancing to a service
```
ctauchen, line 5: PTAL @aaaaaaaalex . In particular, I'll need guidance on the actual annotation and thoughts on making DSR a hard requirement.

Regarding terminology: I think the horse has bolted on Maglev. If it means anything, it means the consistent hashing algorithm, which is the one bit you've added here. There are a lot of other important parts. But because the one part you've implemented is the core function, I think it's okay to use the term as a generic stand-in. If you were using another part, it would not be so easy.

In this draft I've used the term liberally but specified exactly what we mean by that. I hope that strikes the right balance.

aaaaaaaalex:
CC @tomastigera @fasaxc

calico/networking/configuring/add-maglev-load-balancing.mdx
aaaaaaaalex: Annotation TBD - I'll chat to the team tomorrow and we'll come up with something.
All good re: "Maglev" use, thanks!

calico/networking/configuring/add-maglev-load-balancing.mdx
aaaaaaaalex: That makes sense, happy to have it as a prerequisite here

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+This guide shows you how to enable Maglev-style load balancing to a service.
```
aaaaaaaalex, line 7:
```suggestion
This guide shows you how to enable Maglev-style load balancing for a service.
```

calico/networking/configuring/add-maglev-load-balancing.mdx
aaaaaaaalex, line 5:
```suggestion
# Enable Maglev load balancing for a service
```

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
+---
+description: Add Maglev load balancing to a Kubernetes service.
```
aaaaaaaalex, line 2:
```suggestion
description: Enable Maglev load balancing for a Kubernetes service.
```

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+---
+
+# Add Maglev load balancing to a service
+
+This guide shows you how to enable Maglev-style load balancing to a service.
+Maglev's consistent-hashing load balancing algorithm minimizes the amount of dropped packets and improves perfomance.
```
aaaaaaaalex: We're not sure if our implementation has any performance improvement in the area of throughput, so probably best if we only talk about the session-availability angle.

Also, suggest explicitly naming "load-balancer failover" as the scenario where dropped packets (or more precisely, dropped connections)  are minimized

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+This is especially useful for maintaining sessions during a load balancer failover.
+
+## Prerequisites
+
+* Your Kubernetes cluster uses $[prodname] for networking.
+* All your nodes are running on Linux
+* $[prodname] is configured to use the eBPF data plane.
+* You've enabled [direct server return mode](https://docs.tigera.io/calico/latest/operations/ebpf/enabling-ebpf#try-out-direct-server-return-mode).
+* You have the name and namespace of a Service of type `LoadBalancer`.
```
aaaaaaaalex: Type `LoadBalancer`, or alternatively, a regular (ClusterIP) service, where we are advertising the IP with BGP outside of the cluster.

NodePorts don't gain any benefit from Maglev, though. The service IP being pinged must be the same across every load-balancer node.

aaaaaaaalex (commented):
Thanks a mill @ctauchen , failing a good way to confirm the feature is working, I think this gets the crucial info across. Presumably, if users want the gorey details about Maglev, we don't necessarily have to (or should) be the ones to document them.

Left some optional suggestions, and some elaboration in other sections

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+# Add Maglev load balancing to a service
+
+This guide shows you how to enable Maglev-style load balancing to a service.
+Maglev's consistent-hashing load balancing algorithm minimizes the amount of dropped packets and improves perfomance.
+This is especially useful for maintaining sessions during a load balancer failover.
+
+## Prerequisites
+
+* Your Kubernetes cluster uses $[prodname] for networking.
```
aaaaaaaalex: Just thought of something else - while I'm conscious of falling down the slippery slope of over-explaining Maglev, I wonder if it's worth noting here that "loadbalancing" nodes and nodes that host the backing pods should be separate. I'm undecided on it. It might become apparent to the user pretty quickly without us telling them. Maybe it goes without saying that we are mitigating loadbalancer downtime, and not pod downtime

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+* All your nodes are running on Linux
+* $[prodname] is configured to use the eBPF data plane.
```
tomastigera: The next line implies this one. If you want to, say you need to use calico in ebpf mobe and make a reference to how to enable ebpf

calico/networking/configuring/add-maglev-load-balancing.mdx
tomastigera: What does this exactly say? You need calico ebpf, period. Afaict you may use non-calicoy CNI

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
...
+* You've enabled [direct server return mode](https://docs.tigera.io/calico/latest/operations/ebpf/enabling-ebpf#try-out-direct-server-return-mode).
+* You have the name and namespace of a Service of type `LoadBalancer`.
+* You have `kubectl` access to your cluster.
```
tomastigera: also sounds like a given

calico/networking/configuring/add-maglev-load-balancing.mdx
```diff
@@ -0,0 +1,34 @@
+---
+description: Add Maglev load balancing to a Kubernetes service.
+---
+
+# Add Maglev load balancing to a service
+
+This guide shows you how to enable Maglev-style load balancing to a service.
+Maglev's consistent-hashing load balancing algorithm minimizes the amount of dropped packets and improves perfomance.
+This is especially useful for maintaining sessions during a load balancer failover.
```
tomastigera, line 9: This isn't expecially useful, it is the point! This intro should say that maglev loadbalancing reduces disruptions during LBs outage by failing over to other LBs as all LBs are making consisten backend selection without mutual synchronization. It is only used balancing ingress from outside the cluster. All nodes can serve as the LBs. etc.

calico/networking/configuring/add-maglev-load-balancing.mdx
tomastigera: > I wonder if it's worth noting here that "loadbalancing" nodes and nodes that host the backing pods should be separate.

There is no need for them to be separate.

tomastigera (commented):
We should not over explain trivialities like you need to have a k8s cluster to start with. We need to explain what is maglev (can use reference), that is requires calico ebpf (reference, hyperlinks are great, why to wait until "other material, further reading) and how to enable it for a service - highlight the fact that it is per service feature!) etc.

take of your eng hat and put yourself in user's shoes

calico/networking/configuring/add-maglev-load-balancing.mdx
ctauchen: > Afaict you may use non-calicoy CNI

Can you give an example of this? 

I thought we needed this prerequisite to exclude Calico-for-policy-only scenarios, such as using AWS VPC CNI with Calico for policy.

calico/networking/configuring/add-maglev-load-balancing.mdx
ctauchen: Ok, so what works here? Are we saying that we support all services, regardless of type? But only some types benefit from Maglev?

calico/networking/configuring/add-maglev-load-balancing.mdx
ctauchen: You're right, of course. The idea here is to clarify the state of things right here, without requiring people (with Windows nodes) to ferret out that information on another page.

calico/networking/configuring/add-maglev-load-balancing.mdx
ctauchen: Thanks -- I'll get that in with the rewrite.

ctauchen:
> We should not over explain trivialities like you need to have a k8s cluster to start with. We need to explain what is maglev (can use reference), that is requires calico ebpf (reference, hyperlinks are great, why to wait until "other material, further reading) and how to enable it for a service - highlight the fact that it is per service feature!) etc.
> 
> take of your eng hat and put yourself in user's shoes

Yes, thanks. Will see about references, and I'll add a line to clarify that this needs to be done for each service you want it for.

calico/networking/configuring/add-maglev-load-balancing.mdx
ctauchen: @tomastigera What do you think here?

calico/networking/configuring/add-maglev-load-balancing.mdx
aaaaaaaalex: > There is no need for them to be separate.

If theyre co-located, we don't get any benefit from maglev :thinking: churning a LB will kill backing pods

calico/networking/configuring/add-maglev-load-balancing.mdx
tomastigera: The issue with spelling out everything here is that when things change we need to change it in many places. I honestly think itis enough to say, that this feature is only supported with calico ebpf. Those who will get here either know what they are looking for or they can pause, follow a link to "calico ebpf"

calico/networking/configuring/add-maglev-load-balancing.mdx
tomastigera: > > Afaict you may use non-calicoy CNI
> 
> Can you give an example of this?
> 
> I thought we needed this prerequisite to exclude Calico-for-policy-only scenarios, such as using AWS VPC CNI with Calico for policy.

With eBPF we never use calico "just" for policy. We also do the networking and kube-proxy replacement and service implementation. The only bit we do not do is what CNI/IPAM do, allocating IPs and programming routes. I am pretty sure we can do maglev in this case as we can do nodeports in this case as well. We will test it.

tomastigera: > > There is no need for them to be separate.
> 
> If theyre co-located, we don't get any benefit from maglev 🤔 churning a LB will kill backing pods

yes, some backends will die. I think this could be a good "best practice - do not schedule pods on those nodes because ... " but it is equally important to say that all nodes can act like the LBs. It is up to the user to set their network up for their usecase.

calico/networking/configuring/add-maglev-load-balancing.mdx
tomastigera: Any service that is exposed outside of the cluster. As long as it has IP that is routable from outside would work. Type LoadBalancer is special because it may be implemented by cloud, usually something external to k8s. I think calico can now implement it too in some way https://docs.tigera.io/calico/latest/networking/ipam/service-loadbalancer

So LoadBalancer is likely not the best type to use. NodePort would not provide any benefit because it has IP of a node and so it cannot fail over to another node. Either of these types, however, also have a ServiceIP like any other service and that IP is virtual and can be used for maglev.

All this said, the best practice is likely to setup a basic ClusterIP type of a service https://kubernetes.io/docs/concepts/services-networking/service/#type-clusterip 

The docs need to mention that you need to setup your network such that if the LB fails, network will reroute packets to a different node, say using ECMP (link). If the routing does not change, maglev on its own would not help to fix the issue.

tomastigera:
> > take of your eng hat and put yourself in user's shoes

eee thought it is @aaaaaaaalex 's PR hence this comment :sweat_smile:

calico/networking/configuring/add-maglev-load-balancing.mdx
aaaaaaaalex: > > > There is no need for them to be separate.
> > 
> > 
> > If theyre co-located, we don't get any benefit from maglev 🤔 churning a LB will kill backing pods
> 
> yes, some backends will die. I think this could be a good "best practice - do not schedule pods on those nodes because ... " but it is equally important to say that all nodes can act like the LBs. It is up to the user to set their network up for their usecase.

Ah yeah I get you, sounds good

ctauchen:
Closing this here, to be picked up again in 3.32 cycle.

ctauchen:
@aaaaaaaalex Will you take another look to see that this all still reflects what's gone in?

ctauchen:
Closing, content now in #2365

PR #2282, opened 2025-09-16
Add note about openstack slaac subnets

kristiangronas:
Product Version(s): 3.31, 3.32

Issue:

Link to docs preview:

https://deploy-preview-2282--calico-docs-preview-next.netlify.app/calico/next/networking/openstack/ipv6

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico/networking/openstack/ipv6.mdx
+++ b/calico/networking/openstack/ipv6.mdx
@@ -47,10 +47,17 @@ OpenStack (not $[prodname]) controls whether a VM gets IPv4, IPv6, or both addre
     - `sed -i '/PATH/i\new_ip6_prefixlen=128' /sbin/dhclient-script`
     - `systemctl restart network`
 
-1.  Configure IPv6 support in $[prodname] by defining an IPv6 subnet in each Neutron network with:
+1.  Configure IPv6 support in $[prodname] by defining IPv6 subnet(s) with either one or both types of addressing:
 
-    - The IPv6 address range that you want your VMs to use
-    - DHCP enabled
-    - IPv6 address mode set to DHCPv6 stateful
+    DHCPv6 address provisioning subnet:
+      - The IPv6 address range that you want your VMs to use
+      - DHCP enabled
+      - IPv6 address mode set to DHCPv6 stateful
+
+    SLAAC address provisioning subnet: (avoids DHCPv6 leases, but requires a /64 range)
+      - The /64 IPv6 address range that you want your VMs to use
+      - DHCP enabled
+      - IPv6 address mode set to use SLAAC
+      - IPv6 router advertisement mode set to SLAAC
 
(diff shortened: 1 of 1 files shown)


ctauchen:
Hi @kristiangronas, thanks for the submission! It would be helpful if you could provide some context here. Is there a relate issue or PR from projectcalico/calico?

nelljerram:
@ctauchen Yes, this goes with https://github.com/projectcalico/calico/pull/10876

calico/networking/openstack/ipv6.mdx
```diff
@@ -53,4 +53,7 @@ OpenStack (not $[prodname]) controls whether a VM gets IPv4, IPv6, or both addre
     - DHCP enabled
     - IPv6 address mode set to DHCPv6 stateful
 
+1.  Calico now also supports `slaac` as both the address mode and router advertisment mode, either in addition to a subnet with dhcpv6 stateful or instead of.
+    SLAAC addresses are more predictable (both sides use the mac address to generate the ip), but note that the subnet needs an entire /64 CIDR, unlike DHCPv6 stateful
```
nelljerram: I think we should present this instead as an alternative within the preceding bullet.  Something like this...

1.  Configure IPv6 support in $[prodname] by defining an IPv6 subnet in each Neutron network.  For DHCP provisioning, configure the subnet with:

    - The IPv6 address range that you want your VMs to use
    - DHCP enabled
    - IPv6 address mode set to DHCPv6 stateful

    For SLAAC address provisioning, configure the subnet with:

    - The /64 IPv6 address range that you want your VMs to use
    - DHCP enabled
    - IPv6 address mode set to SLAAC
    - IPv6 router advertisement mode set to SLAAC

    We suggest that you initially configure both IPv4 and IPv6 subnets in each network. This allows handling VM images that support only IPv4 alongside those that support both IPv4 and IPv6, and allows a VM to be accessed over IPv4 in case this is needed to troubleshoot any issues with its IPv6 configuration. In principle, though, we are not aware of any problems with configuring and using IPv6-only networks in OpenStack.

nelljerram: @kristiangronas WDYT?

ctauchen: I agree with @nelljerram.  The original text might fit well as release note material in the code PR.

kristiangronas: Agreed, i wasn't sure how to add it earlier, does it look better now?

nelljerram (approved):
Many thanks @kristiangronas

nelljerram:
@ctauchen Do you mind approving the workflows for this?

ctauchen:
LGTM, thanks!

PR #2308, opened 2025-10-08
DOCS-2766: Fix methods of accessing CE web console

ctauchen:
DOCS-2766

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/operations/cnx/access-the-manager.mdx
+++ b/calico-enterprise/operations/cnx/access-the-manager.mdx
@@ -84,7 +84,7 @@ spec:
 
 ### Log in to the $[prodname] web console
 
-Access the $[prodname] web console in your browser at: `https://localhost:9443`
+Access the $[prodname] web console in your browser using the URL for your ingress controller. For example: `https://<ingress-url>`.
 
 </TabItem>
 <TabItem label="Load balancer service" value="Load balancer service-1">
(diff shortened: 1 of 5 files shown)


calico-enterprise/operations/cnx/access-the-manager.mdx
```diff
@@ -84,7 +84,7 @@ spec:
 
 ### Log in to the $[prodname] web console
 
-Access the $[prodname] web console in your browser at: `https://localhost:9443`
+Access the $[prodname] web console in your browser using the URL for your ingress controller. For example: `https://<ingress-url>`.
```
ctauchen, line 87: @Josh-L For each of these, it would be helpful to provide guidance on how to obtain these URLs. Can you help with this?

Josh-L: @ctauchen yes definitely. I'm working through these scenarios for the first time so it's a bit of a learning curve for me as well. Once I'm a bit more confident in the details and have things working I can fill you in.

ctauchen: @Josh-L Following up on this. Did you get any further with these scenarios?

Josh-L: I did yes, but unfortunately I don't think the instructions are generic enough that we should be putting them in our docs. Ingress resources and loadbalancers are going to be specific to how each customer configures them so I think leaving it generic to "<ingress-url>" might be our best bet, with similar text for the loadbalancers.

I think the most important part though is that the text below the tabbed access methods needs to either be generic so that it can apply to any of the tabs above it, or it needs to also be tabbed so that it will appear correct for each tabbed method.

ctauchen: Fair enough. I'll leave it without details. Thanks again for raising this.

PR #2510, opened 2026-02-04
DOCS-2833 Add script that updates Felix config for OSS

ctauchen:
This script updates the FelixConfiguration component with data the OSS and CE code repositories. A new makefile target makes it easy to run with `make update-felix-config`.

DOCS-2833

Product Version(s):

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/scripts/update-felix-config.sh
+++ b/scripts/update-felix-config.sh
@@ -1,0 +1,157 @@
+#!/usr/bin/env bash
+
+###############################################################################
+#                                                                             #
+#   felix-update-config-ce.sh                                                 #
+#   Created: 2026-02-05                                                       #
+#                                                                             #
+#   Description: Synchronizes Felix configuration JSON files from GitHub      #
+#   repos (OSS & Enterprise) into local documentation paths.                  #
+#                                                                             #
+###############################################################################
+
+# --- Safety & Environment ---
+set -euo pipefail
+
+# --- Global Constants (No Magic Numbers) ---
+readonly SUCCESS=0
+readonly E_MISSING_TOKEN=1
+readonly E_INVALID_JSON=2
+readonly VERSIONS_FILE_OSS="calico_versions.json"
+readonly VERSIONS_FILE_CE="calico-enterprise_versions.json"
+
+# --- Global Variables ---
+VERSIONS_OSS=($(jq -r '.[]' "$VERSIONS_FILE_OSS"))
+VERSIONS_CE=($(jq -r '.[]' "$VERSIONS_FILE_CE"))
+CLEANUP_FILES=()
+
+# --------------------------------------------------------------------------- #
+# cleanup()                                                                   #
+# Internal function triggered on script exit to remove all temp files.        #
+# --------------------------------------------------------------------------- #
+cleanup() {
+    local file
+    for file in "${CLEANUP_FILES[@]:-}"; do
+        if [[ -f "$file" ]]; then
+            rm -f "$file"
+        fi
+    done
+}
+
+trap cleanup EXIT
+
+# --------------------------------------------------------------------------- #
+# update_felix_config()                                                       #
+# Downloads a remote JSON file and replaces a local target file.              #
+# --------------------------------------------------------------------------- #
+update_felix_config() {
+    local tmpfile
+    tmpfile=$(mktemp -t remote-url-output.XXXXXX)
+    CLEANUP_FILES+=("$tmpfile")
+
+    printf "%s\n\n" "Begin processing $VERSION"
+    printf "%s\n" "Target path: $LOCAL_PATH"
+    printf "%s\n" "Source URL: $REMOTE_URL"
+
+    curl -fsSL \
+          -H "Authorization: token $GITHUB_TOKEN" \
+          -H "Accept: application/vnd.github.v3.raw" \
+          -o "$tmpfile" \
+          "$REMOTE_URL"
+
+    if jq -e . "$tmpfile" >/dev/null 2>&1; then
+        echo "Success: $tmpfile is valid JSON."
+    else
+        echo "Error: $tmpfile contains invalid JSON." >&2
+        exit "$E_INVALID_JSON"
+    fi
+
+    mv "$tmpfile" "$LOCAL_PATH"
+    printf "%s\n\n\n" "Finished processing $VERSION."
+}
+
+# --------------------------------------------------------------------------- #
+# OSS Functions                                                               #
+# --------------------------------------------------------------------------- #
+
+update_master_OSS () {
+    VERSION="master"
+    LOCAL_PATH="calico/_includes/components/FelixConfig/config-params.json"
+    REMOTE_URL="https://raw.githubusercontent.com/projectcalico/calico/refs/heads/master/felix/docs/config-params.json"
+    update_felix_config
+}
+
+update_versions_OSS () {
+    local -n VERSIONS_ARRAY=$1
+    local i
+    for i in "${!VERSIONS_ARRAY[@]}"; do
+        VERSION="${VERSIONS_ARRAY[$i]}"
+        LOCAL_PATH="calico_versioned_docs/version-${VERSION}/_includes/components/FelixConfig/config-params.json"
+        REMOTE_URL="https://raw.githubusercontent.com/projectcalico/calico/refs/heads/release-v${VERSION}/felix/docs/config-params.json"
+
+        if [[ "$VERSION" == "3.29" ]]; then
+            printf "%s\n" "Skipping $VERSION: No action required."
+            continue
+        fi
+
+        update_felix_config
+    done
+}
+
+# --------------------------------------------------------------------------- #
+# Enterprise Functions                                                        #
+# --------------------------------------------------------------------------- #
+
+update_master_CE () {
+    VERSION="master"
+    LOCAL_PATH="calico-enterprise/_includes/components/FelixConfig/config-params.json"
+    REMOTE_URL="https://api.github.com/repos/tigera/calico-private/contents/felix/docs/config-params.json?ref=master"
+    update_felix_config
+}
+
+update_versions_CE() {
+    local -n VERSIONS_ARRAY=$1
+    local i branch base_version
+
+    for i in "${!VERSIONS_ARRAY[@]}"; do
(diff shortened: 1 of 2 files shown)


lwr20:
How does this script fit into your process?  Do you run it regularly?
Or run it just before every release?

Presumably the intent is that you run it and use the changes it makes to raise a PR?

The script itself looks fine.

ctauchen:
> How does this script fit into your process? Do you run it regularly? Or run it just before every release?
> 
> Presumably the intent is that you run it and use the changes it makes to raise a PR?
> 
> The script itself looks fine.

For now, the idea is just to have this available to run manually for quick and accurate updates. So yes, run and raise a PR. I'd probably aim to run it at release time.

Next steps are to include other products and fold into proper automation. But that was beyond the scope for the moment.

scripts/felix-config-update.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+set -e/-u/pipefail
```
danudey: I think copilot meant that you should add at least one of the "-e/-u/pipefail" options. The actual syntax would be `set -eu -o pipefail`

scripts/felix-config-update.sh
```diff
@@ -0,0 +1,65 @@
...
+ curl -o "$TMPFILE" "$REMOTE_URL"
+
+ if jq -e . "$TMPFILE" >/dev/null 2>&1; then
+     echo "Success: $TMPFILE is valid JSON."
+ else
+     echo "Error: $TMPFILE contains invalid JSON."
+     exit 1
+ fi
+
+ curl -o "$LOCAL_PATH" "$REMOTE_URL"
```
danudey: I always add `-fsSL` to my curl invocations.

* `-f` causes it to return an error code if it gets an HTTP error (so the script ends)
* `-s` causes it not to print the (usually pointless) progress indicator (silent)
* `-S` shows (prints out) the error if one happened
* `-L` follows HTTP redirects instead of saving the redirect response

Long form is `--fail --silent --show-error --location`

IMHO all of these options should always be used whenever using curl in a shell script, as it catches a lot of errors that would otherwise be missed.

scripts/felix-config-update.sh
```diff
@@ -0,0 +1,65 @@
...
+update_versions() {
+  # This function updates the configurations from the versioned branches.
+  for i in "${!VERSIONS[@]}"; do
+      VERSION="${VERSIONS[$i]}"
+      LOCAL_PATH="calico_versioned_docs/version-${VERSION}/_includes/components/FelixConfig/config-params.json"
+      REMOTE_URL="https://raw.githubusercontent.com/projectcalico/calico/refs/heads/release-v${VERSION}/felix/docs/config-params.json"
+
+      # Exclude version 3.29, which came before direct-from-source method.
+      if [ "$VERSION" == "3.29" ]; then
+          printf "%s\n" "Skipping $VERSION: No action required."
```
danudey: You can use `echo "blah blah"` instead of `printf "%s\n" "blah blah"` unless you're doing more complex variable printing (like trying to print values as hex or formatting a number in a specific way). Tends to be more readable.

danudey (changes requested):

scripts/felix-config-update.sh
ctauchen: That's great, thanks. Added.

scripts/felix-config-update.sh
ctauchen: Yeah, I realized that after when i couldn't run the script! That's changed now.

scripts/felix-config-update.sh
ctauchen: I got to that after having trouble with newlines. I think I'll leave it for now, but I take the point that it's cleaner with a plain echo statement.

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
+
```
danudey, line 2: Running `shellcheck` shows warnings for the two arrays (reading from json via `jq`) since, _potentially_, if the input (the version strings) had spaces they would split incorrectly. Since that's almost definitely not going to happen then, for the sake of pleasing the linter, I'd recommend ignoring these results.

```suggestion
# shellcheck disable=SC2207

```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+# --- Global Variables ---
+VERSIONS_OSS=($(jq -r '.[]' "$VERSIONS_FILE_OSS"))
+VERSIONS_CE=($(jq -r '.[]' "$VERSIONS_FILE_CE"))
+CLEANUP_FILES=()
+
+# --------------------------------------------------------------------------- #
+# cleanup()                                                                   #
+# Internal function triggered on script exit to remove all temp files.        #
+# --------------------------------------------------------------------------- #
+cleanup() {
```
danudey, line 33: Another linter complain: `shellcheck` thinks `cleanup()` is unreachable because it doesn't understand traps. Adding a comment here silences these warnings and gives us a clean linter result. Minor, but good to be clean!

```suggestion
# shellcheck disable=SC2317
cleanup() {
```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+# Downloads a remote JSON file and replaces a local target file.              #
+# --------------------------------------------------------------------------- #
+update_felix_config() {
+    local tmpfile
+    tmpfile=$(mktemp -t remote-url-output.XXXXXX)
+    CLEANUP_FILES+=("$tmpfile")
+
+    printf "%s\n\n" "Begin processing $VERSION"
+    printf "%s\n" "Target path: $LOCAL_PATH"
+    printf "%s\n" "Source URL: $REMOTE_URL"
```
danudey: Might be cleaner to replace `printf` here. This also lets us see in the script how the lines will be laid out in the output (with an obvious blank line).

```suggestion
    echo "Begin processing ${VERSION}"
    echo
    echo "Target path: ${LOCAL_PATH}"
    echo "Source URL: ${REMOTE_URL}"
```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+
+    if jq -e . "$tmpfile" >/dev/null 2>&1; then
+        echo "Success: $tmpfile is valid JSON."
+    else
+        echo "Error: $tmpfile contains invalid JSON." >&2
+        exit "$E_INVALID_JSON"
+    fi
+
+    mv "$tmpfile" "$LOCAL_PATH"
+    printf "%s\n\n\n" "Finished processing $VERSION."
```
danudey: `echo -e` will interpret escape sequences like newlines, letting us clean this up a bit.

```suggestion
    echo -e "Finished processing ${VERSION}.\n\n"
```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+update_versions_OSS () {
+    local -n VERSIONS_ARRAY=$1
+    local i
+    for i in "${!VERSIONS_ARRAY[@]}"; do
+        VERSION="${VERSIONS_ARRAY[$i]}"
+        LOCAL_PATH="calico_versioned_docs/version-${VERSION}/_includes/components/FelixConfig/config-params.json"
+        REMOTE_URL="https://raw.githubusercontent.com/projectcalico/calico/refs/heads/release-v${VERSION}/felix/docs/config-params.json"
+
+        if [[ "$VERSION" == "3.29" ]]; then
+            printf "%s\n" "Skipping $VERSION: No action required."
```
danudey:
```suggestion
            echo "Skipping ${VERSION}: No action required."
```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+            branch="release-calient-v${base_version}-1"
+        else
+            branch="release-v${VERSION}"
+        fi
+
+        LOCAL_PATH="calico-enterprise_versioned_docs/version-${VERSION}/_includes/components/FelixConfig/config-params.json"
+        REMOTE_URL="https://api.github.com/repos/tigera/calico-private/contents/felix/docs/config-params.json?ref=${branch}"
+
+        if [[ "$VERSION" == "3.20-2" ]]; then
+            printf "%s\n" "Skipping $VERSION: No action required."
```
danudey:
```suggestion
            echo "Skipping ${VERSION}: No action required."
```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+
+# --- Main Logic ---
+
+if [[ -z "${GITHUB_TOKEN:-}" ]]; then
+    echo "Error: GITHUB_TOKEN is not set." >&2
+    exit "$E_MISSING_TOKEN"
+fi
+
+printf "OSS Versions: %s\n" "${VERSIONS_OSS[*]}"
+printf "CE Versions:  %s\n\n" "${VERSIONS_CE[*]}"
```
danudey:
```suggestion
echo "OSS Versions: ${VERSIONS_OSS[*]}"
echo "CE Versions: ${VERSIONS_CE[*]}"
echo
```

scripts/update-felix-config.sh
```diff
@@ -0,0 +1,157 @@
...
+
+update_master_OSS
+update_versions_OSS VERSIONS_OSS
+
+update_master_CE
+update_versions_CE VERSIONS_CE
+
+printf "%s\n" "All updates are complete."
```
danudey:
```suggestion
echo "All updates are complete."
```

danudey (approved):
LGTM, but some suggestions that IMHO will make the script a bit easier to read/maintain in the long run.

scripts/update-felix-config.sh
ctauchen: That's what I needed! I got into this because I was having trouble with echo and newlines. Thanks!

scripts/update-felix-config.sh
ctauchen: Added, thanks!

scripts/update-felix-config.sh
ctauchen: done, thanks

PR #2557, opened 2026-03-03
Add separate CRD chart to enterprise Helm install docs

caseydavenport:
The `tigera-operator` chart no longer bundles CRDs (see https://github.com/projectcalico/calico/pull/11727) — they're now in a separate `crd.projectcalico.org.v1` chart. The open source Calico install docs already have this, but the enterprise install docs were still missing it in a few places.

This adds the CRD chart download + install step to:
- Standard enterprise Helm install
- Management cluster Helm install (both NodePort and LoadBalancer)
- Managed cluster Helm install

Also fixes the existing CRD install command in the standard enterprise doc (was missing `.tgz` extension and had an unnecessary `--version` flag for a local file).

Fixes https://tigera.atlassian.net/browse/CORE-12418


ctauchen (commented):
Do the copilot comments merit attention? Otherwise LGTM. @caseydavenport

PR #2612, opened 2026-03-27
Added docs for KubeVirt

song-jiang:
Product Version(s):

Calico v3.32

Calico Enterprise v3.23

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed


skoryk-oleksandr:
@song-jiang  LGTM :+1: . 
I was wondering if we should also include a security considerations section? I noticed the IPAM Enhancement spec has a security checklist and I'm not sure if it's meant to be user-facing documentation or just internal.

Some things I was thinking about:

  - The fact that CNI verifies VM pod identity via ownerReferences rather than pod labels - is that worth mentioning so users understand the trust model?
  - RBAC guidance - since a user who can create pods and read VMI resources could potentially forge ownerReferences to claim a VM's IP, should we recommend restricting pod creation permissions?
  - Should we recommend admission controllers (Kyverno, OPA/Gatekeeper) for production to prevent ownerReference spoofing?
  - Maybe a note about Calico component permissions being read-only on KubeVirt resources?

calico/_includes/components/FelixConfig/config-params.json
```diff
@@ -1215,18 +1215,18 @@
           "Required": false,
           "OnParseFailure": "ReplaceWithDefault",
           "AllowedConfigSources": "All",
-          "Description": "Disables WireGuard metrics collection, which the Prometheus client does by default, when\nset to false. This reduces the number of metrics reported, reducing Prometheus load.",
-          "DescriptionHTML": "<p>Disables WireGuard metrics collection, which the Prometheus client does by default, when\nset to false. This reduces the number of metrics reported, reducing Prometheus load.</p>",
+          "Description": "Disables wireguard metrics collection, which the Prometheus client does by default, when\nset to false. This reduces the number of metrics reported, reducing Prometheus load.",
+          "DescriptionHTML": "<p>Disables wireguard metrics collection, which the Prometheus client does by default, when\nset to false. This reduces the number of metrics reported, reducing Prometheus load.</p>",
```
nelljerram, line 1219: I think "WireGuard" was better.  That's what the official [website](https://www.wireguard.com/) says.

Also I don't think this kind of change is an enjoyable use of our time, so I'd also generally question whatever is driving this.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+## Big picture
+
+$[prodname] provides networking for KubeVirt virtual machines (VMs) running on your Kubernetes cluster,
+including persistent IP addresses across VM lifecycle events and support for live migration.
+
+## Value
+
+KubeVirt runs VMs inside Kubernetes pods. When a VM reboots, is evicted, or live-migrates to
+another host, the underlying pod is destroyed and recreated. Without IP persistence, each new pod
+receives a fresh IP address, which breaks existing connections and changes the VM's network identity.
```
nelljerram: s/receives/would receive/ to make it clear that this doesn't actually happen.

nelljerram: So "would receive a fresh IP address, which would break existing connections and change the VM's network identity"

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+$[prodname]'s KubeVirt support ensures that:
+
+- VMs retain the same IP address across reboots, pod evictions, and live migrations.
```
nelljerram: s/VMs retain/A VM retains/

Unnecessary plurals usually add confusion.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+### Supported networking mode: bridge
+
+$[prodname] supports KubeVirt live migration using the **bridge** binding mode. In bridge mode,
+the VM is connected to the pod network through a Linux bridge, and the VM uses the same IP address
+that $[prodname] assigns to the pod. This is required because:
+
+- **IP address persistence** depends on the VM IP matching the pod IP. Bridge mode ensures
+  the VM sees and uses the pod IP directly.
+- **Network policy** is applied on the pod's veth interface. Bridge mode preserves this
+  relationship so policy enforcement works correctly for VM traffic.
```
nelljerram: I don't think this can be correctly stated, because network policy also works in OpenStack, and OpenStack doesn't use bridge mode.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+- **Live migration** relies on detecting gratuitous ARP (GARP) packets from the VM on the
+  pod's veth interface to know when the VM has activated on the target host.
```
nelljerram: This one also.  This works in OpenStack too, and OpenStack doesn't use bridge mode.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+Other KubeVirt networking modes (such as masquerade) are not supported for live migration
+because the VM would use a different IP than the pod IP, breaking IP persistence and
+policy enforcement.
+
+### BGP networking required
+
+Live migration currently requires BGP networking. Overlay networking (VXLAN, IP-in-IP)
+support is planned for a future release.
```
nelljerram, line 46: Do we still support BGP networking with IP-in-IP?  I _think_ we still have an option for that, although I'm also keen for it to disappear as soon as possible!  If so, the first sentence needs to say "Live migration currently requires BGP networking with no overlay".

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+### Policy setup timeout
+
+During live migration, KubeVirt needs to know when the destination host is ready for the VM.
+The `policy_setup_timeout_seconds` CNI configuration parameter interlocks the progress of the
+live migration with policy programming. The CNI plugin delays reporting success until network
+policy is in place on the destination, or until the timeout expires.
+
+### Limitations
+
+- **WireGuard is not supported with live migration.
```
nelljerram: Markup issue here?  It looks wrong to have two stars before WireGuard but not after.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+## Before you begin
+
+- A working Kubernetes cluster with KubeVirt installed.
+- $[prodname] installed with BGP networking.
+- Access to `projectcalico.org/v3` resources, either by installing the $[prodname]
+  [API server](../../operations/install-apiserver.mdx) or by using
+  [calicoctl](../../operations/calicoctl/install.mdx).
```
nelljerram, line 82: Add blank line here?  (Just for readability of the source.)

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+## How to
+
+### Enable live migration on VMs with bridge binding
```
nelljerram: Another variant term.  Can we just say "bridge mode" consistently everywhere?

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+bridge binding with an error like:
+`cannot migrate VMI which does not use masquerade to connect to the pod network`.
+
+:::note
+
+This annotation tells KubeVirt that the CNI plugin ($[prodname]) handles IP address
+preservation and network continuity during migration. $[prodname] ensures the target pod
+receives the same IP and that routes are updated to steer traffic to the new host.
+
+:::
```
nelljerram: I think this note is just repeating things already said.  Better to delete here.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+    {
+      "type": "calico",
+      "policy_setup_timeout_seconds": 10,
+      ...
+    }
+  ]
+}
+```
+
+### Configure live migration route priority
```
nelljerram: I think it's better to delete this section, as the defaults are fine and most users will never have a reason to change them.  In rare cases where changes are needed, it's sufficient that these fields are covered in the FelixConfig and BGPConfig reference sections.

calico/networking/kubevirt/kubevirt-networking.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+For production deployments, consider using admission controllers such as
+[Kyverno](https://kyverno.io/) or [OPA/Gatekeeper](https://open-policy-agent.github.io/gatekeeper/)
+to enforce that only KubeVirt controllers can set `ownerReferences` pointing to
+`VirtualMachineInstance` resources on pods.
+
+### $[prodname] component permissions
+
+$[prodname] components (Felix, confd, the CNI plugin) require only **read** access to KubeVirt
+resources (`VirtualMachineInstance`). They do not create, modify, or delete any KubeVirt resources.
```
nelljerram: Either add "etc." here, or add the specific other resources that we read.  Certainly VirtualMachineInstanceMigration.  Any others?

calico/networking/kubevirt/live-migration-bgp.mdx
```diff
@@ -0,0 +1,229 @@
+---
+description: Configure BGP routing to support KubeVirt live migration across racks and AS boundaries.
+---
+
+# BGP routing for KubeVirt live migration
+
+## Big picture
+
+When KubeVirt VMs live-migrate between hosts, $[prodname] uses route priority to steer traffic to
```
nelljerram:
```suggestion
When a KubeVirt VM live-migrates between hosts, $[prodname] uses route priority to steer traffic to
```

calico/networking/kubevirt/live-migration-bgp.mdx
```diff
@@ -0,0 +1,229 @@
...
+  configure `BGPFilter` resources to encode priority information using BGP communities or other
+  attributes that cross AS boundaries.
+
+### The downward default model is not supported
+
+The [downward default model](../../reference/architecture/design/l3-interconnect-fabric.mdx#the-downward-default-model),
+where ToR switches only advertise a default route downward to compute nodes, is **not compatible
+with live migration**. Live migration requires specific /32 routes for the migrating VM to be
+propagated across racks so that all nodes can route traffic to the new host. A default-route-only
+topology cannot carry per-VM route priority information.
```
nelljerram: This isn't quite right.  It's still fine to use downward default for almost all routes, but the ToR must pass on _elevated_ priority routes in their own right.  I think installers would see that as still _mostly_ using downward default.

To reflect that:
- perhaps "The pure downward default model is not supported" in the title?
- adapt the text to incorporate what I've written in this comment.

calico/networking/kubevirt/live-migration-bgp.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+### Route aggregation
+
+Under normal conditions, $[prodname] aggregates individual /32 workload routes into larger CIDR
+blocks for BGP advertisement, reducing the number of routes in the network. During live migration,
+elevated-priority /32 routes **bypass aggregation** so that the per-host priority information is
```
nelljerram:
```suggestion
elevated-priority /32 routes **bypass aggregation** so that the per-route priority information is
```

calico/networking/kubevirt/live-migration-bgp.mdx
```diff
@@ -0,0 +1,229 @@
...
+  intermediate routers — they naturally prefer the shorter path. However, it is less precise than
+  communities because it relies on the standard BGP decision process, which may be overridden by
+  other attributes (e.g., MED, weight).
+
+You can use either technique or combine them. The example below uses **communities only**, which
+is the recommended approach for most deployments.
+
+#### Step 1: Create a BGPFilter for the ToR peering
+
+Create a `BGPFilter` that uses the `priority` match field and `operations` to tag routes with
```
nelljerram:
```suggestion
Create a `BGPFilter` that uses the `priority` and `operations` fields to tag routes with
```

calico/networking/kubevirt/live-migration-bgp.mdx
```diff
@@ -0,0 +1,229 @@
...
+BGP communities on export, and reconstruct priority from communities on import. The exact
+community values depend on your network infrastructure; coordinate with your network team.
+
+The following example uses community `65000:100` to mark elevated-priority routes (the migration
+target) and `65000:200` for normal-priority routes (the migration source):
```
nelljerram: I don't think a community should be needed or recommended to indicate "normal priority".  Instead "normal priority" should be indicated just by the _absence_ of `65000:100`, and I think that will then simplify this example.

nelljerram: (It makes the example look more complex than it needs to be.)

calico/networking/kubevirt/live-migration-bgp.mdx
```diff
@@ -0,0 +1,229 @@
...
+  not iBGP peers (which use the automatic `local_pref` mapping).
+- **`communities`** (import rules): Matches routes carrying the specified BGP community.
+- **`operations`**: An ordered list of route modifications applied to matching routes:
+  - `addCommunity`: Adds a BGP community to the route.
+  - `setPriority`: Sets the route's kernel metric (priority).
+
+:::note
+
+The `priority` values in the BGPFilter (512 and 1024) must match the `ipv4ElevatedRoutePriority`
+and `ipv4NormalRoutePriority` values in your `FelixConfiguration`. If you customized those values,
```
nelljerram:
```suggestion
and `ipv4NormalRoutePriority` values in your `FelixConfiguration`. In case you customized those values,
```

nelljerram (changes requested):
Looking great, but a few detailed points.  I've only reviewed the OSS files; obviously please copy any markups for Enterprise as well.

calico/_includes/components/FelixConfig/config-params.json
ctauchen: These files are autogenerated from code. I'm guessing that I need to create an exception for the linter.

calico/networking/kubevirt/live-migration-bgp.mdx
ctauchen: Better: `When you live-migrate a KubeVirt VM to a new host, `

ctauchen (commented):
Looks great, thanks for this! I agree with most of Nell's points. I'm happy to merge this as soon as the content-related questions have been addressed.

calico/networking/kubevirt/live-migration-bgp.mdx
song-jiang: Good point. Thanks

PR #2628, opened 2026-04-07
Use CalicoEnterprise variant instead of TigeraSecureEnterprise

caseydavenport:


rene-dekker (approved):

ctauchen:
@caseydavenport LGTM. 

Note: changes to _api.mdx files will be overwritten unless you make corresponding changes in the code repos.

PR #2636, opened 2026-04-11
Use helm template + kubectl apply for CRD charts

caseydavenport:
The CRD Helm charts have grown large enough to exceed the 1MB Kubernetes Secret size limit that Helm uses for release tracking. Since CRDs don't actually need Helm release management, this switches to `helm template` piped to `kubectl apply --server-side` instead.

Also moves `--create-namespace` to the operator `helm install` step in the enterprise docs, since the CRD step no longer needs a namespace.

Fixes CORE-12633

### DIFF
--- a/calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
+++ b/calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
@@ -93,10 +93,10 @@ To install a standard $[prodname] cluster with Helm:
         }
     </CodeBlock>
 
-1. Create the necessary custom resource definitions.
+1. Install the necessary custom resource definitions.
 
    ```bash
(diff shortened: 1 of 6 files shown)


ctauchen:
@caseydavenport Thanks, looks good. Two questions:

- Do we need to make corresponding changes to Banzai? CC @lwr20 
- Does this apply to Calico Cloud? I similar commands there (for example, https://docs.tigera.io/calico-cloud/get-started/install-automated#install-calico-cloud-as-part-of-an-automated-workflow-1)
   ```
   helm upgrade --install calico-cloud-crds calico-cloud/calico-cloud-crds \. 
   --namespace calico-cloud \
   --create-namespace
   ```

caseydavenport:
@ctauchen yeah it might apply to cloud as well - I'm not familiar with that chart though or where it comes from. My particular changes didn't touch it (but that chart my import my changes?)

ctauchen:
Thanks @caseydavenport. I'll follow up on CC implications separately. LGTM.

PR #2642, opened 2026-04-14
fix: add section on allowing encapsulated traffic for HCP clusters

ppeereb1:
Product Version(s):

Calico, Calico Enterprise

This update clarifies a networking requirement when running KubeVirt workloads in hosted clusters with encapsulation.

By default, Calico drops encapsulated traffic (IP-in-IP or VXLAN) originating from workloads for security reasons. This can prevent communication between KubeVirt pods across nodes.

### DIFF
--- a/calico-enterprise/getting-started/install-on-clusters/openshift/hostedcontrolplanes.mdx
+++ b/calico-enterprise/getting-started/install-on-clusters/openshift/hostedcontrolplanes.mdx
@@ -156,6 +156,13 @@ If you have provided a `calico-resources` configmap and the tigera-operator pod 
 check the output of the init-container with `oc logs -n tigera-operator -l k8s-app=tigera-operator -c create-initial-resources`.
 :::
 
+### Ensure encapsulated traffic is allowed on the hosting cluster
+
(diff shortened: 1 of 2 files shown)


matthewdupre (approved):

ctauchen:
@ppeereb1 Thanks for this PR! Unfortunately we need to hold off on merging it.

Overlay mode in hosted control plane environments hasn't been through our formal support process — the feature was built and validated around the BGP/non-overlay topology. Before we can document overlay as a supported option, engineering needs to do a design review and run targeted testing to make sure there aren't hidden gotchas.

Could you file an RFE for this instead? That would get it onto engineering's radar for proper evaluation. If no code changes end up being needed, the support could potentially be back-applied to current releases.

Closing this for now, but an RFE is the way to get this moving. Let me know if you have questions!

ppeereb1:
@ctauchen will do! And I understand that this needs further testing. But if you follow the current doc it will use ip-in-ip encapsulation on a hosted cluster. And currently pod-to-pod traffic on a hosted cluster is broken. So to docs should be written for non encapsulation if thats currently the only supported way.

PR #2654, opened 2026-04-17
Fix v3 mode install and migration docs

caseydavenport:
Fixes three bugs in the v3 mode install and migration docs:

- The Calico quickstart skips installing the `v1_crd_projectcalico_org.yaml` CRDs before the Tigera Operator. Every other operator-based install guide applies them first - added a step to do the same.
- The native v3 CRDs page's Manifest install tab installs the operator but never creates the `Installation` CR. Added a step to apply `custom-resources.yaml`. (Helm install doesn't need this - the `Installation` CR is in `values.yaml`.)
- The CRD migration page doesn't mention that the `MutatingAdmissionPolicy` feature gate must be enabled before starting the migration. Added it to the Before you begin section, mirroring the wording already on the native v3 CRDs page.

Jira: https://tigera.atlassian.net/browse/CORE-12607

### DIFF
--- a/calico/getting-started/kubernetes/quickstart.mdx
+++ b/calico/getting-started/kubernetes/quickstart.mdx
@@ -94,7 +94,13 @@ In this step, you will:
 
 In this step, you will install Calico in your cluster.
 
-1. Install the Tigera Operator and custom resource definitions.
+1. Install the $[prodname] custom resource definitions.
+
+   ```bash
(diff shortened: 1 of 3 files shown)


ctauchen (approved):
LGTM, pending ok from @MichalFupso

MichalFupso (approved):

PR #2660, opened 2026-04-22
Document Inline DNS policy mode prefix-wildcard limitation

ctauchen:
## Summary

- Add a note to the shared `_domain-names.mdx` partial clarifying that `Inline` DNS policy mode only supports prefix wildcards like `*.example.com`.
- Non-prefix wildcards (e.g. `www.*.com`, `update.*.mycompany.com`) only match under `DelayDeniedPacket` or `NoDelay` mode.
- `Inline` applies to `BPFDNSPolicyMode` (default on kernels 5.17+ / RHEL 5.14+) and `DNSPolicyMode` for iptables.

## Context

Flagged by Shaun Crampton in Slack — users following the DNS policy page with an inline data plane can configure middle-wildcard domains that silently never match.

## Test plan

- [ ] Preview the DNS policy page in calico-cloud and calico-enterprise and confirm the note renders under "Domain name matching"
- [ ] Confirm anchor links to `DNSPolicyMode` and `BPFDNSPolicyMode` resolve on both variants

### DIFF
--- a/calico-cloud/_includes/content/_domain-names.mdx
+++ b/calico-cloud/_includes/content/_domain-names.mdx
@@ -18,6 +18,6 @@ With a single asterisk in any part of the domain name, it matches 1 or more path
 
 :::note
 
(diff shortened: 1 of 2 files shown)


ctauchen:
@fasaxc could you confirm the scope of this note before merge?

I wrote it specifically against `Inline` mode based on the Slack thread (so `DelayDeniedPacket` and `NoDelay` still handle non-prefix wildcards). Is that accurate for both `BPFDNSPolicyMode: Inline` and iptables `DNSPolicyMode: Inline`, or does the restriction extend further?

Also: the `calico-cloud` `DNSPolicyMode` reference table doesn't list `Inline` as a value (calico-enterprise does). Happy to pull that fix into a follow-up if the `Inline` iptables value should be documented there too.

ctauchen:
Thanks @copilot-pull-request-reviewer — fixed in [4c67137b](https://github.com/tigera/docs/pull/2660/commits/4c67137b). Confirmed the `../../reference/` pattern matches other partials in `_includes/content/` (e.g. `_entityrule.mdx`, `_selector-scopes.mdx`).

fasaxc (approved):

PR #2673, opened 2026-04-30
Update OSS private registry docs for combined calico/calico image

caseydavenport:
https://github.com/projectcalico/calico/pull/12225 consolidated most OSS Calico components into a single `calico/calico` image, taking the published image set from 13 down to 3 (`calico/calico`, `calico/node`, `calico/whisker`). A few master docs still pointed at per-component images that no longer exist.

- `calico/releases.json` and `calico/variables.js`: prune obsolete components, add `calico/calico` and `calico/whisker`. Also fixes a pre-existing key mismatch between the two files that was silently dropping typha/calicoctl/whisker from the rendered private-registry pull list.
- `operations/calicoctl/install.mdx`: pull `calico/calico` and run calicoctl from inside it.
- `getting-started/.../config-options.mdx`: dikastes ALP sidecar uses `calico/calico` with `command: ["/usr/bin/calico", "component", "dikastes"]`.
- `reference/kube-controllers/configuration.mdx`: refer to the `calico-kube-controllers` pod container instead of the (now nonexistent) `calico/kube-controllers` image.

The dikastes liveness/readiness probes still call `/healthz` - that's a separate fix.


ctauchen:
LGTM, ready to merge?

ctauchen (approved):

PR #2680, opened 2026-04-30
Fix Windows requirements: IPPool patch is reverted by operator

### LINKED ISSUE #9426: The configuration of Calico network when joining a Windows node to the Kubernetes cluster is not effective.
I first configured a Kubernetes Linux cluster and then added a Windows worker node according to the documentation.
﻿
Official documentation: https://docs.tigera.io/calico/latest/getting-started/kubernetes/windows-calico/requirements
﻿
Under the title "Linux platform requirements," I used the command kubectl patch ippool default-ipv4-ippool -p '{"spec":{"ipipMode":"Never","vxlanMode":"Always"}}'. During two validation processes, the first time I disabled the BGP network mode, and the command did not take effect. Later, I switched to the BGP network and modified the command to kubectl patch ippool default-ipv4-ippool -p '{"spec":{"ipipMode":"Never","vxlanMode":"Never"}}', but this command also did not work.
﻿
However, by directly modifying spec.encapsulation: None through kubectl edit installation default -o yaml -n tigera-operator, I was able to see that the network information changed after running kubectl get ippool default-ipv4-ippool -o yaml. This issue has been resolved, and I am only providing this test record as feedback.
fcb4460f591e4c4b861dd9a742c7db32
397c5c9bfab14eacb99130c0415c0be3

caseydavenport:
The Windows requirements page tells users to disable IPIP via `kubectl patch ippool default-ipv4-ippool ...`, but on operator-managed clusters the operator reconciles IP pools from the `Installation` and silently reverts those edits, so the documented step has no effect. That is what projectcalico/calico#9426 hit.

Split the encapsulation step into two tabs: **Operator managed IP pools** patches `spec.calicoNetwork.ipPools[*].encapsulation` on the `Installation`, and **User managed IP pools** patches the `IPPool` directly. Both modes are valid on operator installs, so this framing works on both OSS and Enterprise.

Applied to OSS main, v3.32, v3.31 and Enterprise main, v3.23-1, v3.22-2.

Fixes projectcalico/calico#9426

### DIFF
--- a/calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
+++ b/calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
@@ -2,6 +2,9 @@
 description: Cluster and Windows host requirements you must meet before installing Calico Enterprise for Windows.
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 # Requirements
 
 ## What's supported in this release
@@ -50,11 +53,31 @@ When using Operator install and Windows hostprocess containers (HPC), see [here 
 
 - At least four Linux Kubernetes worker nodes to run $[prodname]'s cluster-wide components that meets [Linux system requirements](../requirements.mdx), and is installed with $[prodname] v3.5.0+
 - Must not be running in eBPF mode
-- VXLAN or BGP without encapsulation is supported if using $[prodname] CNI. IPIP ($[prodname]'s default encapsulation mode) is not supported. Use the following command to turn off IPIP.
+- VXLAN or BGP without encapsulation is supported if using $[prodname] CNI. IPIP ($[prodname]'s default encapsulation mode) is not supported. Turn off IPIP on the default IP pool.
+
+  <Tabs groupId="install-method">
+  <TabItem label="Operator" value="Operator-0">
+
+  Patch the `Installation` resource. The operator reconciles `IPPool` resources from the `Installation` and will silently revert direct edits to the `IPPool`.
+
+  ```bash
+  kubectl patch installation default --type=json \
+    -p '[{"op":"replace","path":"/spec/calicoNetwork/ipPools/0/encapsulation","value":"VXLAN"}]'
+  ```
+
+  The path `/spec/calicoNetwork/ipPools/0` targets the first IP pool. If your cluster has multiple pools, identify the correct index and adjust the path accordingly. Use `encapsulation: None` instead if you are running BGP without encapsulation.
+
+  </TabItem>
+  <TabItem label="Manifest" value="Manifest-1">
+
+  Disable IPIP through `FelixConfiguration`.
 
   ```bash
   kubectl patch felixconfiguration default -p '{"spec":{"ipipEnabled":false}}'
   ```
+
+  </TabItem>
+  </Tabs>
 
 - If using $[prodname] IPAM, strict affinity of IPAM configuration must be set to `true`.
 
--- a/calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
+++ b/calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
@@ -2,9 +2,6 @@
 description: Cluster and Windows host requirements you must meet before installing Calico Enterprise for Windows.
 ---
 
-import Tabs from '@theme/Tabs';
-import TabItem from '@theme/TabItem';
-
 # Requirements
 
 ## What's supported in this release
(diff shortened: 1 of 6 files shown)


calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
```diff
@@ -50,12 +53,32 @@ When using Operator install and Windows hostprocess containers (HPC), see [here
 
 - At least four Linux Kubernetes worker nodes to run $[prodname]'s cluster-wide components that meets [Linux system requirements](../requirements.mdx), and is installed with $[prodname] v3.5.0+
 - Must not be running in eBPF mode
-- VXLAN or BGP without encapsulation is supported if using $[prodname] CNI. IPIP ($[prodname]'s default encapsulation mode) is not supported. Use the following command to turn off IPIP.
+- VXLAN or BGP without encapsulation is supported if using $[prodname] CNI. IPIP ($[prodname]'s default encapsulation mode) is not supported. Turn off IPIP on the default IP pool.
+
+  <Tabs groupId="install-method">
+  <TabItem label="Operator" value="Operator-0">
+
+  Patch the `Installation` resource. The operator reconciles `IPPool` resources from the `Installation` and will silently revert direct edits to the `IPPool`.
```
ctauchen: Something along these lines might be clearer:

To turn off IPIP for the default IP pool, patch the `Installation` resource by running the following command: 

  ```bash
  kubectl patch installation default --type=json \
    -p '[{"op":"replace","path":"/spec/calicoNetwork/ipPools/0/encapsulation","value":"VXLAN"}]'
  ```

To turn off IPIP for multiple IP pools, repeat the same command using the index for each IP pool.

  ```bash
  kubectl patch installation default --type=json \
    -p '[{"op":"replace","path":"/spec/calicoNetwork/ipPools/<ip-pool-index>/encapsulation","value":"VXLAN"}]'
  ```
  Replace `<ip-pool-index>` with the index of an IP pool.

calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
```diff
@@ -50,12 +53,32 @@ When using Operator install and Windows hostprocess containers (HPC), see [here
...
+
+  ```bash
+  kubectl patch installation default --type=json \
+    -p '[{"op":"replace","path":"/spec/calicoNetwork/ipPools/0/encapsulation","value":"VXLAN"}]'
+  ```
+
+  The path `/spec/calicoNetwork/ipPools/0` targets the first IP pool. If your cluster has multiple pools, identify the correct index and adjust the path accordingly. Use `encapsulation: None` instead if you are running BGP without encapsulation.
+
+  </TabItem>
+  <TabItem label="Manifest" value="Manifest-1">
```
ctauchen: We don't support manifest-based installation for Calico Enterprise. Unless there's an exception for Windows, we should have only the operator method here.

ctauchen (commented):
Some suggested wording, and a question about having manifest paths in CE.

calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
caseydavenport: Yeah, I guess it really shouldn't be "Operator vs Manifest" but rather "Operator managed IP pools" and "User managed IP pools"

calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
caseydavenport: Yep, took your wording. Reorganized so the command comes first, then the operator reconciliation note, then the multi-pool / BGP-without-encap caveats.

calico-enterprise/getting-started/install-on-clusters/windows-calico/requirements.mdx
caseydavenport: Reframed the tabs as **Operator managed IP pools** (patch the `Installation`) vs **User managed IP pools** (patch the `IPPool` directly), per your follow-up. That works on both OSS and Enterprise since both kinds of IP pool management are valid on operator installs. Restored the Tabs imports on the Enterprise pages.

caseydavenport:
@ctauchen ready for another look. Took your wording on the Enterprise pages, reframed the tabs as **Operator managed IP pools** vs **User managed IP pools** (works on both OSS and Enterprise since both pool-management modes are valid on operator installs), and added a BGP-without-encap variant under the user-managed tab on the OSS pages.

ctauchen:
Thanks @caseydavenport. Please rebase with latest build fixes, then you should be good to go.

PR #2699, opened 2026-05-07
DOCS-2914: Remove false statement about federation controller

ctauchen:
The federation controller is enabled even when the ENABLED_CONTROLLERS
option is not explicitly specified.

This commit removes the false statement.

DOCS-2914

Product Version(s):

Issue:

https://tigera.atlassian.net/browse/DOCS-2914

Link to docs preview:

SME review:
- [ ] An SME has approved this change. 

DOCS review:
- [ ] A member of the docs team has approved this change.

Additional information:

Merge checklist:

- [ ] Deploy preview inspected wherever changes were made 
- [ ] Build completed successfully
- [ ] Test have passed

PR #2701, opened 2026-05-07
DOCS-2914: Collapse CE kube-controllers config page into resource reference

ctauchen:
## Summary

- Deletes `calico-enterprise/reference/component-resources/kube-controllers/configuration.mdx`. The page documents a manifest install path that doesn't exist for CE (operator-only), and its env-var guidance (e.g. `ENABLED_CONTROLLERS`) isn't applicable to operator-managed installs — making it a source of confusion rather than help.
- Folds the conceptual "what each controller does" prose into the existing `KubeControllersConfiguration` resource page (`calico-enterprise/reference/resources/kubecontrollersconfig.mdx`), which becomes the canonical entry point.
- Updates inbound links in `services-controller.mdx` and `decommissioning-a-node.mdx` to deep-link the relevant section of the resource page.
- Adds a Netlify redirect from the old URL → resource page; updates a stale legacy redirect that pointed at the now-deleted page.
- Removes the sidebar entry for the deleted page.

## Status

**Draft — CE current only, for shape review.** If the approach looks right, I'll propagate the same pattern to:

- `calico-enterprise_versioned_docs/version-3.20-2/`
- `calico-enterprise_versioned_docs/version-3.21-2/`
- `calico-enterprise_versioned_docs/version-3.22-2/`
- `calico-enterprise_versioned_docs/version-3.23-1/`
- `calico-cloud/`
- `calico-cloud_versioned_docs/version-22-2/`

Then add per-version redirects and verify the build before un-drafting.

## Context

Supersedes #2699, which patched a single incorrect sentence on the page being removed (`"The federation controller is disabled by default if ENABLED_CONTROLLERS is not explicitly specified"`). That fix is correct in isolation but applied inside a structurally incorrect Manifest tab — the cleaner outcome is to retire the page.

Jira: https://tigera.atlassian.net/browse/DOCS-2914

## Things worth your eyes

1. Voice/placement of the new bullet list at the top of the resource page (Node / Federated services / Load balancer) — does it belong inline at the top, or split into a dedicated "About the controllers" section between intro and Sample YAML?
2. Pre-existing inconsistency: `LoadBalancerController` has its own section on the resource page but isn't listed in the Controllers table. I added it to the new bullet list but did not patch the table — happy to bundle that in if you prefer.
3. Redirect format: confirm the pattern in `static/_redirects` is right. I added one product-prefixed redirect for `/calico-enterprise/latest/...` and updated the legacy un-prefixed line. Older versioned URLs (3.20, 3.21, 3.23) get their own redirects when I propagate.

## Test plan

- [ ] Resource page renders; the new prose at the top reads cleanly
- [ ] Sidebar shows the Calico Enterprise Kubernetes controllers category with only "prometheus" under it
- [ ] `services-controller.mdx` "Federated services controller" link resolves to `#federatedservicescontroller`
- [ ] `decommissioning-a-node.mdx` "node controller" link resolves to `#nodecontroller`
- [ ] Old URL `/calico-enterprise/latest/reference/component-resources/kube-controllers/configuration` redirects to the resource page
- [ ] Build completes without orphan-link warnings


ctauchen:
PTAL @sabags
For loadbalancer stuff @MichalFupso 

One thing led to another. Turns out, this statement was on a manifest-based page, which isn't supported in CE. What followed was this:
1. I removed the page completely.
2. I added relevant information to the resource page.
3. Doing that, I noticed that we hadn't added some loadbalancer updates there.

FYI @matthewdupre @pasanw

MichalFupso (approved):

pasanw (approved):

PR #2717, opened 2026-05-11
Update enterprise private registry docs for combined tigera/calico image

caseydavenport:
Enterprise companion to https://github.com/tigera/docs/pull/2673. The enterprise side of the mono-image consolidation (https://github.com/tigera/calico-private/pull/11681, https://github.com/tigera/operator/pull/4782) drops standalone builds for voltron, guardian, ui-apis, queryserver, linseed, l7-collector, ingress-collector, packetcapture, es-gateway, elasticsearch-metrics, prometheus-service, and compliance (server/controller/reporter/snapshotter). Operator points each one at `tigera/calico` with `calico component <name>` as the container command.

- Add a `tigera/calico` entry to `calico-enterprise/releases.json` so it shows up in the rendered private-registry pull list.
- Remove the folded components so we stop telling users to pull images that no longer exist.

`intrusion-detection-controller` stays in the list - that image is now a thin layer on top of `tigera/calico` carrying the GeoIP databases, so it's still published. Same goes for `cnx-manager` (the UI is still its own image; only the `ui-apis` backend got folded). `dikastes` also stays standalone in enterprise for now, unlike OSS.

```release-note
None
```


ctauchen:
LGTM, thanks

PR #2724, opened 2026-05-12
docs: L7 logs restructure and eBPF collector

lucastigera:
## Summary

Restructures the **L7 logs** docs section to introduce the new eBPF-based collector alongside the existing Envoy collector, with a shared overview and per-collector enable pages.

The Istio Waypoint collector will land in a follow-up PR — its placeholder page is set to `draft: true` so it's visible during local development but excluded from production builds.

## L7 logs section, after this PR

| Sidebar entry | File | What it is |
| --- | --- | --- |
| Overview | `overview.mdx` | **New.** Big picture, value, the available collectors, and a comparison table to help pick one. Carries the shared "before you begin" log-storage note. |
| Enable eBPF collector | `enable-ebpf.mdx` | **New.** Enable via `L7ObservabilityEnabled` on `FelixConfiguration`. Limitations: kernel 5.17+, HTTP/1.0–1.1 only, TLS metadata only. |
| Enable Waypoint collector | `enable-waypoint.mdx` | **Placeholder, dev-only.** Marked `draft: true` so it's hidden in production builds. Content lands in the Waypoint PR. |
| Enable Envoy collector | `configure.mdx` | Existing page; sidebar label updated. Still labelled deprecated and being phased out. |
| Data types | `datatypes.mdx` | Refreshed schema reference. New **Populated by** column. Adds `protocol`, `protocol_version`, `collector_type`, `collector_name`, and `tls_*` fields. HTTP-only / TLS-only fields tagged inline (e.g. `_HTTP only._ URL that the request was made against.`). |
| Aggregation | `aggregation.mdx` | **New.** Documents the `L7LogsFileAggregation*` fields (including the new `L7LogsFileAggregationTLSSNI`) and includes an **interactive aggregation playground** — a React component (`_includes/components/L7AggregationSandbox`) where readers toggle aggregator checkboxes and watch sample events regroup with two-tone highlight (green = added / count up, red = removed / count down) and per-group expand/collapse. |

## Other changes

- `dashboards.mdx`, `get-started-cem.mdx`, `kibana.mdx`: external "L7 logs" links repointed from `configure.mdx` (Envoy-only) to the new `overview.mdx`.
- `alp.mdx`: left pointing at `configure.mdx#kibana` per maintainer note (page will be deprecated).
- Vale lint clean on the new pages (`data plane`, `Waypoint`, `plain text`).

## Open items / known gaps

- `L7ObservabilityEnabled` and `L7LogsFileAggregationTLSSNI` won't render in the Felix Configuration tables until `_includes/components/FelixConfig/config-params.json` is regenerated from upstream calico-private. Field-level anchors (`#L7ObservabilityEnabled`, `#L7LogsFileAggregationTLSSNI`) will activate at that time.
- Versioned-tree mirroring intentionally not done in this PR — changes land in the `next` tree (`calico-enterprise/`) only.

## Test plan

- [x] Local Docusaurus dev build (`yarn start-next`) compiles and renders all six pages.
- [x] Sidebar order verified in `.docusaurus` route data.
- [x] Aggregation playground toggles cleanly, columns hold their widths, group expand/collapse works.
- [x] No remaining bare references to renamed files (`configure-bpf.mdx`, `configure-istio-waypoint.mdx`, `datatypes-wip.mdx`, `datatypes-legacy.mdx`).
- [x] Vale CI lint passing.
- [ ] Production build (`yarn build-next`) excludes `enable-waypoint.mdx` — to be verified by reviewers.


calico-enterprise/observability/elastic/l7/overview.mdx
```diff
@@ -0,0 +1,55 @@
+---
+description: Overview of Calico Enterprise L7 logs, the available collectors, and how to choose between them.
+---
+
+# L7 logs
+
+## Big picture
```
ctauchen: rm

calico-enterprise/observability/elastic/l7/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+Platform operators and development teams use L7 logs to:
+
+- See how applications interact, request by request.
+- Spot anomalous behavior — attempts to access restricted URLs, unexpected user agents, scans for particular paths.
+- Inspect TLS sessions by SNI, negotiated version, and cipher suite.
+
+L7 logs complement L3/4 flow logs, which expose only the source/destination pair without request-level detail.
+
+## Collectors
```
ctauchen, line 19: Might be worth saying what a collector actually is.

calico-enterprise/observability/elastic/l7/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+$[prodname] supports multiple L7 collectors. They all feed the same aggregation and reporting pipeline; each entry on the log file is tagged with `collector_type` and `collector_name` so consumers can filter or join sources as needed.
```
ctauchen: It's not clear what the distinction is between type and name. Are you saying we can have an epbf type called A and another ebpf type called B?

calico-enterprise/observability/elastic/l7/overview.mdx
```diff
@@ -0,0 +1,55 @@
...
+| | eBPF | Envoy (legacy) |
+| --- | --- | --- |
+| Additional infrastructure | None — runs inside Felix | Sidecar Envoy per workload |
+| Protocol coverage | HTTP/1.x, TLS handshake | HTTP |
+| Opt-in granularity | Cluster-wide (FelixConfiguration) | Per workload (label + annotation) |
+| Service-mesh compatibility | Any data plane | Not compatible with Istio |
+
+Multiple collectors can coexist in the same cluster, with each L7 log entry tagged by `collector_type` and `collector_name` so consumers can filter or join sources as needed.
+
+:::note
```
ctauchen: This should not be buried in a note. Probably best as a prerequisite to each 'enable' procedure.

For clarity, consider working through an example. "For example, if you want to maintain a 7-day retention period for a three-node cluster, you need to add at least 21 GB to your existing log storage."

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
+---
```
ctauchen: The file name is misleading. Perhaps `enable-ebpf-collector` or simlar.

ctauchen: same for waypoint.

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
...
+description: Enable the kernel TCP-layer eBPF collector to capture HTTP and TLS L7 logs without a proxy or sidecar.
+---
+
+# Enable eBPF collector
```
ctauchen: What about something more like "Collect L7 logs using the eBPF collector / Waypoint collector"

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+Enable the kernel TCP-layer eBPF collector to capture HTTP request/response and TLS handshake metadata without deploying a proxy or injecting sidecars. For background on what the collector does and how it compares to other options, see [L7 logs overview](overview.mdx).
+
+## Before you begin
+
+### Limitations
+
+* Requires Linux kernel 5.17 or later on every node.
```
ctauchen: I'd call this a prerequisite. Consider removing the Before you begin. Then do this:

Limitations
* etc
Prerequisites
* kernel version

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+```bash
+kubectl patch felixconfiguration default --type=merge -p '{"spec":{"l7ObservabilityEnabled":true}}'
+```
+
+To disable the collector, set the field back to `false`. Felix tears down its BPF programs and cleans up pinned maps; no node reboot is required.
+
+```bash
+kubectl patch felixconfiguration default --type=merge -p '{"spec":{"l7ObservabilityEnabled":false}}'
+```
```
ctauchen: Is there an expectation that it wouldn't be cleaned up? I'm wondering if we can remove all of this.

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
...
+### Limitations
+
+* Requires Linux kernel 5.17 or later on every node.
+* HTTP parsing covers HTTP/1.0 and HTTP/1.1 only, because their requests and responses travel as plain text on the wire. HTTP/2 and HTTP/3 are not captured as HTTP, and encrypted HTTP traffic surfaces only as TLS metadata.
+* HTTP capture tracks only the most recent request per socket; when a single connection carries multiple back-to-back requests, some may be missed.
+* TLS capture is limited to handshake metadata (SNI, version, cipher suite). The collector does not decrypt application traffic.
+
+## Enable the collector
+
+Enable the eBPF L7 collector by setting the [L7ObservabilityEnabled](../../../reference/component-resources/node/felix/configuration.mdx#L7ObservabilityEnabled) field to `true` on the default `FelixConfiguration`. The setting is independent of `bpfEnabled` — the collector works with any data plane.
```
ctauchen: For a single step procedure, use an unordered list:

```suggestion
* Enable the eBPF L7 collector by setting the [L7ObservabilityEnabled](../../../reference/component-resources/node/felix/configuration.mdx#L7ObservabilityEnabled) field to `true` on the default `FelixConfiguration`. The setting is independent of `bpfEnabled` — the collector works with any data plane.
```

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
...
+## View L7 logs
+
+The eBPF collector writes L7 events as JSON to a log file on each node. By default the file is at `/var/log/calico/l7logs/l7.log`. The directory is controlled by the [L7LogsFileDirectory](../../../reference/component-resources/node/felix/configuration.mdx#L7LogsFileDirectory) field on `FelixConfiguration`.
+
+To tail the log on a node:
+
+```bash
+tail -f /var/log/calico/l7logs/l7.log
+```
+
```
ctauchen:
```suggestion
## Additional information

* L7 log data types
```

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
```diff
@@ -0,0 +1,42 @@
...
+## View L7 logs
+
+The eBPF collector writes L7 events as JSON to a log file on each node. By default the file is at `/var/log/calico/l7logs/l7.log`. The directory is controlled by the [L7LogsFileDirectory](../../../reference/component-resources/node/felix/configuration.mdx#L7LogsFileDirectory) field on `FelixConfiguration`.
+
+To tail the log on a node:
+
+```bash
+tail -f /var/log/calico/l7logs/l7.log
+```
```
ctauchen: I think this can all move to a single line after the enable step.

This could also act as a verification step, come to think of it.

1. To verify that L7 logs are being collected, look for output in the default log file location on each node:

calico-enterprise/observability/elastic/l7/aggregation.mdx
```diff
@@ -0,0 +1,23 @@
+---
+description: Control how Calico Enterprise aggregates L7 log entries by selecting which fields participate in the aggregation key.
+---
+
+import L7AggregationKeyDemo from '../../../_includes/components/L7AggregationKeyDemo';
+
+# L7 log aggregation
```
ctauchen: I think the new component is handy for illustrating how aggregation works. At this stage, I have two thoughts.

1. To what extent is this L7-specific? I get that the fields may be different, but it seems that the main idea here is to help people understand aggregation generally, not L7 aggregation. If that's the case, perhaps it would be better to use a tool like this in a general doc about aggregation?

2. For me, it would be useful to see all parts of this dynamic in action. 1) a full list of all the communications 2) the configuration (as a code block, rendered dynamically according to the checkboxes, and 3) the actual flow logs (what the log would actually look like, not the isolated terms as cells in a table.

calico-enterprise/observability/elastic/l7/overview.mdx
lucastigera: Do you mean removing the entire Big Picture session, or just the title?

calico-enterprise/observability/elastic/l7/overview.mdx
lucastigera: We could have multiple collectors under the same type. For example, our current eBPF collector is attached to the TCP layer, so `collector_name = ebpf_tcp`. We could also have an eBPF collector attached through uprobes, in which case it would be `collector_name = ebpf_uprobe`.

That said, I don't think we need to call that out here since the distinction isn't really clear in this version. I removed `collector_name` to avoid confusion.

calico-enterprise/observability/elastic/l7/enable-ebpf.mdx
lucastigera: Is your question whether the disable path is clear enough that it doesn’t need to be documented? If so, I agree - we are just flipping back the field. I removed the disable block.

calico-enterprise/observability/elastic/l7/aggregation.mdx
lucastigera: Do you mean moving this page to `observability/elastic/aggregation.mdx` and making it more generic?

If so, I created a new page there with a different interaction, based on your suggestion for something we could test.

WDYT?

calico-enterprise/observability/elastic/l7/overview.mdx
lucastigera: I removed the title, let me know if that is what you meant.

alexh-tigera:
@lucastigera WDYT about adding a link to observability/elastic/l7/configure#view-l7-logs-in-the-web-console in the `enable-ebpf-collector.mdx` file?

calico-enterprise/observability/elastic/l7/enable-ebpf-collector.mdx
```diff
@@ -0,0 +1,40 @@
...
+* Linux kernel 5.17 or later on every node.
+* Additional log storage for the L7 logs. L7 logs require a minimum of 1 additional GB of log storage per node, per one-day retention period. Adjust your [Log Storage](../../../operations/logstorage/adjust-log-storage-size.mdx) before enabling the collector. For example, to maintain a 7-day retention period on a three-node cluster, add at least 21 GB to your existing log storage.
+
+## Enable the collector
+
+* Enable the eBPF L7 collector by setting the [L7ObservabilityEnabled](../../../reference/component-resources/node/felix/configuration.mdx#L7ObservabilityEnabled) field to `true` on the default `FelixConfiguration`. The setting is independent of `bpfEnabled` — the collector works with any data plane.
+
+```bash
+kubectl patch felixconfiguration default --type=merge -p '{"spec":{"l7ObservabilityEnabled":true}}'
+```
```
ctauchen: Must be indented to stay in line with the list item above.

calico-enterprise/observability/elastic/l7/enable-ebpf-collector.mdx
```diff
@@ -0,0 +1,40 @@
...
+* HTTP parsing covers HTTP/1.0 and HTTP/1.1 only, because their requests and responses travel as plain text on the wire. HTTP/2 and HTTP/3 are not captured as HTTP, and encrypted HTTP traffic surfaces only as TLS metadata.
+* HTTP capture tracks only the most recent request per socket; when a single connection carries multiple back-to-back requests, some may be missed.
+* TLS capture is limited to handshake metadata (SNI, version, cipher suite). The collector does not decrypt application traffic.
+
+## Prerequisites
+
+* Linux kernel 5.17 or later on every node.
+* Additional log storage for the L7 logs. L7 logs require a minimum of 1 additional GB of log storage per node, per one-day retention period. Adjust your [Log Storage](../../../operations/logstorage/adjust-log-storage-size.mdx) before enabling the collector. For example, to maintain a 7-day retention period on a three-node cluster, add at least 21 GB to your existing log storage.
+
+## Enable the collector
```
ctauchen, line 20: I think we can combine 'enable' and 'verify' under the enable section. Two steps, where step 2 verifies that step 1 did its job.

calico-enterprise/observability/elastic/l7/enable-ebpf-collector.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+* Enable the eBPF L7 collector by setting the [L7ObservabilityEnabled](../../../reference/component-resources/node/felix/configuration.mdx#L7ObservabilityEnabled) field to `true` on the default `FelixConfiguration`. The setting is independent of `bpfEnabled` — the collector works with any data plane.
```
ctauchen: independent of `bpfEnabled`

Consider making the implied meaning here explicit. People might not know that particular key. What you mean, I think, is that you can use the eBPF collector even if you're CE installation runs on nftables or whatever.

Come to think of it, that would be valuable information for a person who is evaluating the L7 logs feature. Some might not get as far as the enablement step before they give up, thinking that it's only for people using eBPF as a data plane. Consider making this clear in the L7 overview and intro to this page.

calico-enterprise/observability/elastic/l7/datatypes.mdx
```diff
@@ -1,36 +1,47 @@
 ---
-description: Reference of key/value fields that Calico Enterprise sends to Elasticsearch for L7 logs, including durations, byte counts, and HTTP request metadata.
+description: Reference of key/value fields that Calico Enterprise sends to Elasticsearch for L7 logs, with the collector that populates each field.
 ---
 
 # L7 log data types
 
 ## Big picture
```
ctauchen: I don't think we need this subhead. A small intro and table is okay.

calico-enterprise/observability/elastic/l7/aggregation.mdx
ctauchen: Yes, that's just what I meant. I think it works better there and in that format. If you're happy with that, we can leave it there and remove this page.

lucastigera:
> @lucastigera WDYT about adding a link to observability/elastic/l7/configure#view-l7-logs-in-the-web-console in the `enable-ebpf-collector.mdx` file?

Nice catch @alexh-tigera . There are three things holding me back from adding that link. Let me know if you agree.

1. That page has a deprecation notice at the top. I could add a Service Graph section to the eBPF collector page instead, but then the other two points come into play.
2. The Service Graph page only covers HTTP metrics, not TLS. So linking to it would only be partially correct!? hehe
3. The goal of this first iteration of the eBPF collector is to collect and expose the data in the L7 log file. Plumbing that data all the way to the Manager UI was an extra step to prepare for the next phase, which is to build a visual interface for users to interact with. I think we need more time to figure out the best way to present that information.

Those are the reasons I only pointed users to the L7 log file in the **Verify the collector** section of the eBPF collector page.

alexh-tigera:
> I think we need more time to figure out the best way to present that information.

Good point. Let's leave it for now.

PR #2729, opened 2026-05-15
docs: note Kibana 7→8 migration known issue on 3.21 and 3.22 release notes

tianfeng92:
## Summary

- Add an upgrade note on the 3.21.8 and 3.22.5 release sections calling out the known Kibana 7→8 saved-object migration issue when upgrading from Calico Enterprise 3.20 or earlier.
- The fix shipped in 3.21.8 and 3.22.5 — the note is placed on those releases to direct readers to use them (or later).

## Test plan

- [ ] Verify the rendered note appears on the 3.21.8 section on the 3.21 release-notes page.
- [ ] Verify the rendered note appears on the 3.22.5 section on the 3.22 release-notes page.

### DIFF
--- a/calico-enterprise_versioned_docs/version-3.21-2/release-notes/index.mdx
+++ b/calico-enterprise_versioned_docs/version-3.21-2/release-notes/index.mdx
@@ -272,6 +272,12 @@ To update an existing installation of Calico Enterprise 3.21, see [Install a pat
 
 May 25, 2026
 
+:::note[Upgrading from Calico Enterprise 3.20 or earlier]
+
+Calico Enterprise 3.21 and later upgrade Kibana from 7.x to 8.x, and there is a known issue with the Kibana 7 → 8 saved-object migration when upgrading from Calico Enterprise 3.20 or earlier. This release includes the fix — upgrade directly to 3.21.8 or later to avoid it.
+
+:::
+
 #### Bug fixes
 
 * Fixes an issue where `HostEndpoint` policies block UDP return traffic for SNAT'd pod egress when using the eBPF dataplane.
--- a/calico-enterprise_versioned_docs/version-3.22-2/release-notes/index.mdx
+++ b/calico-enterprise_versioned_docs/version-3.22-2/release-notes/index.mdx
@@ -369,6 +369,12 @@ To update an existing installation of Calico Enterprise 3.22, see [Install a pat
 
 May 20, 2026
 
+:::note[Upgrading from Calico Enterprise 3.20 or earlier]
+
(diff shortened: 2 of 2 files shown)


calico-enterprise_versioned_docs/version-3.21-2/release-notes/index.mdx
```diff
@@ -220,6 +220,12 @@ To update an existing installation of Calico Enterprise 3.21, see [Install a pat
 
 May 13, 2026
 
+:::note[Upgrading from Calico Enterprise 3.20 or earlier]
+
+Calico Enterprise 3.21 and later upgrade Kibana from 7.x to 8.x, and there is a known issue with the Kibana 7 → 8 saved-object migration when upgrading from Calico Enterprise 3.20 or earlier. The fix is included in this release — upgrade directly to 3.21.7 or later to avoid it.
```
rene-dekker: Shouldn't it be the upcoming v3.21.8?

ctauchen:
@tianfeng92 Can you provide more information about this? This seems ultimately to be a bug fix statement, so I'm not sure why we need to add a special admonition.

calico-enterprise_versioned_docs/version-3.22-2/release-notes/index.mdx
```diff
@@ -293,6 +293,12 @@ Use v3.22.4 instead, which includes a fix for this issue.
 
 :::
 
+:::note[Upgrading from Calico Enterprise 3.20 or earlier]
+
+Calico Enterprise 3.21 and later upgrade Kibana from 7.x to 8.x, and there is a known issue with the Kibana 7 → 8 saved-object migration when upgrading from Calico Enterprise 3.20 or earlier. The fix is included in this release — upgrade directly to 3.22.3 or later to avoid it.
```
rene-dekker: Shouldn't this be 3.22.5?

calico-enterprise_versioned_docs/version-3.21-2/release-notes/index.mdx
tianfeng92: Good catch — updated both to reference the upcoming releases (3.21.8 and 3.22.5).

tianfeng92:
> @tianfeng92 Can you provide more information about this? This seems ultimately to be a bug fix statement, so I'm not sure why we need to add a special admonition.

@ctauchen  The admonition is intentional because this is an upgrade-path hazard, not a typical runtime bug. Users upgrading from 3.20 or earlier will hit a Kibana crashloop during the saved-object migration — by the time they see it, they're already stuck in a broken state. The note is placed at the top of the release section so it's visible before someone starts the upgrade, rather than buried in the bug-fix list where it reads like a post-facto fix. Once 3.21.8 / 3.22.5 ship with the fix included, we can remove the admonition and leave just the bug-fix line item.

rene-dekker (approved):

PR #2751, opened 2026-05-26
Rectify RN message for Openshift IdP

rene-dekker:
Correct the namespace for the v3.21 pages where the manager is installed in the tigera-manager namespace.


ctauchen:
LGTM

PR #2761, opened 2026-05-29
Document Kubernetes 1.34+ requirement for native v3 CRDs

### LINKED ISSUE #12863: CRD migration to projectcalico.org/v3 does not work on K8s 1.33 where MutatingAdmissionPolicy is Alpha, not beta
The guide here states that as long as you have the MutatingAdmissionPolicy feature gate enabled you are good, but then incorrectly states:

```
Native projectcalico.org/v3 CRDs rely on MutatingAdmissionPolicies for defaulting, which
are currently a beta Kubernetes feature and are not enabled by default.
```

This is NOT true.  They are beta in 1.34+ but in 1.33 they are alpha but still functioning.

The Installation object has this:

```
  - lastTransitionTime: "2026-05-29T14:36:17Z"
    message: 'Error listing MutatingAdmissionPolicies: no matches for kind "MutatingAdmissionPolicy"
      in version "admissionregistration.k8s.io/v1beta1"'
    observedGeneration: 4
    reason: ResourceReadError
    status: "True"
    type: Degraded
```

But in K8s 1.33 MutatingAdmissionPolicy is ALPHA:

```
 k explain MutatingAdmissionPolicy
GROUP:      admissionregistration.k8s.io
KIND:       MutatingAdmissionPolicy
VERSION:    v1alpha1
```

The Operator should not assume the API is beta and should work with K8s 1.33 where the API is alpha.

caseydavenport:
The native `projectcalico.org/v3` CRD docs said enabling the `MutatingAdmissionPolicy` feature gate was all you needed, but left out the Kubernetes version floor. MutatingAdmissionPolicy is only available as the beta `admissionregistration.k8s.io/v1beta1` API on Kubernetes 1.34 and later. On 1.33 and earlier it exists only as alpha (`v1alpha1`), which the operator and webhooks don't support, so the install/migration fails with a `ResourceReadError` on the `Installation`.

Updated the install, migration, and reference pages (Calico and Calico Enterprise, including the versioned snapshots) to call out the 1.34+ requirement, that 1.34/1.35 still need the feature gate enabled, and that the feature is GA and on by default in 1.36+.

Also a drive-by fix: the OpenShift CLI install link in the identity provider docs pointed at docs.okd.io, whose TLS certificate has expired and was failing the link check. Repointed it to the equivalent docs.openshift.com page.

Fixes https://github.com/projectcalico/calico/issues/12863

### DIFF
--- a/calico-enterprise/operations/cnx/configure-identity-provider.mdx
+++ b/calico-enterprise/operations/cnx/configure-identity-provider.mdx
@@ -186,7 +186,7 @@ When configuring your cluster, you may be asked for the following inputs:
 </TabItem>
 <TabItem label="OpenShift" value="OpenShift-2">
 
-1. Install the [OpenShift CLI (oc)](https://docs.okd.io/4.9/cli_reference/openshift_cli/getting-started-cli.html).
+1. Install the [OpenShift CLI (oc)](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html).
 
 1. Create values for some required variables. `MANAGER_URL` is the URL where the $[prodname] web console will be accessed,
    `CLUSTER_DOMAIN` is the domain (excl. port) where your OpenShift cluster is accessed (run `oc status` to get it) and `CLIENT_SECRET` is a value of your choosing.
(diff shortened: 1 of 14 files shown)


ctauchen:
LGTM, thanks

ctauchen (approved):

PR #2763, opened 2026-06-01
Remove FIPS mode docs page

caseydavenport:
Removes the FIPS mode documentation from the current (next) Open Source docs, since FIPS mode is being removed from Calico (https://github.com/projectcalico/calico/pull/12883, https://github.com/tigera/operator/pull/4875).

Deletes the FIPS how-to page and its operations index card and sidebar entry. The released versioned docs (3.29 through 3.32) are left as-is since FIPS shipped in those releases, and the generated installation API reference will pick up the deprecation on its next regeneration against the operator change.


ctauchen:
LGTM, pending approvals from CS and PM.

PR #2764, opened 2026-06-03
[CORE-12848] OpenStack doc fixes

nelljerram:
- System requirements should say "tap*" instead of "cali*"

- System requirements should say "v3.30+" instead of just "v3.30"

- In the discussion of software components, clarify that some of them need to be provisioned by the user, and that an alternative routing daemon may be used instead of BIRD

- Update uses of `networking-calico` as a Git repository to say   `projectcalico/calico` instead.  `networking-calico` in this sense has been incorrect since we created the OSS monorepo.

- Update other remaining uses of "networking-calico" to just "Calico".  The history of the "networking-calico" term is not of interest to our readers, and there is no useful distinction to be implied by saying "networking-calico" instead of just "Calico".

### DIFF
--- a/calico/_includes/partials/_system-requirements.mdx
+++ b/calico/_includes/partials/_system-requirements.mdx
@@ -5,9 +5,14 @@ import Link from '@docusaurus/Link';
 
 * x86-64, arm64, ppc64le, or s390x processor
 
+<If condition={props.platform !== 'openstack'}>
 * Calico must be able to manage `cali*` interfaces on the host.
   When IPIP is enabled (the default), Calico also needs to be able to manage `tunl*` interfaces.
   When VXLAN is enabled, Calico also needs to be able to manage the `vxlan.calico` interface.
+</If>
+<If condition={props.platform === 'openstack'}>
+* Calico must be able to manage `tap*` interfaces on the host.
+</If>
 
 * Linux kernel 5.10 or later with [required dependencies](#kernel-dependencies).
   The following distributions have the required kernel, its dependencies, and are known to work well with $[prodname].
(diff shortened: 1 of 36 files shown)


calico/_includes/partials/_system-requirements.mdx
```diff
@@ -226,9 +231,9 @@ $[prodname] supports the [OpenShift Container Platform](https://docs.openshift.c
 ## OpenStack requirements
 
 The Calico Neutron driver is written in Python 3 and so requires an OpenStack release that runs with Python 3.
-Subject to that, we aim to develop and maintain the Neutron driver for Calico (networking-calico) so that its master code works with OpenStack master or any previous Python 3 release, on any operating system, independently of the deployment mechanism that is used to install it.
+Subject to that, we aim to develop and maintain the Neutron driver for Calico so that its master code works with OpenStack master or any previous Python 3 release, on any operating system, independently of the deployment mechanism that is used to install it.
 
-However, we recommend using OpenStack Caracal or later, and our active support and testing of Calico v3.30 with OpenStack is with Caracal.
+However, we recommend using OpenStack Caracal or later, and our active support and testing of Calico v3.30+ with OpenStack is with Caracal.
```
ctauchen:
```suggestion
However, we recommend using OpenStack Caracal or later, and our active support and testing of Calico $[version] with OpenStack is with Caracal.
```
For any given version, let's talk only about that version.

ctauchen (commented):
One small suggestion, otherwise LGTM. Ok to move to versions.

nelljerram:
@ctauchen OK, I've copied to all the versions now.  I'm happy for this to merge if you are also happy.

PR #2798, opened 2026-06-23
[EV-6717] Document enabling Calico Ingress Gateway via the Helm gatewayAPI toggle

xiumozhan:
## What

Documents the new ability to enable Calico Ingress Gateway (CIG) at Helm install time via the tigera-operator chart's `gatewayAPI.enabled` toggle. Enterprise-only, in the current ("next") `calico-enterprise/` docs.

Following review feedback, this keeps the existing kubectl flow as the documented enablement process and records the Helm capability where our docs normally cover Helm: the Helm installation reference, plus short pointers and a release note.

## Changes

- `reference/installation/helm_customization.mdx`: adds Gateway API to the customizable-resources list and a `gatewayAPI` block to the sample `values.yaml`.
- `networking/ingress-gateway/create-ingress-gateway.mdx`: the enablement step stays kubectl-only; a tip notes that Helm installs can set `gatewayAPI.enabled: true` in `values.yaml`, linking to the Helm installation reference.
- `networking/ingress-gateway/customize-ingress-gateway.mdx`: a tip notes that GatewayAPI spec fields can also be set at install time under `gatewayAPI:` in `values.yaml`, linking to the Helm installation reference.
- `release-notes/index.mdx`: adds an enhancement entry for enabling CIG during Helm installation.

## Dependency and timing

The `gatewayAPI.enabled` chart toggle ships via calico-private PR #12357 (targets the next Calico Enterprise release). This docs change should merge in lockstep with that release so the instructions match the shipped chart.


calico-enterprise/reference/installation/helm_customization.mdx
```diff
@@ -17,6 +17,7 @@ You can customize the following resources and settings during $[prodname] Helm-b
 - [Policy recommendation](api.mdx#policyrecommendationspec)
 - [Authentication](api.mdx#authenticationspec)
 - [Application layer](api.mdx#applicationlayerspec)
+- [Gateway API (Calico Ingress Gateway)](api.mdx#gatewayapi)
```
electricjesus: Nit: the rest of this list links to the `*Spec` anchors (e.g. `#applicationlayerspec`, `#policyrecommendationspec`). For consistency, link to `api.mdx#gatewayapispec` here. Both anchors resolve, so non-blocking.

calico-enterprise/networking/ingress-gateway/create-ingress-gateway.mdx
```diff
@@ -33,7 +36,31 @@ You need to do the following:
...
+   <TabItem label="Helm (at install time)" value="helm" default>
+
+   When you [install $[prodname] with Helm](../../getting-started/install-on-clusters/kubernetes/helm.mdx), set `gatewayAPI.enabled` to `true` in your `values.yaml`:
+
+   ```yaml title='values.yaml'
+   gatewayAPI:
+     enabled: true
+   ```
+
+   The Tigera Operator creates the `tigera-secure` `GatewayAPI` resource as part of the installation. To customize the gateway, set any [`GatewayAPI` spec field](../../reference/installation/api.mdx#gatewayapi) under `gatewayAPI:` — for example, to [define additional gateway classes](customize-ingress-gateway.mdx#configure-multiple-gateway-classes):
```
electricjesus: Two small things on this line:

1. The link text says "spec field" but points to `#gatewayapi` (the resource section). `#gatewayapispec` lands the reader on the spec fields and matches the rest of the docs.
2. The sentence offers "define additional gateway classes" as the example, but the YAML just below only sets the default `tigera-gateway-class` — so it doesn't actually demonstrate an *additional* class. Either reword to "define gateway classes", or add a second class to the example so it matches the prose.

calico-enterprise/networking/ingress-gateway/create-ingress-gateway.mdx
```diff
@@ -33,7 +36,31 @@ You need to do the following:
    allow traffic through the gateway by adding the `tigera-gateway` namespace to the list of excluded namespaces in the
    `namespaceSelector` field.
 
-1. To enable Gateway API support, create a `GatewayAPI` resource with the name `tigera-secure`:
+1. Enable Gateway API support by creating a `GatewayAPI` resource named `tigera-secure`. You can enable it at install time with Helm, or at any time with `kubectl`. Both methods create the same `tigera-secure` resource and the default `tigera-gateway-class`.
+
+   <Tabs groupId="cig-enable">
```
electricjesus: Worth previewing the rendered page — `<Tabs>` nested inside an ordered-list item can render oddly in Docusaurus (list numbering / indentation). Low risk, just confirm it looks right in the build.

electricjesus (commented):
Thanks Patrick — docs read well and the tigera-secure naming lines up with the chart. A few small inline nits below, all non-blocking. One housekeeping ask: mind stripping the "🤖 Generated with Claude Code" footer from the PR description? We keep AI attribution out of tigera repo PR bodies.

ctauchen (commented):
Thanks for this, @xiumozhan. 

I see from the tickets that we've added Helm support for CIG, which is the reason for this PR. But we don't tend to document Helm flows like this in our documentation. We have Helm install guides, tell people to take a look at values.yaml, and refer to the helm customization guide.

Without stong justifications, I'd be inclined to leave the CIG enablement process as-is (kubectl only). Update the customization page, add a release note, and perhaps add a note or link referring to this capability somewhere on the Create an ingress gateway page.

What are your thoughts here?

xiumozhan:
@ctauchen Agreed — I checked the other task pages and none of them document a Helm enablement flow inline, so the tabs were off-convention. Restructured in 5fc8ebc6 along the lines you suggested:

- `create-ingress-gateway.mdx` is back to the kubectl-only enablement step, with a short tip noting that Helm installs can set `gatewayAPI.enabled: true` in `values.yaml`, linking to the Helm installation reference.
- The Helm installation reference keeps the Gateway API list entry and the `gatewayAPI` block in the sample `values.yaml`.
- The tip on `customize-ingress-gateway.mdx` now just points at the Helm installation reference instead of restating the mechanism.
- Added a release-note entry under **New features and enhancements** for enabling CIG during Helm installation.

Also updated the PR description to match the new scope and stripped the Claude Code footer per @electricjesus's note.

calico-enterprise/reference/installation/helm_customization.mdx
xiumozhan: Fixed in 1d10de7a — the list entry now links to `api.mdx#gatewayapispec`, matching the sibling `*Spec` links.

calico-enterprise/networking/ingress-gateway/create-ingress-gateway.mdx
xiumozhan: Addressed in 1d10de7a (anchor changed to `#gatewayapispec`, second class added to the example). This paragraph has since been removed entirely in 5fc8ebc6 — per Chris's review the enablement step is kubectl-only again, with a tip linking to the Helm installation reference.

calico-enterprise/networking/ingress-gateway/create-ingress-gateway.mdx
xiumozhan: Moot as of 5fc8ebc6 — the tabs are removed and the step is plain kubectl again, so there's no nested-Tabs rendering to verify.

PR #2800, opened 2026-06-25
[CORE-12848] Document the OpenStack Neutron-etcd resync process

nelljerram:
Add a new networking/openstack/resync.mdx page covering the resync behaviour as it now exists after the post-v3.32 changes: one-shot startup resync (gated by [calico] startup_resync), the on-demand calico-resync command, per-item INFO logging, and performance and concurrency characteristics. Also note the removal of the old periodic resync and deprecation of resync_interval_secs and resync_max_interval_secs.

Cross-reference the new page from configuration.mdx and add it to the calico sidebar.


calico/networking/openstack/resync.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+## When resync runs
+
+There are two ways a resync happens:
+
+- **Startup resync** &mdash; a one-shot resync that runs automatically when the
+  Neutron server starts or is restarted. This is the normal, hands-off
+  mechanism, and is enabled by default.
+
+- **On-demand resync** &mdash; run explicitly by an operator using the
```
coutinhop: nit since it can be inferred from context: I'd prefer to use the term `user` or `administrator` to avoid possible confusion with automated operators such as the Tigera operator, but don't feel too strongly about it in this context

calico/networking/openstack/resync.mdx
```diff
@@ -0,0 +1,244 @@
...
+  only one of those performs startup resyncs; in this case you would configure
+  `always` on the chosen control node and `never` on the others.
+
+:::note
+
+Unlike the driver's other background jobs (etcd compaction, agent-status
+watching, and endpoint-status watching), which elect a single active worker
+dynamically through an etcd key so that the work fails over cleanly, the
+startup resync is governed by this static `startup_resync` config switch.  It
+runs once rather than continuously, so operator-level control over whether and
```
coutinhop: again, maybe change `operator` to `administrator`?

coutinhop (approved):

calico/networking/openstack/configuration.mdx
```diff
@@ -59,3 +59,20 @@ When specified, the value of `openstack_region` must be a string of lower case a
...
+[Resync between Neutron and etcd](resync.mdx) for details.
+
+| Setting        | Default Value | Meaning                                                                                  |
+| -------------- | ------------- | ---------------------------------------------------------------------------------------- |
+| startup_resync | always        | Whether to run a full Neutron DB &rarr; etcd resync when Neutron server starts. Set to `never` to suppress it. |
+
+:::note
+
+The `resync_interval_secs` and `resync_max_interval_secs` options that
+configured the old periodic resync are deprecated and have no effect.  See
```
ctauchen:
```suggestion
configured the old periodic resync have been removed.  See
```
In our usage, 'deprecated' suggests a thing still functions, though it's scheduled for removal.

calico/networking/openstack/resync.mdx
```diff
@@ -0,0 +1,244 @@
...
+deleting anything in etcd that no longer corresponds to Neutron.  Its purpose
+is to recover from the rare cases where the incremental path can leave etcd out
+of date, for example if the Neutron server crashed part way through handling a
+request, or if etcd was briefly unreachable when an update happened.
+
+## When resync runs
+
+There are two ways a resync happens:
+
+- **Startup resync** &mdash; a one-shot resync that runs automatically when the
```
ctauchen:
```suggestion
- **Startup resync**:  a one-shot resync that runs automatically when the
```
Colons are preferred in this context.

calico/networking/openstack/resync.mdx
```diff
@@ -0,0 +1,244 @@
...
+  Neutron server starts or is restarted. This is the normal, hands-off
+  mechanism, and is enabled by default.
+
+- **On-demand resync** &mdash; run explicitly by an operator using the
+  [`calico-resync`](#the-calico-resync-command) command, optionally scoped to
+  particular networks, subnets, ports, or security groups.
+
+:::note
+
+In releases up to and including v3.32, the driver also ran a _periodic_ resync,
```
ctauchen, line 39: You mention 3.32 a few times. Will this doc be added to 3.33?

ctauchen (commented):
A few nits, otherwise LGTM.

calico/networking/openstack/configuration.mdx
nelljerram: The problem is that they haven't been removed.  I think "deprecated and have no effect" is the accurate description here.  More broadly, I think "deprecated" can include both "deprecated and has no effect" and "deprecated and we encourage a new approach, but this way still works too".  I feel pretty sure we have prior examples of both of those.

calico/networking/openstack/resync.mdx
nelljerram: Yes, it should.  But that will happen automatically won't it, when the 3.33 docs are cut?

calico/networking/openstack/configuration.mdx
ctauchen: The functionality has been removed, despite whatever remnants you are leaving in the code. We can't use 'deprecated' here because they don't work anymore.

You're right that general usage of the word 'deprecated' is squishy. If you know of examples in our docs, by all means let me know and I'll set them right.

calico/networking/openstack/resync.mdx
ctauchen: Yes, of course. My mind must have jumped a release!

calico/networking/openstack/configuration.mdx
nelljerram: Fair enough, I've pushed a commit to stop saying "deprecated" for these options.

PR #2802, opened 2026-06-26
docs(gateway): fix default-deny step for namespaced mode

electricjesus:
Product Version(s):
Calico Enterprise 3.24 and Calico Open Source v3.33, the versions where the gateway proxy runs in the Gateway's own namespace. Only the unversioned pages are touched. Older snapshots predate namespaced mode and are left alone.

Issue:
https://github.com/tigera/operator/pull/4970

Link to docs preview:
- Calico Open Source: https://deploy-preview-2802--calico-docs-preview-next.netlify.app/calico/next/networking/ingress-gateway/create-ingress-gateway
- Calico Enterprise: https://deploy-preview-2802--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/ingress-gateway/create-ingress-gateway

Both links point at the `calico-docs-preview-next` build. The `tigera` preview still serves the old text on the `next` routes, so it is not the one to read for this PR.

SME review:

DOCS review:

Additional information:

This PR started out as a fix for the old step that told users to exclude the `tigera-gateway` namespace. #2873 already fixed that, so this is now a smaller follow-up on top of it.

#2873 added a reference NetworkPolicy that users apply in each Gateway namespace. It allows DNS and the hop to the gateway controller. With https://github.com/tigera/operator/pull/4970 the operator renders `calico-system.envoy-gateway-proxy`, which already allows those two things, plus inbound traffic to the proxy. So the reference policy would ask users to write rules the operator now owns.

What the operator's policy leaves alone on purpose is the backend hop. Its egress rules end in a `Pass`, which hands the decision to the user's own tier so they stay in control of which backends the proxy can reach. Under default deny that hop is denied, and the gateway returns `503` after Envoy's upstream connect timeout.

So the step now covers only the user's part:

- Proxy egress to the backend, and backend ingress from the proxy. Both are needed, because default deny applies to the backend pod too. The earlier version of this page only showed the egress half.
- On Enterprise, egress to the Kubernetes API server. The proxy pod runs a log collector that talks to it, and the operator's policy does not cover that hop.

The example selects proxies by `k8s-app == "calico-gateway-api-proxy"`, the Calico-owned label that #4970 stamps on the proxy pod template. The page used to use Envoy Gateway's own `app.kubernetes.io/name == "envoy"`.

Both policies go in the `default` tier at `order: 10`, so they sit after `calico-system` and take effect before a default-deny.

This should merge in the same release as https://github.com/tigera/operator/pull/4970. The text describes what the operator does once that lands.

Merge checklist:
- [ ] Test have passed

### DIFF
--- a/calico-enterprise/networking/ingress-gateway/create-ingress-gateway.mdx
+++ b/calico-enterprise/networking/ingress-gateway/create-ingress-gateway.mdx
@@ -30,8 +30,11 @@ You need to do the following:
 ## Create an ingress gateway
 
 1. If you are using a [global default deny policy](../../network-policy/beginners/kubernetes-default-deny.mdx),
-   allow traffic through the gateway by adding the `tigera-gateway` namespace to the list of excluded namespaces in the
-   `namespaceSelector` field.
+   you do not need to allow the gateway proxy's own traffic. The Tigera Operator allows it through a
+   policy in the `calico-system` tier. You do need to allow the gateway proxy to reach your backend
+   workloads. The proxy pods carry the label `gateway.envoyproxy.io/owning-gateway-name`, so you can
+   select them in an egress rule. Without that rule, the gateway is reachable but returns `503` because
+   the proxy cannot connect to the backend.
 
 1. To enable Gateway API support, create a `GatewayAPI` resource with the name `tigera-secure`:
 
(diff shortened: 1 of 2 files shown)


ctauchen:
Waiting on operator change and testing before review.

ctauchen:
@electricjesus Are you still pursuing this? The code PR looks stale.

PR #2805, opened 2026-06-26
Document v3 CRD mode as the default for new installs

caseydavenport:
Makes v3 CRD mode the documented default for new Calico installs, matching the operator change in https://github.com/tigera/operator/pull/4973. New installs now get `projectcalico.org/v3` served directly by CRDs with no aggregation API server. The v1 / aggregation-API-server path stays documented as an opt-out. Existing and upgraded clusters are unaffected - the operator picks the mode from the CRDs present and never flips an existing cluster.

What's here:
- Reframed the Native v3 CRDs pages (Open Source + Enterprise) as the default rather than a tech preview, and fixed the Kubernetes floor to 1.32+ (the old 1.34+ and per-version v1beta1/v1alpha1 claims were wrong).
- Flipped the fresh-install guides for both products to apply the v3 CRD manifest (`v3_projectcalico_org.yaml`). The operator and manifest installs that stay on the aggregation API server got a v3 opt-out note instead.
- Added "your CRD mode is preserved" notes to the upgrade guides for both products, with the v3 manifest called out as the alternative for clusters already on native v3.
- Release notes for both products.

Note on Enterprise: this assumes `v3_projectcalico_org.yaml` ships at the published EE manifests location. The release tooling already builds and publishes it (hashreleases and tagged releases both include it), so it should line up with real releases. The one gap is the `ee/master/` dev location, which is stale and hasn't picked up the granular bundles - flagged to the DE team separately.


calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
```diff
@@ -95,8 +95,14 @@ To install a standard $[prodname] cluster with Helm:
 
 1. Install the necessary custom resource definitions.
 
+   :::note[What changed]
```
ctauchen: We should consider removing the note here and on the other installation guides. If someone performs an installation and notices the missing api server, they'll find plenty of information in your v3 CRDs pages and the release notes.

One reason to keep the note, or something similar, is to make things easier for people who really need the api server. Is this a significant group? From what I read, there is little reason that anyone would prefer that option.

Best path is to present one option in the main guide and provide the less common alternative in another location.

calico-enterprise/getting-started/upgrading/upgrading-calico-to-calico-enterprise/upgrade-to-tsee/helm.mdx
```diff
@@ -45,6 +45,12 @@ The following steps assume the Calico deployment is installed on `tigera-operato
 
 1. Install the $[prodname] custom resource definitions.
 
+   :::note[CRD mode]
```
ctauchen: If we're maintaining both modes, then a more user-friendly path might look like this:

Prerequisite: You know what CRD mode you've got (Is there a simple check?)

1. Install the custom resource definitions for your cluster's mode:

TAB 1 / TAB 2

code code code

calico-enterprise/operations/native-v3-crds.mdx
```diff
@@ -181,6 +109,34 @@ In API server mode, tier RBAC is enforced for all operations (create, update, de
...
+
+Pull and install the v1 CRD chart before the Tigera Operator:
+
+```bash
+helm pull tigera-ee/crd.projectcalico.org.v1 --version $[releaseTitle]$[helmPre]
+helm template calico-crds crd.projectcalico.org.v1-$[chart_version_name].tgz | kubectl apply --server-side -f -
+```
+
+</TabItem>
+<TabItem label="Manifest install" value="apiserver-manifest">
```
ctauchen: We don't support manifest installations in Calico Enterprise. Is there something special about this case?

ctauchen: Perhaps this is just Tigera Operator and Helm installation types?

calico-enterprise/release-notes/index.mdx
```diff
@@ -18,6 +18,8 @@ This version of Calico Enterprise is based on [Calico Open Source $[openSourceVe
 
 ## New features and enhancements
 
+- New installs default to v3 CRD mode: `projectcalico.org/v3` resources are served directly by CRDs, with no aggregation API server. Existing and upgraded clusters are unaffected. To install with the aggregation API server instead, apply the v1 CRDs before installing the operator. v3 CRD mode requires Kubernetes 1.32 or later. See [Native v3 CRDs](../operations/native-v3-crds.mdx).
```
ctauchen, line 22: I'll do more closer to release, but I suspect it would be good to add more here about how this works for upgrades.

Would also be good to mention if we're deprecating the api server. Is that still the plan?

ctauchen (commented):
A few suggestions for you. Thanks!

calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
caseydavenport: Yeah, agreed. I pulled the note from all the install guides (OSS and Enterprise). The main guide just documents the v3 install now, and the API server opt-out lives on its own page - Native v3 CRDs for Enterprise, the API server page for Open Source.

calico-enterprise/getting-started/upgrading/upgrading-calico-to-calico-enterprise/upgrade-to-tsee/helm.mdx
caseydavenport: Good idea, restructured both upgrade guides this way. There is a simple check: `kubectl api-versions | grep crd.projectcalico.org` (output means the cluster is on the aggregation API server, no output means native v3). Added that as a step, then tabs for each mode.

calico-enterprise/operations/native-v3-crds.mdx
caseydavenport: Good catch, you're right. I relabeled to Helm and Tigera Operator and dropped the "manifest install" framing on the Enterprise page, both the install pointer near the top and the API server mode tab lower down.

calico-enterprise/release-notes/index.mdx
caseydavenport: Yep, still the plan. The aggregation API server is deprecated and will be removed in a future release (still supported for now). I added a deprecation entry to the release notes and the native v3 CRD pages, both pointing at the migration guide. Left the rest of the upgrade detail to you for closer to release.

ctauchen:
LGTM. Build failure is bad links that have since been fixed. Merge when ready.

caseydavenport:
Going to hold off on merging this one just yet until we get all of the testing done. Thanks for the review!

PR #2851, opened 2026-07-14
[CORE-13122] OpenStack: document the one-shot policy name migration for etcd clusters

nelljerram:
Calico v3.32 changed how policies in the default tier are stored in the etcd datastore.  Kubernetes clusters migrate existing data automatically via kube-controllers, but OpenStack clusters do not run kube-controllers as a service, so the migration must be run once by hand after upgrading from a version earlier than v3.32
(projectcalico/calico#13221).

Add the procedure to the OpenStack upgrade page: symptoms if skipped, preconditions (all nodes upgraded first; duplicate-pair cleanup), the docker invocation including TLS and username/password auth variants, completion output, idempotency, and verification.

The version-3.32 copy differs in using the standalone calico/kube-controllers image (the combined calico/calico image is v3.33+) and in noting that the mode requires v3.32.2 or later.

### DIFF
--- a/calico/operations/upgrading/openstack-upgrade.mdx
+++ b/calico/operations/upgrading/openstack-upgrade.mdx
@@ -209,6 +209,10 @@ following ways, until each one is migrated or recreated:
   or later before you run the migration.  On a Kubernetes cluster this
   ordering is verified automatically; on OpenStack it is your responsibility.
 
+- **Install and configure `calicoctl`.**  You need to be able to run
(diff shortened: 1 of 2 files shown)


ctauchen (approved):
LGTM

ctauchen:
To be published with OSS 3.32.2 patch release.

ctauchen:
Changes added to https://github.com/tigera/docs/pull/2981

PR #2877, opened 2026-07-23
Add reference docs for calicoctl datastore migrate-policy-names

caseydavenport:
Adds the v3.32 reference page for the new `calicoctl datastore migrate-policy-names` command (projectcalico/calico#13257).

The command rewrites pre-v3.32 policy names in an etcdv3 datastore so the stored name matches the v3 resource name. You only need it on an etcdv3 datastore (for example OpenStack) that was upgraded in place from before v3.32 - a Kubernetes datastore is migrated automatically by kube-controllers.

This is the reference-doc companion to Nell's OpenStack upgrade procedure in #2851, and like that PR it's aimed at the v3.32.2 release. OSS only for now, since the command hasn't merged to Calico Enterprise yet.

The master (Next) version of this page moved to the calicoctl cobra PR (#2878), which can merge now. This PR carries only the 3.32 page, so it stays gated on v3.32.2.

Related: CORE-13170

```release-note
None
```


nelljerram (approved):
LGTM.  But if the plan is to merge this when 3.32.2 is released, couldn't we also immediately populate Enterprise Next and Cloud Next at that time, and hence include those changes in this PR?

Obviously will need to get @ctauchen 's review also.

caseydavenport:
@nelljerram I left it out of those docs given those products don't include OpenStack or etcdv3 datastore support

nelljerram:
> @nelljerram I left it out of those docs given those products don't include OpenStack or etcdv3 datastore support

That makes perfect sense.  But the PR description makes it sound like it was for a different reason related to code merge timing: "OSS only for now, since the command hasn't merged to Calico Enterprise yet."

ctauchen:
LGTM, will hold and publish with OSS 3.32.2

ctauchen:
Changes added to https://github.com/tigera/docs/pull/2981.

PR #2878, opened 2026-07-23
Update calicoctl reference docs for the cobra migration

caseydavenport:
calicoctl moved from docopt to cobra (projectcalico/calico#12624), so the standalone binary now prints stock cobra `--help` output. The reference pages still showed the old docopt-style help blocks, so this swaps in the real cobra output (captured by building calicoctl from master) and converts the docopt-style flag sub-sections to cobra syntax to match.

The help blocks also include the per-command descriptions and usage examples added in projectcalico/calico#13302. This should land after that merges (and after calicoctl is rebuilt from master), so the captured output stays in sync.

Also removes the `convert` command docs - that command was removed from the CLI, so it's gone from both the Next docs and 3.32.

The cobra migration is master-only (3.32 still uses docopt), so the help-block updates here only touch the Next docs. The convert removal applies to both.

This also adds the Next `datastore migrate-policy-names` reference page in cobra form. The 3.32 version of that page lives in #2877, which is gated on the v3.32.2 patch release - nothing here depends on it, so this can merge independently.

CORE-13178

```release-note
None
```


ctauchen:
Build fails because of a broken link. Copilot flagged this one I think:

5:45:58 PM:   - Broken link on source page path = /calico/latest/reference/calicoctl/overview:
5:45:58 PM:      -> linking to /calico/latest/reference/calicoctl/convert

Assuming that fixes the build, I'm happy with the rest. LGTM and merge when ready.

ctauchen (approved):

PR #2882, opened 2026-07-27
Fix Enterprise CRD manifest references for native v3 CRD mode

caseydavenport:
Product Version(s): Calico Enterprise master

#2805 landed most of what this PR was doing. What's left are the manifest names and hosts it didn't touch.

`operator-crds.yaml` is a legacy alias for `v1_crd_projectcalico_org.yaml`. The two are byte-identical in calico-private, and the open source docs already use the v1 name. Renamed in the three remaining spots: the aggregation API server tab of both Helm upgrade pages, and the API server mode section of the native v3 CRDs page. Left the archive page and the ingress gateway canary tutorial alone, since those point at older artifacts that really are named `operator-crds.yaml`.

The managed cloud install includes (AKS, EKS, GKE, Generic, OpenShift) get the same rename. They still install the v1 bundle, which is now inconsistent with v3 being the default for new installs, but that's a behavior change rather than a naming fix so I've left it out of this PR.

The Enterprise CRD migration guide was copied from open source and pointed at `$[manifestsUrl]` (docs.tigera.io) rather than `$[filesUrl]` (downloads.tigera.io/ee), and claimed v3.32 as the version floor.

Not fixed here, filed separately as https://tigera.atlassian.net/browse/CORE-13214: `migration.projectcalico.org_datastoremigrations.yaml` is not published in any release artifact, so step 2 of the migration guide is a 404 in both products. That needs a build change, not a docs change.

Link to docs preview:

SME review:

DOCS review:

### DIFF
--- a/calico-enterprise/_includes/components/InstallAKS.js
+++ b/calico-enterprise/_includes/components/InstallAKS.js
@@ -20,7 +20,7 @@ export default function InstallAKS(props) {
       <ol>
         <li>
           <p>Install the Tigera Operator and custom resource definitions.</p>
-          <CodeBlock>kubectl create -f {filesUrl}/manifests/operator-crds.yaml</CodeBlock>
+          <CodeBlock>kubectl create -f {filesUrl}/manifests/v1_crd_projectcalico_org.yaml</CodeBlock>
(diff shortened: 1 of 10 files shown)


ctauchen (approved):

ctauchen:
LGTM, but needs rebase and merge conflict fixing.

PR #2904, opened 2026-08-04
Document boot-time network protection for non-cluster hosts

hjiawei:
Add a page for the non-cluster host boot-time network protection landing in Calico Enterprise v3.24.0-3.0 (EV-6755): what the boot ruleset allows, how to check it, how to change the allowed ports, how to opt out, and how to recover a locked-down or stalled host.

The feature is on by default wherever calico-node.service is enabled, so the install page gets a note at the enable step and the troubleshooting page gets the two symptoms an operator is most likely to hit.

Also correct the certificate renewal entry: calico-noncluster-host-init is now the "calico component node noncluster-host-init" subcommand rather than a standalone binary.

Product Version(s):

Calico Enterprise v3.24 EP3.

Issue:

Link to docs preview:

SME review:

DOCS review:

Additional information:

Merge checklist:

- [ ] Test have passed

### DIFF
--- a/calico-enterprise/getting-started/bare-metal/boot-protection.mdx
+++ b/calico-enterprise/getting-started/bare-metal/boot-protection.mdx
@@ -1,0 +1,186 @@
+---
+description: Protect a non-cluster host from the moment its network stack starts until Calico Enterprise applies network policy.
+---
+
+# Boot-time network protection
+
+## Big picture
+
+Block inbound traffic to a non-cluster host from the start of boot until $[prodname] applies network policy.
+
+## Value
+
+On a non-cluster host, $[prodname] cannot enforce policy right away.
+The Calico Node service (the agent) must start, connect to Typha, and sync the datastore first.
+Until it does, no policy is in force and nothing blocks inbound traffic.
+
+On a healthy boot this takes a few seconds, but there is no time limit on it.
+If Typha is unreachable, if the certificate cannot be renewed, or if the agent keeps restarting, the host stays unprotected until you fix the problem.
+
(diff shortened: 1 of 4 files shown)


calico-enterprise/getting-started/bare-metal/boot-protection.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+To confirm that the ruleset was applied early enough during the last boot, check that it came before the network was set up:
+
+```bash
+journalctl -b -o short-monotonic | grep -E "boot-protection|Preparation for Network"
+```
+
+:::note
+
+After you upgrade the `calico-node` package on an existing host, protection starts at the next restart of `calico-node.service`, not only at the next reboot.
```
alexh-tigera: Is this note needed? This is slightly confusing because earlier in the docs it says that

>  Protection covers boot only.
> If the agent restarts without a reboot, its rules stay in the kernel and keep enforcing, so there is no gap to cover.

alexh-tigera (approved):
Other than a minor nitpick, LGTM

calico-enterprise/getting-started/bare-metal/boot-protection.mdx
hjiawei: Agreed — removed. The section already says protection is on by default with nothing to enable, and the upgrade timing detail belongs in the release notes.

hjiawei:
This change is targeting Calico Enterprise v3.24 EP3 release.

ctauchen (approved):
LGTM, thanks!

PR #2905, opened 2026-08-04
DOCS-2992: Add a pre-upgrade step to delete the Compliance resource

caseydavenport:
Neither pass of the compliance removal (#2896 and the one before it) added an upgrade step, so anyone upgrading keeps a Compliance resource that nothing manages any more. This adds a pre-upgrade step to the Enterprise operator, Helm, and OpenShift upgrade pages: delete the resource, and remove it from any GitOps tooling so it does not come back. Deleting it before the upgrade is what lets Kubernetes garbage collect the compliance workloads and RBAC that the resource owns, while the operator still has a compliance controller around to clean up its status entry.

Applied to Calico Enterprise next only. The code removal went to master, and compliance still ships in the v3.24-1 and v3.24-2 release branches, so the versioned docs for those releases should keep describing it as present.

Jira: https://tigera.atlassian.net/browse/DOCS-2992


ctauchen:
Version-specific upgrade notes like this might fit better in the release notes, particularly when it's such a small procedure. If you don't have objections, I can hold this and include this there until the versions with compliance are EOL.

caseydavenport:
SGTM!

ctauchen:
Merging now as is, will fold into general pre-upgrade work in another PR.

PR #2907, opened 2026-08-06
Update the CRD migration docs for DatastoreMigration v1

caseydavenport:
Product Version(s): Calico v3.33 (`calico/`), Calico Enterprise next (`calico-enterprise/`)

Issue: https://tigera.atlassian.net/browse/CORE-12573

Link to docs preview:

The `DatastoreMigration` CRD is promoted from `migration.projectcalico.org/v1beta1` to `v1` in v3.33. The v3.33 CRD serves both versions, with `v1beta1` deprecated and `v1` as storage, so applying it over the v3.32 CRD is a plain update. Calico-side changes are in https://github.com/projectcalico/calico/pull/13383 and https://github.com/projectcalico/calico/pull/13449.

This supersedes #2888, which documented the earlier plan of deleting the v1beta1 CRD before applying the new one. That turned out to be dangerous: deleting the CRD cascades to the `DatastoreMigration` CR and runs its finalizer, which either deletes every `crd.projectcalico.org` CRD (CR in `Complete`) or rolls the migration back (any other phase).

Changes:

- Example CR uses `migration.projectcalico.org/v1`.
- New final cleanup step: delete the CRD, but only once the CR is gone. This is also what lets a future release drop `v1beta1` without a storage version migration.
- Warning that a `DatastoreMigration` CRD must never be deleted while a CR still exists.
- New upgrade section covering v3.32 to v3.33. The CRD is not part of an install, so the default answer is "nothing to do"; the table covers the cases where it isn't.
- Abort is only possible up to `Converged`, and there is nothing to abort back to on a cluster that already migrated.

The same changes are mirrored on the unversioned Enterprise page, minus the upgrade section, since I don't know which Enterprise release picks this up.

SME review:

DOCS review:

### DIFF
--- a/calico-enterprise/operations/crd-migration.mdx
+++ b/calico-enterprise/operations/crd-migration.mdx
@@ -49,6 +49,14 @@ The locked window is typically short (seconds to a few minutes depending on clus
 - Cluster is currently running in API server mode (the aggregated API server is deployed)
 - **If using GitOps (ArgoCD, Flux):** pause sync before starting the migration. These tools may interfere with the API group switchover. You'll update your manifests to use `projectcalico.org/v3` after migration completes.
 
+:::warning
+
+Never delete the `DatastoreMigration` CRD while a `DatastoreMigration` CR still exists.
+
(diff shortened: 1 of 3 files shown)


ctauchen (commented):
LGTM

MichalFupso (approved):

PR #2908, opened 2026-08-07
Explain the two ways to upgrade CRDs with Helm

caseydavenport:
The `tigera-operator` chart no longer ships the CRDs, and the upgrade docs only showed the manual apply. That leaves users with `manageCRDs: true` in their values wondering whether the manual step is still needed, and users on GitOps tooling wondering whether they need a second HelmRelease. Came up in [Slack](https://calicousers.slack.com/archives/C0BCA117T/p1786105269352289).

Adds a short section to the Helm upgrade instructions covering both options and why we recommend applying the CRDs yourself: the new CRDs land before the new operator starts, so new configuration can be set ahead of the upgrade. Also adds the `helm template` form of the CRD apply next to the raw manifest URL, which is easier to wire into GitOps tooling.

Applied to Calico latest, Calico 3.32, and Calico Enterprise. The Enterprise version also notes that `manageCRDs` does not cover the Prometheus and ECK CRDs.

### DIFF
--- a/calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
+++ b/calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
@@ -62,6 +62,11 @@ These steps differ based on your cluster type. If you are unsure of your cluster
 
 :::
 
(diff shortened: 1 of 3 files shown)


electricjesus (approved):

ctauchen:
LGTM, feel free to merge.

PR #2930, opened 2026-08-18
PMREQ-821: Document UI access through Calico Ingress Gateway

vara2504:
## Description

Documents exposing the UIs through Calico Ingress Gateway, the feature added by
tigera/operator#5146 (Whisker) and #5032 (Manager).

- **Enterprise** — a "Calico Ingress Gateway" tab on the web console access page,
  beside ingress, load balancer, port forwarding and OpenShift routes, plus a row
  in the options table. Covers the `GatewayAPI` prerequisite, `spec.ingressGateway`
  with optional `gatewayNamespace` and `gatewayClassName`, finding the gateway
  address, and turning it off again. A note covers OIDC: `Authentication`'s
  `managerDomain` must match the hostname, and the identity provider needs the two
  callback URLs.
- **Open source** — the same for the `Whisker` CR, next to the existing
  port-forwarding instructions, including that Whisker has no authentication of its
  own and that Calico restricts access to the gateway's proxy pods while the field
  is set.


ctauchen:
Thanks @vara2504. I'll go ahead and merge this for testing. I may do a bit of reorganizing before the release. I'll tag you for review if I do.

PR #2917, opened 2026-08-13
DOCS-2966: Publish Calico Enterprise 3.22.7

ctauchen:
Starter PR for the Calico Enterprise 3.22.7 patch release. Sets the version strings in releases.json, variables.js, and the release notes. Tracked by https://tigera.atlassian.net/browse/DOCS-2966.

Still to do:

- Replace the TBD bug fixes with the real list. 
- Add known issues.
- Confirm the release date. This PR uses the August 13 ETA from the ticket.
- Run the autogen third-party version sync from release-calient-v3.22.
- Update the operator API docs if needed.

To discuss:

- The ECK bump from https://tigera.atlassian.net/browse/EV-6724 is not in this PR. It is a third-party pin, so the autogen sync picks it up. Worth watching, because the 3.22.6 known issues already tell customers the fix lands in 3.22.7.
- The date is under pressure. The ticket says the release is being held for this month's golang patch.

PR #2932, opened 2026-08-19
DOCS-2966: Correct the Envoy Gateway version and regenerate the 3.22 Installation API reference

ctauchen:
Follow-up to the 3.22.7 publishing PR #2917, from a tag-to-tag diff of calico-private v3.22.6 to v3.22.7 and operator v1.40.12 to v1.40.14 against the draft release notes.

Three changes.

First, the Installation API reference for 3.22 is regenerated against operator v1.40.14. The release notes announce the new spec.cni.specVersion field, but the reference did not document it. The regeneration adds the specVersion row to the CNISpec table and a CNISpecVersion enum section. It also resolves cross-reference links the generator previously left as plain text. This clears the "update the operator API docs if needed" item from the publishing PR description.

Second, the Envoy Gateway version is corrected from v1.8.2 to v1.8.3, applying the review suggestion on #2917. third_party/envoy-gateway/Makefile at tag v3.22.7 reads ENVOY_GATEWAY_VERSION = v1.8.3. The envoyVersion variable is corrected in the same commit, from 1.7.2 to 1.8.3. That variable renders on the customize an ingress gateway page as the Envoy Gateway version that custom configuration must be valid for, and it was stale for 3.22.6 as well, where the shipped version was v1.8.0.

Third, two version downgrades in the v3.22.0-1.0 entry of releases.json are reverted, addressing the review comments on that hunk. The downgrades were coreos-alertmanager v0.28.1 to v0.28.0 and eck-elasticsearch-operator 2.16.1 to 2.16.0. Both came from the release tooling's pinned version template, which lags what the images are built from. The third_party Makefiles at tag v3.22.0-1.0 read v0.28.1 and v2.16.1, so the original values were right.

On the source of truth question the review raised, the third_party Makefiles are authoritative and calico/_data/versions.yml is stale at release tags. At tag v3.22.6 that file records alertmanager v0.28.1 while the Makefile reads v0.32.1, and the release branch has since been corrected to v0.32.1. Every other historical correction in #2917 is confirmed against the Makefile at its tag and is kept.

Disposition of the remaining review comments on #2917:

- The TBD placeholder under Bug fixes is already replaced with the real list.
- The Dikastes suggestion is already applied.
- The suggestion to derive releaseTitle, filesUrl and chart_version_name from releases[0].title is not taken here. It changes the shared publish template for every version stream and does not belong in a patch release.
- The comment about the sentence "A fix is included in Calico Enterprise 3.22.7." sitting above the 3.22.7 section is a false positive. It belongs to the 3.22.6 known issue and forward-references correctly, and every entry ends with the same "To update an existing installation" footer.

Not addressed here, for a decision by the release owners:

- coreos-prometheus in the v3.22.5 entry reads v3.11.1 but the Makefile at tag v3.22.5 reads v3.11.2. This mismatch predates #2917, which deliberately left pre-existing mismatches alone, so it is reported rather than changed.
- The ECK bullet credits 3.4.1 with the transport CA rotation fix, which is in 3.4.0 upstream, and names only the destination version. The real move is 2.16.1 to 3.4.1, a major-line jump in a patch release. The jump also installs two new CRDs, autoopsagentpolicies.autoops.k8s.elastic.co and packageregistries.packageregistry.k8s.elastic.co, and adds ClusterRole rules without which the ECK operator restart-loops and Elasticsearch is never created.
- Two user-facing changes have no bullet. Windows flow logs reported zero packet and byte counts for flows whose VFP entry came from a policy match, fixed under CORE-12154. And calico-node now mounts securityfs read-only at /sys/kernel/security, under CORE-13107, which fixes kernel lockdown detection but will be rejected by hostPath allowlist policies until they are updated.
- The release note says the non-cluster host packages "disable nftables on RHEL 8". The commit message for that change describes it, but the diff does not implement it and NFTablesMode appears nowhere in the packaging at tag v3.22.7. Only the BPF half shipped. The rpm also gained a Requires on nftables, which can fail an install where that package is unavailable.

Netlify does not build a deploy preview for a pull request based on a publish branch, so there is no preview link for this PR. The changed pages are:

- /calico-enterprise/3.22/reference/installation/api, for the new specVersion field and the CNISpecVersion enum
- /calico-enterprise/3.22/release-notes/, for the Envoy Gateway version in the Ingress Gateway bullet
- /calico-enterprise/3.22/networking/ingress-gateway/customize-ingress-gateway, for the rendered Envoy Gateway version

Once this merges into publish/ce-3.22.7, the deploy preview on #2917 covers all three.


calico-enterprise_versioned_docs/version-3.22-2/release-notes/index.mdx
```diff
@@ -425,7 +425,7 @@ August 19, 2026
 * Fixed Felix's periodic route resync not repairing Calico-owned routes modified in place by another process, unnecessary reprogramming of unchanged IPv6 multi-path routes, and a spurious conntrack cleanup when removing an IPAM block route.
 * Fixed Calico advertising a Service IP over BGP from a node whose only local endpoint was not Ready, black-holing traffic for services using `externalTrafficPolicy: Local`.
 * Fixed restarting the `calico-early` container while calico-node was running causing a transient BGP flap that briefly withdrew pod and egress gateway routes.
-* Fixed the Calico Ingress Gateway control plane crash-looping on clusters whose Gateway API CRD set omits `ListenerSet`, `TLSRoute`, or `BackendTLSPolicy`, such as OpenShift. The bundled Envoy Gateway is updated to v1.8.2.
+* Fixed the Calico Ingress Gateway control plane crash-looping on clusters whose Gateway API CRD set omits `ListenerSet`, `TLSRoute`, or `BackendTLSPolicy`, such as OpenShift. The bundled Envoy Gateway is updated to v1.8.3.
```
ctauchen: FYI @rene-dekker This is what I'm seeing in the envoy-gateway makefile.

PR #2944, opened 2026-08-21
DOCS-2997: Write the L2 bridge networking concept page

ctauchen:
Fills in the concept page for L2 bridge networking. Second in the DOCS-2997 sequence, building on the structure already merged.

The page explains why the feature exists, what a Network resource models, and how frames move between a workload, the bridge, and your fabric. It covers the choice between a Calico-managed bridge and a bridge you prepare yourself, and why that choice depends on whether Calico can have an uplink of its own rather than on how many network cards a node has.

Three sections carry points that came out of the engineering review:

- Policy enforcement does not change on an L2 network. The page says so early, because it is the first thing a security reviewer asks.
- VMs and pods are configured differently, so the page establishes that before any procedure depends on it.
- The reason Calico refuses to build a bridge on a shared uplink is explained in terms of who owns the node's IP address.

The page is explanation only. It contains no configuration to apply and no numbered steps. Those live in the two setup guides.

Two things for reviewers to weigh in on:

- The mode explanation and the host address placement would both land better as diagrams. They are prose for now. Worth deciding whether to add figures before this ships.
- The Kubernetes integration limits are summarised here and stated in full on the support page. Check that the split reads sensibly.

Changed page: https://deploy-preview-2944--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/l2-bridge/about-l2-bridge

### DIFF
--- a/calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
+++ b/calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
@@ -125,8 +125,13 @@ It is a question of whether $[prodname] can have an uplink of its own.
 If there is an interface on the node that $[prodname] can take over — one that is not carrying the node's own IP address — then $[prodname] creates and owns the bridge.
 
 This is the simpler option wherever it applies.
-There is nothing for you to configure correctly on the host, nothing that a reboot can undo, and your node's own address never moves.
+There is nothing for you to configure correctly on the host, and your node's own address never moves.
 $[prodname] creates the bridge, enslaves the trunk, sets every option it needs, and cleans up after itself.
+
+There is also nothing to keep in step.
+With a bridge $[prodname] owns, the VLAN list on the trunk cannot drift from the `Network`, because nothing else is managing it.
+On a bridge you built, those are two sources of truth: if they diverge, the next time your host configuration is reapplied it can strip a VLAN $[prodname] added, and traffic on that VLAN stops without anything reporting it.
+See [Prepare an existing bridge](byo-bridge.mdx).
 
 ### The uplink is shared with the host
 
--- a/calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
+++ b/calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
@@ -6,92 +6,257 @@ description: How Calico Enterprise puts VMs and pods on your existing VLANs, whi
 
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
 ## Why L2 networking in $[prodname]
 
-Why the feature exists, and what $[prodname] adds that a hand-built bridge cannot.
+A virtual machine moving onto Kubernetes usually arrives with an identity that other systems already depend on.
+Its IP address is in DNS.
+Firewall rules name it.
+License servers are bound to it, monitoring dashboards chart it, and other applications hold its address in their own configuration.
+
+Re-addressing that machine means changing all of those things too.
+For a handful of VMs that is an afternoon.
+For an estate of thousands it is the largest single cost of moving to Kubernetes, and it is often the reason the move stalls.
+
+L2 bridge networking removes that cost.
+A VM keeps its IP address, its MAC address, and its place on the same broadcast domain as the machines it already talks to.
+The network around it does not have to change.
+
+You could build something similar yourself with a Linux bridge and a VLAN trunk.
+What you cannot get that way is policy.
+Because the workload's interface is a $[prodname] interface, it gets network policy and flow visibility on that L2 link, exactly as a pod on the $[prodname] pod network does.
+
+It is worth being straightforward about the tension here.
+$[prodname] was designed to separate security from connectivity and to scale by routing at layer 3.
+L2 bridge networking runs against that grain deliberately: it adapts $[prodname] to the network you already have, rather than asking you to change the network to suit $[prodname].
 
 ## What it is
 
-L2 bridge networking is additive and opt-in. A cluster that creates no `Network` resources is unchanged.
+L2 bridge networking is additive and opt-in.
+A cluster that creates no `Network` resources behaves exactly as it did before: the default $[prodname] pod network, its IPAM, and its data plane are untouched.
+
+When you do create one, you get a logical L2 network that spans the nodes you select.
+Workloads attached to it sit on one of your existing VLANs, reachable from anything else on that VLAN — inside the cluster or outside it.
 
 ## VMs and pods are not the same here
 
-The feature is aimed at VMs, and VMs are configured differently from pods. This shapes every procedure that follows.
+This feature is aimed at VMs, and a VM is configured differently from a pod.
+The difference shapes every procedure in this section, so it is worth establishing early.
+
+For a **VM**, you declare the interface in the VM specification, alongside the VM's other devices.
+KubeVirt builds the launcher pod for you, and you do not edit its annotations by hand.
+
+For a **pod**, you annotate the pod directly with the Multus network you want.
+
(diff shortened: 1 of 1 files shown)


calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
```diff
@@ -6,92 +6,257 @@ description: How Calico Enterprise puts VMs and pods on your existing VLANs, whi
...
+A VM keeps its IP address, its MAC address, and its place on the same broadcast domain as the machines it already talks to.
+The network around it does not have to change.
+
+You could build something similar yourself with a Linux bridge and a VLAN trunk.
+What you cannot get that way is policy.
+Because the workload's interface is a $[prodname] interface, it gets network policy and flow visibility on that L2 link, exactly as a pod on the $[prodname] pod network does.
+
+It is worth being straightforward about the tension here.
+$[prodname] was designed to separate security from connectivity and to scale by routing at layer 3.
+L2 bridge networking runs against that grain deliberately: it adapts $[prodname] to the network you already have, rather than asking you to change the network to suit $[prodname].
```
fasaxc, line 45: Bit Claude-y?

calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
```diff
@@ -6,92 +6,257 @@ description: How Calico Enterprise puts VMs and pods on your existing VLANs, whi
...
 
 ## What it is
 
-L2 bridge networking is additive and opt-in. A cluster that creates no `Network` resources is unchanged.
+L2 bridge networking is additive and opt-in.
+A cluster that creates no `Network` resources behaves exactly as it did before: the default $[prodname] pod network, its IPAM, and its data plane are untouched.
+
+When you do create one, you get a logical L2 network that spans the nodes you select.
+Workloads attached to it sit on one of your existing VLANs, reachable from anything else on that VLAN — inside the cluster or outside it.
```
fasaxc: Section is called "What it is", but it doesn't answer that

calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
```diff
@@ -2,96 +2,271 @@
...
+Your node's IP address does not belong to $[prodname].
+It belongs to whatever configures the host's network — systemd-networkd, NetworkManager, or something your provisioning system drives.
+Moving that address onto a bridge is not a single operation: afterwards, the component that manages it has to be doing DHCP or static addressing on the *new* interface, or the address does not come back after a reboot.
+
+$[prodname] will not reconfigure or disable those components.
+If it tried and got it wrong, the result is a node that has lost its own network with nothing left to fix it from.
+
+So the boundary is ownership, not capability.
+On a bridge you built, $[prodname] still does everything else it would otherwise do: it attaches workloads, programs their policy, sets their VLAN membership, and locks their ports down.
+It simply does not touch the parts your host configuration owns.
```
lwr20: This boundary is not quite held today, on the BYO-bridge path.

Attaching L2 workloads lowers the bridge MTU to Calico's auto-detected value (1480), and because a Linux bridge takes the minimum MTU of its ports, the VLAN sub-device carrying the host's own address and default route drops with it. The operator had explicitly configured the trunk at 1500 and set no MTU on the bridge.

So Calico does change something the host configuration owns. Raised as CORE-13428 — flagging it here because this paragraph is the promise it contradicts.

lwr20 (commented):
Checked against `2026-08-20-v3-24-2-kettle` on OpenShift. The model and the eBPF-only rationale read accurately against the shipped behaviour.

calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
ctauchen: Agreed. Holding this for the editorial pass with the rest of the language cleanup, rather than in this PR.

lwr20:
Read this end to end against a live rig (RKE2 + Calico Enterprise v3.24 EP2, BPF dataplane, BYO bridge, plus an openSUSE Leap Micro node standing in for a customer's SLE Micro). I couldn't fault the page — so rather than silence, here is **which assertions I've now actually measured**, since the review note says assertions are assumed sound unless someone says otherwise.

**Verified on a live cluster:**

- *"Every packet in and out of the workload is still evaluated against your policy"* and *"None of this differs between the two bridge modes"* — confirmed. A deny-all-egress policy on an L2 workload took it from 0% to 100% packet loss, and removing it restored 0%. Also confirmed with a HostEndpoint on the bridge sub-interface: an L2 workload's traffic was subjected to the policy, and a control pod on a node without the HEP reached the same target fine.
- *"The interface is not even brought up until that is done"* — confirmed. 1200 samples across workload attachment, zero leak windows. Felix's log shows the ordering explicitly (`WEP's Network not yet ready; forcing admin-down` … `WEP's Network now ready; releasing admin-down`).
- *"It validates the source MAC and source IP address … validates the sender fields in ARP"* — confirmed. Source-IP spoofing, source-MAC spoofing, ARP poisoning and VLAN hopping were all blocked.
- *"Two workloads on the same VLAN on the same node are bridged locally. Their traffic never leaves the node"* — confirmed, and worth keeping prominent: it cost me a false negative. A test whose two endpoints are on one node cannot detect anything wrong with the trunk.
- *"Traffic that arrives on the trunk without a tag is handled by the native VLAN"* — confirmed. Setting `nativeVLAN` produced `100 PVID Egress Untagged` on the trunk, exactly as described.
- *"A node has one trunk in this release"* — consistent with everything I saw.

**One phrasing nuance.** In *"$[prodname] can have its own uplink"*:

> *"There is nothing for you to configure correctly on the host, nothing that a reboot can undo, and your node's own address never moves."*

All true, but a reader may take the middle clause as implying a BYO bridge *doesn't* survive a reboot. It does, when configured persistently — I rebooted a NetworkManager-based node three times and the bridge, its VLANs and the node's address all came back unaided, with Calico reprogramming its VLANs automatically.

What a reboot genuinely cannot undo in managed mode is something narrower and, on the evidence, more valuable: with a bridge Calico owns, **the trunk's VLAN list cannot drift from the `Network`**, because nothing else is managing it. On a BYO bridge those are two sources of truth, and if they diverge the next NetworkManager/nmstate reconcile silently strips Calico's VLAN — CORE-13475, reproduced on both RKE2 and OpenShift/RHCOS. Detail is on #2946.

Suggest tightening that clause to say what it means — something like *"nothing on the host that can drift out of step with the `Network`"* — rather than leaving "reboot" to carry it.

calico-enterprise/networking/l2-bridge/about-l2-bridge.mdx
ctauchen: Fixed, thanks, and it is a fair catch that this paragraph is the promise it contradicts. The ownership section now says explicitly that one thing crosses the line: workload interfaces are attached at the cluster-wide MTU, a bridge takes the lowest MTU of its ports, and a bridge you built can drop with it. It points at the BYO page for pinning it. CORE-13428 noted.

ctauchen:
Thanks Lance, and thanks for listing what you measured rather than staying silent — knowing which assertions are now tested is worth as much as the corrections.

Taken on the phrasing. "Nothing that a reboot can undo" is gone. You are right that it invited the wrong inference, and your reboot testing settles it: a persistently configured BYO bridge does come back unaided. The section now names what managed mode actually prevents, close to your wording: with a bridge Calico owns the VLAN list on the trunk cannot drift from the Network, because nothing else is managing it. It also says what happens on a BYO bridge when the two diverge, and links to 2946 where the detail lives.

I have also qualified the ownership boundary paragraph further up, following your MTU comment: attaching workloads can lower the MTU of a bridge you built, and of a VLAN device above it, so that promise is no longer unqualified.

CORE-13475 is now warned about on 2946 in two places, by the trunk example and as a verification round for later host changes.

ctauchen:

PR #2945, opened 2026-08-21
DOCS-2997: Write the L2 bridge support and limitations reference

ctauchen:
Fills in the support and limitations reference for L2 bridge networking. Third in the DOCS-2997 sequence.

The page is a lookup table rather than prose, so a reader can rule the feature in or out quickly. It records the data plane requirement, the kernel floor, the prerequisites, the one supported bridge topology, the addressing rules, and what is not in this release.

Several entries exist because they are easy to miss and expensive to discover later:

- The kernel floor is 5.18, which is higher than the floor for the eBPF data plane on its own.
- Installing Multus is not sufficient. The multi-interface mode also has to be set, and nothing reports the omission until a workload fails to get connectivity.
- Connect-time load balancing cannot be disabled for a single interface. This affects pods on an L2 network but not VMs.
- A flow that crosses two L2 segments on one node can be double-counted in flow logs.

Two entries need an engineering check before this ships. The rows for Kubernetes Services, NAT, Maglev, and nested BGP come from the design document rather than from tested behaviour, and this is an expensive page to be wrong on. Flagging rather than dropping them.

Changed page: https://deploy-preview-2945--calico-docs-preview-next.netlify.app/calico-enterprise/next/reference/l2-bridge-support

### DIFF
--- a/calico-enterprise/reference/l2-bridge-support.mdx
+++ b/calico-enterprise/reference/l2-bridge-support.mdx
@@ -6,36 +6,155 @@ description: What Calico Enterprise L2 bridge networking supports and does not s
 
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-Check this page before you design around L2 bridge networking. The constraints are narrower than the rest of $[prodname] networking.
+Read this page before you design around L2 bridge networking.
+Its requirements are narrower than the rest of $[prodname] networking, and several of them cannot be changed after a cluster is built.
 
 ## Status
 
+L2 bridge networking is a tech preview feature in $[prodname] 3.24.
+The API may change in response to feedback from tech preview users.
+
+There is nothing extra to install.
+The `Network` resource and the components that act on it ship with the $[prodname] operator.
+
 ## Data plane support
+
+The eBPF data plane is required, and it is the only data plane that will support this feature.
+
+| Data plane | Supported |
+| ---------- | --------- |
+| eBPF | Yes. Required. |
+| iptables | No, and not planned. |
+| nftables | No. Deferred to a later release. |
+| Windows HNS | No. Out of scope. |
+
+For why the feature is tied to eBPF, see [About L2 bridge networking](../networking/l2-bridge/about-l2-bridge.mdx).
 
 ## Platform requirements
 
+Linux nodes only, on a recent kernel.
+
+| Requirement | Value |
+| ----------- | ----- |
+| Operating system | Linux |
+| Kernel | 5.18 or later |
+
+The kernel requirement is specific to this feature.
+It is higher than the floor for the eBPF data plane on its own, because $[prodname] uses a bridge option that older kernels do not provide.
+
 ## Prerequisites
+
+All of these must be in place before you create a `Network`.
+
+- An active $[prodname] license.
+  Creating a `Network` resource is license-gated.
+  Existing `Network` resources keep working if a license lapses.
+- The eBPF data plane, enabled cluster-wide.
+- [Multus](../networking/configuring/multiple-networks.mdx), for attaching workloads to a network as an additional interface.
+  Multus is not shipped by $[prodname].
+- `multiInterfaceMode: Multus` set in the `calicoNetwork` section of the [Installation](installation/api.mdx) resource.
+  Installing Multus is not enough on its own, and nothing reports the omission until a workload fails to get connectivity.
 
 ## Bridge and topology support
 
+There is one supported topology: a VLAN-aware bridge with a trunk uplink.
+
+| Arrangement | Supported |
+| ----------- | --------- |
+| VLAN-aware bridge, trunk uplink | Yes |
+| VLAN sub-device as the bridge uplink | No |
+| One bridge per VLAN, each fed by a VLAN sub-device | No |
+| More than one trunk on a node | No |
+
+A bridge you prepare yourself must meet three requirements, which $[prodname] checks and will not correct for you: VLAN filtering enabled, an explicitly set MAC address that is not the trunk's own, and the 802.1Q tag protocol rather than 802.1ad.
+For the full contract, including what $[prodname] does not require, see the [Network resource](resources/network.mdx).
+For the procedure, see [Prepare an existing bridge](../networking/l2-bridge/byo-bridge.mdx).
+
 ## Addressing
+
+IPv4 only, and $[prodname] must be the source of the address.
+
+- IPv6 is not supported.
+  The data plane drops IPv6 traffic on L2 interfaces.
+- At most one address per address family, per interface.
+- Addresses cannot be added to or removed from an interface after it is set up.
+- An external DHCP server on the VLAN cannot assign workload addresses.
+  $[prodname] IPAM must be authoritative, because policy and reverse path filtering depend on $[prodname] knowing the address.
+- KubeVirt's own in-pod DHCP service is fully supported, and is the normal way a VM receives the address $[prodname] assigned to it.
 
 ## Traffic types
 
+Two categories of traffic are handled differently from the pod network.
+
+- Broadcast and multicast traffic is dropped, in both directions.
+- Reverse path filtering on pod traffic is strictly enforced.
+
 ## Kubernetes integration
+
+Several Kubernetes networking features do not extend to L2 interfaces.
+
+| Feature | On an L2 interface |
+| ------- | ------------------ |
+| Kubernetes Services | Not available |
+| NAT | Not available |
+| Maglev load balancing | Not available |
+| Nested BGP to local workloads | Not supported |
+| Service mesh | Not supported |
+
+Connect-time load balancing is a host-wide setting and cannot be disabled for one interface.
+This does not affect VMs, which resolve Services inside their own kernel.
+It does affect a pod on an L2 network: a cluster IP address can be resolved and translated on the way out of the pod's primary interface, then dropped by reverse path filtering.
 
 ## Primary and secondary interfaces
 
+An L2 network is normally an additional interface, but it can be a workload's primary interface instead.
+
+Setting it as the primary interface means the workload has no route to or from the Kubernetes network, unless your own network routes back into the cluster.
+For a pod that usually rules it out, because kubelet cannot reach the pod for readiness and liveness checks.
+For a VM behaving like a traditional server on a VLAN, it is often the intended configuration.
+
 ## Observability
+
+Network policy and flow logs work on L2 interfaces, which is the main reason to use this feature rather than a bridge you manage yourself.
+
+One caveat applies in this release.
+A flow that crosses two L2 segments on the same node, leaving to your router and returning, is recorded once per segment.
+Such flows can be double-counted in flow log totals, and a flow log record can close while the flow is still running.
+Flows that stay within one segment are unaffected.
 
 ## Sizing guidance
 
+Prefer one `Network` with several VLANs over several Networks with one VLAN each.
+Every `Network` adds per-node work, so consolidating VLANs into one Network costs less.
+
+Concrete scale limits have not been established for this release.
+Treat the tech preview as suitable for evaluation and pilot workloads rather than for sizing a production estate.
+
 ## Lifecycle
+
+Two operations to plan for, because neither is guarded.
+
+- Deleting a `Network` while workloads are still attached to it removes the bridge from under them.
+  Delete the workloads first.
+- A `Network` cannot be changed from one type to another.
+  An L2 network cannot become a VRF network, or the reverse.
+
+Downgrading $[prodname] after `Network` resources are in use is not supported.
 
 ## Not in this release
 
-Features that are planned or under consideration, with no committed timing.
+These are known gaps rather than permanent limitations.
+No timing is committed for any of them.
+
+- VXLAN interconnect between L2 segments, and to nodes outside the cluster.
+- eBPF-accelerated forwarding within the bridge.
+- Protection against deleting a `Network` that is still in use, and status reporting on the `Network` resource.
+- Broadcast and ARP traffic in flow logs.
+- $[prodname] acting as the layer 3 gateway for a new segment.
+- EVPN.
+  This is planned as separate work.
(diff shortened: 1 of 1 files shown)


calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
+| One bridge per VLAN, each fed by a VLAN sub-device | No |
+| More than one trunk on a node | No |
+
+A bridge you prepare yourself must meet three requirements, which $[prodname] checks and will not correct for you: VLAN filtering enabled, an explicitly set MAC address that is not the trunk's own, and the 802.1Q tag protocol rather than 802.1ad.
+For the full contract, including what $[prodname] does not require, see the [Network resource](resources/network.mdx).
+For the procedure, see [Prepare an existing bridge](../networking/l2-bridge/byo-bridge.mdx).
+
 ## Addressing
 
+IPv4 only, and $[prodname] must be the source of the address.
```
fasaxc: Slight subtlety:
- Kubevirt runs a DHCP server inside the pod, so the VM may still use DHCP to get its IP, but external DHCP is not supported once VM is moved into k8s
- VM may have had a DHCP address before being migrated in, but it's assigned in Calico IPAM after being moved in.

calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
+
+The kernel requirement is specific to this feature.
+It is higher than the floor for the eBPF data plane on its own, because $[prodname] uses a bridge option that older kernels do not provide.
+
 ## Prerequisites
 
+All of these must be in place before you create a `Network`.
+
+- An active $[prodname] license.
+  Creating a `Network` resource is license-gated.
```
fasaxc: Checking on this, Networks predates L2 (was added by VRF) and I'm not sure if it's true

fasaxc (dismissed):

lwr20 (commented):
Checked against `2026-08-20-v3-24-2-kettle` on OpenShift. Most of this page verified correct — `allowedUses: L2Workload`, `multiInterfaceMode: Multus`, `vlan_filtering 1`, `vlan_protocol 802.1Q` and the single-trunk constraint all match the shipped build.

calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
+For why the feature is tied to eBPF, see [About L2 bridge networking](../networking/l2-bridge/about-l2-bridge.mdx).
+
 ## Platform requirements
 
+Linux nodes only, on a recent kernel.
+
+| Requirement | Value |
+| ----------- | ----- |
+| Operating system | Linux |
+| Kernel | 5.18 or later |
```
lwr20: This excludes RHEL 9 / RHCOS, where the feature demonstrably works.

The validation cluster runs kernel `5.14.0-570.78.1.el9_6` and is fully functional. The real requirement is the bridge port `locked` flag (upstream 5.18), which Red Hat backports — `locked on` is set on every Calico access port there.

As written, an OpenShift admin checking `uname -r` concludes the feature is unsupported on the platform it was developed on. Suggest stating the capability, with a note that RHEL 9 / RHCOS satisfies it despite the version number.

calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
+- Addresses cannot be added to or removed from an interface after it is set up.
+- An external DHCP server on the VLAN cannot assign workload addresses.
+  $[prodname] IPAM must be authoritative, because policy and reverse path filtering depend on $[prodname] knowing the address.
+- KubeVirt's own in-pod DHCP service is fully supported, and is the normal way a VM receives the address $[prodname] assigned to it.
+
 ## Traffic types
 
+Two categories of traffic are handled differently from the pod network.
+
+- Broadcast and multicast traffic is dropped, in both directions.
```
lwr20, line 134: This reads as absolute, but ARP demonstrably works — it has to, for the feature to function at all.

`arping` from a peer gets unicast replies, and the page later lists "Broadcast and ARP traffic in flow logs" as not-in-release, implying ARP traffic exists. #2944 also says Calico validates ARP sender fields.

Worth saying explicitly whether ARP is excepted from this, and if so how.

calico-enterprise/reference/l2-bridge-support.mdx
lwr20: MTU is not mentioned anywhere in this page or the other five, and it is not neutral on an L2 network.

On the test cluster `br-l2` and every workload port run at 1480 while the trunk is 1500 — the operator auto-detects an encapsulation-adjusted MTU even though `L2Workload` pools force `ipipMode`/`vxlanMode: Never`. A VM keeping its IP and MAC lands on a 20-byte-smaller path than it had on the same VLAN, and peers on that VLAN are still at 1500.

Raised as CORE-13428, which also covers a host-networking side effect. This section is probably where the constraint belongs.

calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
+There is nothing extra to install.
+The `Network` resource and the components that act on it ship with the $[prodname] operator.
+
 ## Data plane support
 
+The eBPF data plane is required, and it is the only data plane that will support this feature.
+
+| Data plane | Supported |
+| ---------- | --------- |
+| eBPF | Yes. Required. |
```
lwr20, line 41: Worth noting: this requirement is documented but not enforced anywhere I can find.

The `Network` CRD carries five CEL validation rules (type immutability, name prefix, nativeVLAN range, VLAN range ordering) and none check the dataplane, so a `Network` applied on an iptables cluster is accepted by the API. Forklift validates non-BPF destinations; Calico itself does not.

I have not tested what happens next, so this is a question rather than a defect — but if the answer is "workloads attach without BPF anti-spoof" it matters a lot more than if it is "nothing happens".

calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
 
+All of these must be in place before you create a `Network`.
+
+- An active $[prodname] license.
+  Creating a `Network` resource is license-gated.
+  Existing `Network` resources keep working if a license lapses.
+- The eBPF data plane, enabled cluster-wide.
+- [Multus](../networking/configuring/multiple-networks.mdx), for attaching workloads to a network as an additional interface.
+  Multus is not shipped by $[prodname].
+- `multiInterfaceMode: Multus` set in the `calicoNetwork` section of the [Installation](installation/api.mdx) resource.
```
lwr20: Worth a prerequisite bullet here: **NADs that delegate to Calico must declare `cniVersion` 1.0.0 or lower.**

The plugin advertises `["0.1.0","0.2.0","0.3.0","0.3.1","0.4.0","1.0.0"]` (`CNI_COMMAND=VERSION /opt/cni/bin/calico`). A NAD declaring `1.1.0` fails CNI ADD with `incompatible CNI versions` and the pod never leaves `ContainerCreating`. That ceiling is deliberate and self-consistent — `Installation.spec.cni.specVersion` stops at `1.0.0` too — so this is a limit to document, not a defect.

It doesn't bite anyone following the docs (#2947's NAD uses `0.3.1`). It bites people hand-authoring a NAD, which is the normal path on RKE2 where Multus is installed by hand.

Tested on RKE2 v1.33.5+rke2r1, EE `2026-08-20-v3-24-2-kettle`, Multus v4.3.0. Claude-assisted, verified on a live cluster.

calico-enterprise/reference/l2-bridge-support.mdx
```diff
@@ -2,40 +2,169 @@
...
 ## Prerequisites
 
+All of these must be in place before you create a `Network`.
+
+- An active $[prodname] license.
+  Creating a `Network` resource is license-gated.
+  Existing `Network` resources keep working if a license lapses.
+- The eBPF data plane, enabled cluster-wide.
+- [Multus](../networking/configuring/multiple-networks.mdx), for attaching workloads to a network as an additional interface.
+  Multus is not shipped by $[prodname].
```
lwr20: Worth a note on this bullet: on a cluster with `kube-proxy` disabled — which eBPF mode requires, and eBPF is a prerequisite two bullets up — **Multus needs the real API server endpoint too**, not just Calico.

Multus builds `/etc/cni/net.d/multus.d/multus.kubeconfig` from `KUBERNETES_SERVICE_HOST` / `KUBERNETES_SERVICE_PORT`. With no explicit value, Kubernetes injects the ClusterIP, so the kubeconfig names `https://10.43.0.1:443` — which only resolves once the eBPF dataplane is programmed. On a freshly booted node that isn't true yet, so **every pod sandbox on that node fails for minutes**:

```
multus failed (add): error waiting for pod:
Get "https://10.43.0.1:443/api/v1/namespaces/.../pods/..." context deadline exceeded
```

Fix is the Multus equivalent of the `kubernetes-services-endpoint` ConfigMap:

```
kubectl -n kube-system set env ds/kube-multus-ds \
  KUBERNETES_SERVICE_HOST=<api-host> KUBERNETES_SERVICE_PORT=<api-port>
```

Measured on an RKE2 rig: before, a node reboot left `virt-handler` crashlooping and the VM unrunnable for several minutes; after, the node was back to `Ready` in 38s with zero sandbox failures.

The gap is that the two halves live on separate pages. [Enable eBPF](../operations/ebpf/enabling-ebpf.mdx) covers the ConfigMap in detail but frames it purely as a Calico concern ("the operator will ... do a rolling update of $[prodname] to pass on the change"). This page requires Multus and correctly notes it isn't shipped by $[prodname] — but nothing states the general rule: *with kube-proxy disabled, anything that must reach the API server before the dataplane is programmed needs the real endpoint, not the ClusterIP.*

Tested on RKE2 v1.33.5+rke2r1, EE `2026-08-20-v3-24-2-kettle`, Multus v4.3.0. Claude-assisted, verified on a live cluster.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Fixed. The section opener no longer reads as "DHCP does not work here". It now says Calico IPAM is the authority but the guest still receives its address over DHCP, from KubeVirt inside the pod, and there is a new bullet for the case you describe, where a VM that used DHCP before it moved keeps that address through Calico IPAM.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Removed. I could not confirm either claim, so the bullet is now just "an active Calico Enterprise license". If you can tell me on your return whether the Network admission path checks the licence, and what happens to an existing Network when one expires, I will put the detail back.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Added. There is now an MTU section on this page covering both consequences you describe: a workload can end up on a smaller MTU than the machines it shares the VLAN with, and on a prepared bridge the VLAN device carrying the node's address drops too. It points at Installation.spec.calicoNetwork.mtu and at pinning the bridge. The same fact is now on the BYO page and in the ownership section of 2944.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Added as a prerequisite here, and 2947 now says the same next to the attachment definition example, since that is where someone hand-authoring one will be looking. Thanks for the version list, that is exactly the detail that makes it actionable.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Added as a prerequisite, with the failure described as you found it: with kube-proxy disabled the API server's cluster IP does not resolve until the dataplane is programmed, so Multus cannot reach it on a freshly booted node and pod sandboxes fail until it can. The bullet points at the eBPF page for the ConfigMap equivalent. I have not tried to state the general rule on that page as well; that feels like a separate change to the eBPF docs rather than something to fold in here.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Taken, thanks, and it was the right call to push on this. The platform requirement now states the capability rather than the version: support for the bridge port locked flag, which arrived upstream in 5.18, with a note that distributions backporting it satisfy the requirement at a lower reported version and that RHEL 9 and RHCOS are the notable case. It also tells the reader how to check a node directly, looking for locked on in bridge -d link show, so nobody has to reason from uname -r at all. The prerequisite on 2947 says the same in shorter form.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Leaving this open rather than guessing, and tagging @aaaaaaaalex since Shaun is away.

Alex, the question is what happens to ARP on an L2 network. This page currently says, flatly, that broadcast and multicast traffic is dropped in both directions. Lance points out ARP demonstrably works, arping from a peer gets unicast replies, and it has to work for the feature to function at all. The page also lists "Broadcast and ARP traffic in flow logs" as not in this release, which implies ARP traffic exists, and the concept page says Calico validates ARP sender fields.

So the two statements cannot both be right as written. What is the actual behaviour: is ARP excepted from the broadcast drop, is it handled specially, proxied or validated and then forwarded, and does the answer differ between requests and replies, or between workload-to-workload and workload-to-external? Once I know I will state it precisely here rather than softening the sentence.

One more on this page while I have you, from Lance and not yet addressed. The eBPF data plane is documented as required, but nothing appears to enforce it: the Network CRD's CEL rules do not check the dataplane, so a Network applied on an iptables cluster is accepted by the API. Nobody has tested what happens next. If the answer is "workloads attach without BPF anti-spoof" that matters a great deal more than "nothing happens", and it changes how strongly this page has to warn about it.

calico-enterprise/reference/l2-bridge-support.mdx
ctauchen: Leaving this one open too, and I have put the question to @aaaaaaaalex in the ARP thread on this page alongside it, since both are behaviour questions rather than wording.

Nothing on the page has changed for now. If the answer is that workloads attach without BPF anti-spoof on an iptables cluster, this needs a warning rather than a prerequisite bullet, so I would rather write it once I know which it is.

ctauchen:

PR #2946, opened 2026-08-21
DOCS-2997: Write the guide for preparing an existing bridge

ctauchen:
Fills in the guide for preparing an existing bridge. Fourth in the DOCS-2997 sequence. It comes before the main setup guide because its output feeds it.

This is the page for nodes whose only uplink also carries the node's own IP address, where Calico cannot create the bridge itself.

The guide is written around configuring the host so it brings the bridge up correctly at boot. Moving a running IP address onto a bridge reaches the same end state by a much more dangerous route, so the page does not teach it.

What the page covers:

- The three properties Calico checks, and the ordering warning that enabling VLAN filtering takes a host address dark unless VLAN membership already exists.
- What Calico deliberately does not require, which keeps the required list credible.
- Why the bridge MAC address must be set explicitly and must not be the trunk's.
- The VLAN membership the bridge device needs in its own right, which is the step most likely to be missed and the one most likely to take a node off the network.
- Two hazards that are not settings: a bridge named with the prefix Calico treats as its own, and the workload startup cost of spanning tree.
- Verification in two rounds, because the failure modes that matter most only appear when the first workload attaches.

This is the page in the set that most needs an engineering read, for two reasons.

The annotated NetworkManager example is constructed from the documented requirements rather than copied from a working node. It carries a caution telling readers to validate it in a lab first. Alex has a working configuration from the EP2 hash release, and if we can get it, it should either replace this example or be added alongside it. I would rather not ship an invented host network configuration without someone confirming it.

The failure behaviour described in the last section, and the three requirements, come from the code audit rather than from testing. Worth a confirmation that they still hold.

One known interim issue: the link to the concept page uses a section anchor that does not exist in main until the concept page PR merges. The file exists, so the link resolves, but the anchor will not jump until then.

Changed page: https://deploy-preview-2946--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/l2-bridge/byo-bridge

### DIFF
--- a/calico-enterprise/networking/l2-bridge/byo-bridge.mdx
+++ b/calico-enterprise/networking/l2-bridge/byo-bridge.mdx
@@ -6,66 +6,308 @@ description: Configure a Linux bridge on your nodes so Calico Enterprise can use
 
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-Use this guide when the uplink that must carry your VLANs is also the one your node's IP address depends on. $[prodname] cannot take that interface over on its own, so you build the bridge and $[prodname] uses it.
+Use this guide when the interface that must carry your VLANs is also the one your node's IP address depends on.
+$[prodname] cannot take that interface over on its own, so you build the bridge and $[prodname] uses it.
+For why, see [Who owns the host network configuration](about-l2-bridge.mdx#who-owns-the-host-network-configuration).
 
-The goal is not to move a running address onto a bridge. It is to have the host bring the bridge up correctly at boot, so the node's address arrives in the right place the first time.
+If $[prodname] can have an interface of its own, you do not need this guide.
+Use [Connect workloads to an existing VLAN](connect-vlan.mdx) instead.
+
+:::note
+
+The goal is not to move a running IP address onto a bridge.
+It is to have the host bring the bridge up correctly at boot, so the node's address arrives on the right device the first time.
+Configure this the way you configure the rest of your host networking, and verify it survives a reboot.
+
+:::
 
 ## Before you begin
 
-This procedure changes how your nodes get their IP addresses. Read this section before you start.
+This procedure changes how your nodes obtain their IP addresses.
+A mistake takes the node off the network, and you will not be able to fix it over the network.
+
+You need:
+
+- Root access to the node, and ownership of its network configuration.
+- Out-of-band access to the node, such as a console, or a tested way to roll the configuration back.
+- The name of the interface that will become the trunk.
+- The VLAN your host's own traffic uses, or confirmation that it is untagged.
+- The VLAN IDs your workloads will use.
+- The node's IP address, prefix length, and default gateway.
+
+Work on one non-critical node first, confirm it, and only then apply the same configuration to the rest of the cluster.
 
 ## What you are building
 
-Before and after, so the configuration further down makes sense.
+Today, your node's IP address sits directly on a physical interface, and that interface carries the node's traffic.
+
+Afterwards, that interface is a port on a bridge and carries no address of its own.
+The bridge carries your workload VLANs, and the node's address sits either on the bridge or on a VLAN device layered on top of it.
+$[prodname] attaches workloads to the same bridge as access ports.
 
 ## What $[prodname] requires
 
-Three settings on the bridge. $[prodname] checks all three and will not use a bridge that is missing any of them.
+$[prodname] checks three properties of the bridge and refuses to use it if any are missing.
+It does not set them for you, because your host network configuration owns them and would undo the change.
+
+| Requirement | Setting | Why |
+| ----------- | ------- | --- |
+| VLAN filtering enabled | `vlan_filtering=1` | Without it the bridge ignores VLAN tags entirely, so segments are not isolated from each other. |
+| An explicitly set MAC address | Any address, set deliberately | Stops the bridge's address changing as workloads come and go. See below. |
+| The 802.1Q tag protocol | `vlan_protocol=802.1Q` | $[prodname] programs 802.1Q VLAN membership and creates 802.1Q sub-devices. On an 802.1ad bridge none of it matches. |
+
+:::caution
+
+Enable VLAN filtering **before** you put the node's address on the bridge, or as part of the same configuration change.
+Turning it on makes the kernel start enforcing VLAN membership immediately, and an address on a bridge with no VLAN membership stops receiving and sending traffic the moment filtering comes on.
+
+:::
 
 ### Why the MAC address must be pinned, and must not be the trunk's
 
-Two separate failures, both avoided by the same setting.
+A Linux bridge has no MAC address of its own.
+By default the kernel borrows one from whichever attached port has the numerically lowest address, and it re-runs that choice every time a port is added or removed.
+
+On an L2 bridge the ports are workload interfaces, which appear and disappear constantly.
+Left to itself, the bridge's address changes under your node, and if the node's IP address is on that bridge, its traffic stops being delivered.
+
+Setting an address explicitly stops the kernel from ever reconsidering it.
+
+Do not use the trunk interface's own MAC address for the bridge.
+When two devices in the same path present the same address, some network card drivers reject the second programming attempt and the bridge is left non-functional.
+Use a distinct address.
+
+A locally administered address is a good choice, because it cannot collide with a real card.
+Set the second least significant bit of the first byte: `02:00:00:00:00:01` and `0a:1b:2c:3d:4e:5f` are both locally administered.
 
 ### What $[prodname] does not require
 
-Just as useful as the required list, and it keeps the required list credible.
+Just as useful as the list above, and it keeps that list credible.
+
+- **Enslaving the trunk.**
+  $[prodname] adds the trunk interface to the bridge itself.
+  The interface only has to exist and be named in the `Network`.
+- **A particular default PVID.**
+  $[prodname] does not read or change the bridge's default PVID.
+- **Removing an existing VLAN device.**
+  If you already have a VLAN sub-device on the bridge, $[prodname] reuses it rather than failing.
+- **Bringing the bridge up.**
+  $[prodname] brings up a bridge that is administratively down.
+- **Self VLAN memberships.**
+  $[prodname] never adds or removes the bridge device's own VLAN memberships, or their PVID and untagged flags.
+  They are yours to manage.
 
 ### Two hazards that are not settings
 
-A naming collision that deletes your bridge, and an option that quietly slows workload startup.
+**Do not name the bridge with a `calb-` prefix.**
+$[prodname] treats that prefix as meaning it owns the bridge, and it deletes bridges it owns when they are no longer needed.
+A bridge of your own named that way can be deleted from under you.
+Nothing rejects the name when you create the `Network`, so this is on you to avoid.
+
+**Spanning Tree Protocol is allowed, but it costs workload startup time.**
+Every new access port waits out a forwarding delay before it passes traffic, which delays every workload attaching to the bridge.
+$[prodname] warns once and leaves STP alone on a bridge it did not create.
+Turn it off unless you need it.
 
 ## Configure the host
 
-Write this as persistent configuration. A change that does not survive a reboot is a trap.
+Write this as persistent configuration, using whatever manages networking on your nodes.
+A change made with `ip` commands alone disappears at the next reboot, and the node comes back with its address on an interface that is now a bridge port.
+
+The steps below describe what the configuration has to achieve.
+The example that follows shows one way to express it.
+
+1. Create a bridge, with VLAN filtering enabled, the 802.1Q tag protocol, STP disabled, and an explicitly set MAC address that is not the trunk's.
+2. Add the trunk interface to the bridge as a port, and give that port membership of every VLAN it must carry, including the VLAN your host traffic uses.
+3. Give the bridge device itself membership of the VLAN your host traffic uses.
+   See [Self VLAN membership](#self-vlan-membership).
+4. Remove the IP configuration from the physical interface, so it carries no address.
+5. Put the node's IP address, prefix, and default gateway on the bridge or on a VLAN device above it, depending on whether your host traffic is tagged.
+   See [Where the host's address goes](#where-the-hosts-address-goes).
+6. Apply the configuration and confirm the node is still reachable.
+7. Reboot the node and confirm it comes back with the same addressing.
 
 ### Where the host's address goes
 
-Two shapes, depending on whether your host's own traffic is tagged.
+Two shapes, depending on your host's own traffic.
+
+**Untagged host traffic.**
+Put the node's address directly on the bridge device.
+The bridge port facing the trunk needs the host's traffic to arrive untagged, which means setting a native VLAN on the `Network` later so $[prodname] knows which VLAN untagged traffic belongs to.
+
+**Host traffic on its own VLAN.**
+Create a VLAN device on top of the bridge for that VLAN — for example, VLAN 100 on `br-l2` gives `br-l2.100` — and put the node's address on the VLAN device.
+The bridge itself carries no address.
+
+The distinction matters and is easy to invert.
+A VLAN device **above** the bridge, holding the host's address, is the supported arrangement.
+A VLAN device **below** the bridge, standing in for the trunk, is not supported.
 
 ### Self VLAN membership
 
-The step that is easiest to miss, and the one most likely to take a node off the network.
+For the host to receive traffic on a VLAN through the bridge, the bridge device needs membership of that VLAN in its own right.
+This is separate from the membership you give the trunk port, and it is the step most likely to be missed.
+
+Without it, the bridge forwards that VLAN's traffic between its ports but never delivers any of it to the host, so a node whose address depends on that VLAN goes dark.
+
+$[prodname] does not add this membership and does not remove one you have added.
+
+Some configuration tools have no field for it.
+Where that is the case, set it with a hook that runs when the bridge comes up:
+
+```bash
+bridge vlan add dev br-l2 vid 100 self
+```
+
+Confirm it with `bridge vlan show`, where the bridge device appears in the list alongside its ports.
 
 ### An annotated example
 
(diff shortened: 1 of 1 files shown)


calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -6,66 +6,308 @@ description: Configure a Linux bridge on your nodes so Calico Enterprise can use
...
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-Use this guide when the uplink that must carry your VLANs is also the one your node's IP address depends on. $[prodname] cannot take that interface over on its own, so you build the bridge and $[prodname] uses it.
+Use this guide when the interface that must carry your VLANs is also the one your node's IP address depends on.
```
ctauchen: We can't assume that someone landing on this page knows the full situation. You must situate the reader by explaining what this page is for but also how it fits in the overall story. For example, it should be clear in this first paragraph who this applies to (people who need the BYO option) and who it doesn't (people who can do managed, people who have no need for L2).

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -6,66 +6,308 @@ description: Configure a Linux bridge on your nodes so Calico Enterprise can use
...
+$[prodname] cannot take that interface over on its own, so you build the bridge and $[prodname] uses it.
+For why, see [Who owns the host network configuration](about-l2-bridge.mdx#who-owns-the-host-network-configuration).
 
-The goal is not to move a running address onto a bridge. It is to have the host bring the bridge up correctly at boot, so the node's address arrives in the right place the first time.
+If $[prodname] can have an interface of its own, you do not need this guide.
```
ctauchen: Together with the earlier part, I think we need to explain this very literally with the one NIC or two+ NICs description.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -6,66 +6,308 @@ description: Configure a Linux bridge on your nodes so Calico Enterprise can use
...
+Use [Connect workloads to an existing VLAN](connect-vlan.mdx) instead.
+
+:::note
+
+The goal is not to move a running IP address onto a bridge.
+It is to have the host bring the bridge up correctly at boot, so the node's address arrives on the right device the first time.
+Configure this the way you configure the rest of your host networking, and verify it survives a reboot.
+
+:::
```
ctauchen: This is a weird note. Remove.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -2,70 +2,322 @@
...
-Two separate failures, both avoided by the same setting.
+A Linux bridge has no MAC address of its own.
+By default the kernel borrows one from whichever attached port has the numerically lowest address, and it re-runs that choice every time a port is added or removed.
+
+On an L2 bridge the ports are workload interfaces, which appear and disappear constantly.
+Left to itself, the bridge's address changes under your node, and if the node's IP address is on that bridge, its traffic stops being delivered.
+
+Setting an address explicitly stops the kernel from ever reconsidering it.
+
+Do not use the trunk interface's own MAC address for the bridge.
```
fasaxc: The issue is a little more subtle than that.  @aaaaaaaalex explained that not using the trunk's MAC isn't enough on the affected card, you also have to remove the trunk's fdb entry.  Seems a bit messy.  We might want to name the specific driver and give an explicit workaround for that card (and then drop the general "don't use the trunk's MAC advice" since the trunk's MAC is a good choice if you don't have a buggy driver!)

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -2,70 +2,322 @@
...
+When two devices in the same path present the same address, some network card drivers reject the second programming attempt and the bridge is left non-functional.
+Use a distinct address.
+
+A locally administered address is a good choice, because it cannot collide with a real card.
+Set the second least significant bit of the first byte: `02:00:00:00:00:01` and `0a:1b:2c:3d:4e:5f` are both locally administered.
 
 ### What $[prodname] does not require
 
-Just as useful as the required list, and it keeps the required list credible.
+Just as useful as the list above, and it keeps that list credible.
```
fasaxc, line 116: Yep, good to tell the reader that we're trying to be credible 😆

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -2,70 +2,322 @@
...
+
+- **Enslaving the trunk.**
```
fasaxc: Hmm, this might be a bug on our side, we really want the admin to configure it so the trunk is already attached when Calico starts

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -2,70 +2,322 @@
...
+  $[prodname] adds the trunk interface to the bridge itself.
+  The interface only has to exist and be named in the `Network`.
+- **A particular default PVID.**
+  $[prodname] does not read or change the bridge's default PVID.
+- **Removing an existing VLAN device.**
+  If you already have a VLAN sub-device on the bridge, $[prodname] reuses it rather than failing.
+- **Bringing the bridge up.**
+  $[prodname] brings up a bridge that is administratively down.
+- **Self VLAN memberships.**
+  $[prodname] never adds or removes the bridge device's own VLAN memberships, or their PVID and untagged flags.
```
fasaxc: I think we add, but don't remove.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -2,70 +2,322 @@
...
 ### Two hazards that are not settings
 
-A naming collision that deletes your bridge, and an option that quietly slows workload startup.
+**Do not name the bridge with a `calb-` prefix.**
+$[prodname] treats that prefix as meaning it owns the bridge, and it deletes bridges it owns when they are no longer needed.
+A bridge of your own named that way can be deleted from under you.
+Nothing rejects the name when you create the `Network`, so this is on you to avoid.
+
+**Spanning Tree Protocol is allowed, but it costs workload startup time.**
+Every new access port waits out a forwarding delay before it passes traffic, which delays every workload attaching to the bridge.
```
fasaxc: I think it's two forwarding delays by default.  The forwarding delay can be set to something lower

fasaxc (dismissed):

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
```diff
@@ -2,70 +2,322 @@
...
+Setting an address explicitly stops the kernel from ever reconsidering it.
+
+Do not use the trunk interface's own MAC address for the bridge.
+When two devices in the same path present the same address, some network card drivers reject the second programming attempt and the bridge is left non-functional.
+Use a distinct address.
+
+A locally administered address is a good choice, because it cannot collide with a real card.
+Set the second least significant bit of the first byte: `02:00:00:00:00:01` and `0a:1b:2c:3d:4e:5f` are both locally administered.
 
 ### What $[prodname] does not require
```
lwr20, line 114: One more thing Calico does not require, but does change: the bridge MTU.

Calico attaches workload veths at the cluster-wide auto-detected MTU (1480 on the test cluster). A Linux bridge takes the minimum MTU of its ports, so `br-l2` drops from 1500 to 1480 when the first workload attaches — and `br-l2.100`, which holds the node's address and default route, drops with it. The trunk was explicitly configured `mtu=1500` and no MTU was set on the bridge, so this was not the operator's choice.

Raised as CORE-13428. Given this page is the one that tells operators how to hand a bridge over, it is probably worth a line telling them to pin the bridge MTU or `Installation.spec.calicoNetwork.mtu`.

lwr20 (commented):
Checked against `2026-08-20-v3-24-2-kettle` on an OpenShift cluster using exactly this arrangement — BYO `br-l2` over a single trunk, host address on `br-l2.100`. The requirements and hazards on this page all match what I found.

lwr20:
Testing L2 on a NetworkManager-based node turned up a hazard this page could close with a warning.

The keyfile example is correct and, importantly, **safe** — because it declares every VLAN on the trunk:

```ini
[bridge-port]
vlans=10,20,100
```

Our own OpenShift rig is safe for the same reason (`NodeNetworkConfigurationPolicy`, `trunk-tags: [10, 20, 100]`).

The hazard is what happens **later**. Calico's VLANs live in the `Network` CR; the node's live here. Nothing keeps the two in sync. Add a VLAN to a `Network` and forget to update this file, and that VLAN's L2 dies at the next connection reactivation or nmstate re-apply — 100% loss, node `Ready`, pod `Running`, `Network.status` empty, and it does **not** self-heal. Only a `calico-node` restart brings it back. Measured on an RKE2 + Leap Micro node; filed as CORE-13475.

It affects all three shapes this page describes, including `nativeVLAN`, where the native PVID/untagged flags revert too, so the host's own untagged path breaks as well.

Suggested addition, near the example:

> **Keep this list in step with the `Network`.** Every VLAN the `Network` uses must appear in `vlans=` here, on every node. If you add a VLAN to a `Network` without adding it here, that VLAN keeps working until the next time this connection is reactivated — then it stops, silently, and does not recover on its own.

Two smaller suggestions:

- Worth showing the **nmstate** form alongside the keyfile, since that is what OpenShift users actually use: `port: [{name: eno1, vlan: {mode: trunk, trunk-tags: [{id: 10}, {id: 20}, {id: 100}]}}]`.
- The OpenShift policy sets `vlan-default-pvid: 0`. Without it, NetworkManager's default PVID of 1 leaves stray `1 PVID Egress Untagged` entries on the bridge and its ports. Harmless here, but confusing when reading `bridge vlan show` during troubleshooting.

Also worth saying plainly somewhere on this page: **`managedBridge` is not exposed to any of this**, because Calico owns the bridge and NetworkManager leaves it alone. Verified — a full `nmcli con reload` + `con up` changed nothing on a managed bridge.

ctauchen:
Three questions for @aaaaaaaalex on this page, since Shaun is on leave and he named you as the person to ask.

These would normally go in the review threads they belong to, but a pending review of my own is blocking threaded replies for the moment. The replies to Shaun's five comments are written and will follow.

1. Self VLAN membership. Shaun corrected us: Calico adds the bridge's own VLAN memberships but never removes them. You needed a NetworkManager dispatcher hook to run "bridge vlan add dev br-l2 vid 100 self" for your management VLAN. I have reconciled those as follows, and I would like to know if it is right: Calico adds self-membership for the VLANs named in the Network, and the host's own VLAN is not named in the Network, so that membership stays the operator's job and persists because Calico never removes one. Is that the correct explanation?

2. CORE-13089, where a trunk not named in the Network goes net-blind on the eBPF RPF check. The ticket describes the BYO case and the page now carries a warning on that basis. Can the managed path hit the same failure, or is it structurally impossible there because the trunk is always named in hostConfig? If it can, the warning should be on the VLAN setup guide too.

3. Your Intel ice depoisoning hook. The page now describes the mechanism, removing the physical card's MAC from the bridge forwarding database when the bridge comes up, and notes that this needs unique MACs on the bridge and its sub-interfaces. It does not publish a script. Would you like your hook in the docs? An OpenShift reader on an ice NIC has to write one either way, and yours is known to work. The same question applies to the self-VID hook, which is already in the page as a two-line example.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
ctauchen: Done, and the polarity is flipped as you suggested. The general "do not use the trunk's MAC" instruction is gone, along with the driver-rejection explanation, which was not what happens.

In its place there is an Intel ice subsection: the kernel's MAC allow-list poisoned by the card's own address, filter programming stopping early, workload and management addresses dropped inbound with no counter and no log, and the fix being removal of the card's MAC from the bridge forwarding database, which needs unique MACs on the bridge and its sub-interfaces. It also warns not to tcpdump the trunk, since promiscuous mode masks the fault.

I have described the mechanism rather than publishing a script. Alex has a working hook and I would rather ship his than invent one, so I will ask him.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
ctauchen: Fair. That line is going in the editorial pass.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
ctauchen: Fixed. The bullet is gone, and the requirement is in the procedure instead: attach the trunk as part of the host configuration, so it is already a bridge port before Calico starts. That matches what you told Pedro in May and what Alex's NNCP does. Nothing in the docs now depends on the current tolerance, so it can be fixed without a docs change.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
ctauchen: Fixed, thanks. The page now says Calico adds membership for the VLANs named in the Network and never removes one, including one the operator added. It also says the host's own VLAN is theirs to add, because it is not named in the Network, and that memberships therefore accumulate and need clearing up on a node that should no longer carry them. Lance found the same behaviour from the other side in CORE-13456, so the isolation framing will go in the release note rather than on this page.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
ctauchen: Fixed. It now says two forwarding delays, around 30 seconds at the kernel default of 15, and mentions lowering forward_delay for anyone who needs STP on.

lwr20:
Separate from my drift comment above — two of this page's hazards bit me for real while building a NetworkManager-based node, so they're worth confirming as load-bearing rather than cautious:

- *"Do not use the trunk interface's own MAC address for the bridge."* I set the bridge's MAC to a live vNIC's address and **isolated the node** — it had to be recovered out of band. The warning is accurate and the consequence is a dark node, not a degraded one.
- *"On an 802.1ad bridge none of it matches."* Confirmed by toggling: `vlan_protocol 802.1ad` → 100% packet loss, back to `802.1Q` → 0%, twice. Filed separately as CORE-13462 because nothing reports it, which this page already says plainly.
- *"$[prodname] fails closed … The only signal is the Felix log."* Confirmed, and the refusal message is genuinely one of the better errors in the product — it names the bridge, the Network and the fix. Credit where due; the problem is only that it never leaves the log.
- The `:::caution` about enabling VLAN filtering **before** the address goes on the bridge is right, and the failure is immediate and total when you get it the wrong way round.

**One gap in "Verify before handing the bridge over."** The checklist is thorough for setup — including "reboot the node, repeat steps 1 to 4", which I confirmed works: a NetworkManager node came back after three reboots with bridge, VLANs and address all intact and Calico reprogramming its VLANs unaided.

What it doesn't cover is **later** change. Every check is framed as "before you hand the bridge over" or "after the first workload attaches". But the failure mode that actually bites in production is a host-network change made weeks afterwards — an nmstate re-apply, a config-management run, a connection reactivation — which silently strips Calico's VLAN off the trunk (CORE-13475, reproduced on RKE2 and OpenShift/RHCOS).

Suggest a third round, or a line under the existing checklist:

> **Re-run step 4 after any later change to the host's network configuration.** A change that reapplies the bridge or trunk configuration can remove VLAN memberships $[prodname] added. Nothing reports it: the node stays `Ready`, workloads stay `Running`, and only traffic on the affected VLAN stops.

That converts the page's existing (correct) instinct — check `bridge vlan show` — into the one habit that would catch this class of failure.

calico-enterprise/networking/l2-bridge/byo-bridge.mdx
ctauchen: Added to the list, as a "does not require, but does change" entry, with the instruction to pin the MTU on the bridge and its VLAN devices or set it on the Installation. Also stated on the support reference and in the ownership section of 2944, since that paragraph was the one promising Calico does not touch what the host owns.

ctauchen:
Applied, close to your suggested wording, and thank you for it — this is the most consequential thing anyone has found on this page.

By the trunk example there is now a caution saying the vlans= list must stay in step with the Network on every node, that the two are separate sources of truth with nothing syncing them, that a VLAN added to the Network but not here keeps working until the next reactivation or configuration management run and then stops silently, that nothing reports it, and that only a calico-node restart recovers it. It also notes the native VLAN case, where the PVID and untagged flags revert too.

Both smaller suggestions are in as well: the nmstate trunk-tags form alongside the keyfile, since that is what OpenShift readers write, and vlan-default-pvid: 0 with your reason for it, that the stray 1 PVID Egress Untagged entries are harmless but confusing when reading bridge vlan show during troubleshooting.

That managedBridge is not exposed to any of this is now stated in two places: at the end of the verification section here, and in the bridge mode comparison on 2944.

ctauchen:
Very useful to have these confirmed as load-bearing rather than cautious, thank you.

The verification gap is fixed. There is now a third round, after any later change to the host's network configuration, telling the reader to re-run bridge vlan show and confirm the trunk still carries every VLAN the Network uses, with the recovery step and a line saying why this is the round worth building a habit around: the other two catch setup mistakes, this one catches the change made weeks later.

On the trunk MAC, I need to flag a conflict rather than quietly keep or drop it. Shaun asked us to remove that instruction in his review of this page. His reasoning was that the trunk's MAC is a good choice on a working driver, and that on the one driver where it fails, an Intel ice card, avoiding it is not sufficient anyway because you also have to remove the card's MAC from the bridge forwarding database. So the page no longer carries the general rule and instead has an Intel ice section describing that fault and its workaround.

Your result cuts the other way: you set the bridge's MAC to a live vNIC's address and isolated the node. I suspect these are two different failures rather than a disagreement, since duplicating a MAC that is live on the segment is a problem in its own right, and your rig is on VMware where a vSwitch may drop it. But I would rather not decide that from the outside. Could you say whether the address you used was the trunk's own, and whether that node was on an ice NIC? Chris is deciding how to reconcile the two, and your answer is what settles it.

The 802.1ad and fail-closed confirmations are noted, and I have passed the credit for the refusal message on.

ctauchen:

PR #2947, opened 2026-08-21
DOCS-2997: Write the guide for connecting workloads to a VLAN

ctauchen:
Fills in the main setup guide for L2 bridge networking. Fifth in the DOCS-2997 sequence. It follows the guide for preparing an existing bridge, because a reader on a shared uplink needs that first.

This is the path for nodes where Calico can take over an interface of its own, so Calico creates and owns the bridge.

The guide leads with VMs. The feature is aimed at them, and a VM is configured differently from a pod: the interface is declared in the VM specification, and the guide says plainly that the launcher pod KubeVirt creates is not annotated by hand. Pods follow as the secondary case.

Points worth a reviewer's attention:

- Setting the multi-interface mode gets a step of its own. Installing Multus is not sufficient, nothing reports the omission, and the symptom shows up much later as a workload with two interfaces called eth0 and no connectivity.
- The IP pool step explains why the pool is marked as an L2 workload pool. Leaving it out gives you a blackhole route and a BGP advertisement for addresses that live on your own VLAN.
- The first-match-wins rule for host configuration entries is stated explicitly, since nothing in the API hints at it.
- The guide notes that KubeVirt's bridge interface binding is unrelated to the Calico L2 bridge. The same word means two different things in adjacent configuration and readers will trip on it.

Three things I could not verify and would like checked:

- The VM specification example is written from KubeVirt's documented API rather than copied from a working VM. Alex has manifests from the EP2 hash release that would be better.
- The attachment definition sets vlan as a JSON number. Confirm that is the expected type.
- The section on using an L2 network as the primary interface describes the capability and its consequence, but points at the Network reference for the annotation names rather than naming them, because I could not confirm the exact keys. That reference page is still to be written, so the pointer needs filling in.

Changed page: https://deploy-preview-2947--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/l2-bridge/connect-vlan

### DIFF
--- a/calico-enterprise/networking/l2-bridge/connect-vlan.mdx
+++ b/calico-enterprise/networking/l2-bridge/connect-vlan.mdx
@@ -35,7 +35,9 @@ Every item here is a hard requirement.
 The failures they cause are cheaper to avoid than to diagnose.
 
 - $[prodname] running the eBPF data plane.
-- Linux kernel 5.18 or later on every node that will carry the network.
+- A kernel supporting the bridge port `locked` flag on every node that will carry the network.
+  That arrived upstream in 5.18, but some distributions backport it: RHEL 9 and RHCOS report 5.14 and meet the requirement.
+  See [Platform requirements](../../reference/l2-bridge-support.mdx#platform-requirements).
 - An active $[prodname] license.
 - [Multus](../configuring/multiple-networks.mdx) installed.
   Every procedure on this page attaches the L2 network as an additional interface, which is what Multus does, so it is required here.
--- a/calico-enterprise/networking/l2-bridge/connect-vlan.mdx
+++ b/calico-enterprise/networking/l2-bridge/connect-vlan.mdx
@@ -25,7 +25,9 @@ Use this guide when your L2 trunk is separate from your host's primary NIC — f
 $[prodname] creates the bridge, enslaves that interface as the trunk, and attaches your workloads to it.
 
 If the only uplink on your nodes also carries the node's own IP address, $[prodname] cannot create the bridge.
-Start with [Prepare an existing bridge](byo-bridge.mdx), then return here from step 3 onwards.
+Start with [Prepare an existing bridge](byo-bridge.mdx), which covers building the bridge and writing the `Network` for it.
+Come back here for the IP pools, the reserved addresses, the attachment definitions, and attaching your first workload.
+Skip [Create the Network resource](#create-the-network-resource): the examples in it use `managedBridge`, which is the case where $[prodname] builds the bridge, not yours.
 
 ## Before you begin
 
@@ -35,7 +37,9 @@ The failures they cause are cheaper to avoid than to diagnose.
 - $[prodname] running the eBPF data plane.
 - Linux kernel 5.18 or later on every node that will carry the network.
 - An active $[prodname] license.
-- [Multus](../configuring/multiple-networks.mdx) installed, for attaching the L2 network as an additional interface.
+- [Multus](../configuring/multiple-networks.mdx) installed.
+  Every procedure on this page attaches the L2 network as an additional interface, which is what Multus does, so it is required here.
+  A workload taking an L2 network as its primary interface is configured with annotations instead and does not need Multus; see [Primary and secondary interfaces](../../reference/l2-bridge-support.mdx#primary-and-secondary-interfaces).
   OpenShift includes Multus by default.
   On other platforms it is a third-party component that $[prodname] does not ship.
 - KubeVirt installed, if you are attaching VMs.
--- a/calico-enterprise/networking/l2-bridge/connect-vlan.mdx
+++ b/calico-enterprise/networking/l2-bridge/connect-vlan.mdx
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
 
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-Use this guide when $[prodname] can take over an uplink of its own. If the only uplink on your nodes also carries the node's own IP address, start with [Prepare an existing bridge](byo-bridge.mdx) and return here afterwards.
+Use this guide when $[prodname] can take over an interface of its own.
+$[prodname] creates the bridge, enslaves that interface as the trunk, and attaches your workloads to it.
+
+If the only uplink on your nodes also carries the node's own IP address, $[prodname] cannot create the bridge.
+Start with [Prepare an existing bridge](byo-bridge.mdx), then return here from step 3 onwards.
 
 ## Before you begin
 
-Check all of these before you start. Each one is a hard requirement, and the failures they cause are easier to avoid than to diagnose.
+Every item here is a hard requirement.
+The failures they cause are cheaper to avoid than to diagnose.
+
+- $[prodname] running the eBPF data plane.
+- Linux kernel 5.18 or later on every node that will carry the network.
+- An active $[prodname] license.
+- [Multus](../configuring/multiple-networks.mdx) installed.
+  Multus is a third-party component and $[prodname] does not ship it.
+- KubeVirt installed, if you are attaching VMs.
+- An interface on each node that $[prodname] can take over, carrying your VLANs as a trunk, and **not** carrying the node's own IP address.
+
+For the full set of constraints, see [L2 bridge support and limitations](../../reference/l2-bridge-support.mdx).
 
 ## Set the multi-interface mode
 
-$[prodname] needs to know that Multus is in use. Nothing warns you if this is missed, so do it before anything else.
+$[prodname] has to be told that Multus is in use.
+This is a separate setting from installing Multus itself, and nothing reports its absence.
+The symptom appears much later, as a workload that comes up with two interfaces called `eth0` and no connectivity at all.
+
+1. Check the current value:
+
+   ```bash
+   kubectl get installation default -o jsonpath='{.spec.calicoNetwork.multiInterfaceMode}'
+   ```
+
+2. If it is not `Multus`, set it:
+
+   ```bash
+   kubectl patch installation default --type=merge \
+     -p '{"spec":{"calicoNetwork":{"multiInterfaceMode":"Multus"}}}'
+   ```
+
+3. Wait for `tigerastatus` to report all components as available:
+
+   ```bash
+   kubectl get tigerastatus
+   ```
 
 ## Gather your network details
 
-Collect these values first. Getting the subnet or the interface name wrong surfaces later as a workload that will not start.
+Collect these before you write any manifests.
+A subnet or interface name that does not match reality surfaces later as a workload that will not start.
+
+For each VLAN you intend to use:
+
+- The VLAN ID.
+- The subnet on that VLAN, in CIDR notation.
+- The default gateway on that subnet.
+- The range of addresses you are willing to let $[prodname] assign.
+
+And for the nodes:
+
+- The name of the trunk interface.
+- Whether that name is the same on every node.
+  If it is not, you need one host configuration entry per naming scheme.
 
 ## Create an IP pool for each workload VLAN
 
-The pool must exist before the `Network`, and its CIDR must line up with the VLAN's subnet.
+The pool has to exist before the `Network`.
+$[prodname] IPAM chooses a pool by checking which pools fall inside the VLAN's subnet, so a pool that does not match fails at workload creation with an error that does not obviously point back here.
+
+A VLAN that carries only host traffic does not need a pool.
+Create pools only for the VLANs your workloads will use.
+
+1. Create a pool whose CIDR sits inside the VLAN's subnet:
+
+   ```yaml
+   apiVersion: projectcalico.org/v3
+   kind: IPPool
+   metadata:
+     name: vlan-10-pool
+   spec:
+     cidr: 10.10.0.0/24
+     allowedUses:
+       - L2Workload
+     disableBGPExport: true
+   ```
+
+2. Apply it, and repeat for each workload VLAN.
+
+`allowedUses: L2Workload` marks the pool as belonging to an L2 network.
+Without it, $[prodname] treats the block as part of the routed pod network: it programs a blackhole route for it and advertises it over BGP, neither of which is correct for addresses that live on your VLAN.
+
+$[prodname] does not assign the first and last addresses of a subnet, or the gateway address.
+Size the pool with that in mind.
 
 ## Create the Network resource
 
-One `Network` describes the whole L2 network: its VLANs, and how each node reaches them.
+One `Network` describes the whole L2 network: the VLANs it carries, and how each node reaches them.
+
+Prefer one `Network` carrying several VLANs over several Networks carrying one each.
+Every `Network` adds work for $[prodname] on each node it selects.
+
+1. Write the `Network`, listing each VLAN with its subnet, and one host configuration entry describing the bridge and the trunk:
+
+   ```yaml
+   apiVersion: projectcalico.org/v3
+   kind: Network
+   metadata:
+     name: vlan-trunk
+   spec:
+     l2Bridge:
+       hostConfig:
+         - nodeSelector: ''
+           bridge:
+             managedBridge: {}
+           hostConnections:
+             - trunkPort:
+                 interface:
+                   name: eno2
+       vlans:
+         - vlan: { id: 10 }
+           subnets:
+             - cidr: 10.10.0.0/24
+         - vlan: { id: 20 }
+           subnets:
+             - cidr: 10.20.0.0/24
+   ```
+
+2. Apply it.
+3. Confirm $[prodname] created the bridge on a node that the selector matches:
+
+   ```bash
+   ip -br link show type bridge
+   ```
+
+`managedBridge: {}` tells $[prodname] to create and own the bridge.
+It derives the bridge name per node, so do not expect a name you chose.
+An empty `nodeSelector` applies the entry to every node.
 
 ### Tagged trunk
 
+The configuration above is the strict case: every VLAN is tagged on the wire, and the switch port facing the node is configured as a trunk carrying VLANs 10 and 20.
+
+This is the arrangement to prefer.
+Tagging is explicit in both directions, so there is no ambiguity about which VLAN a frame belongs to.
+
 ### Untagged traffic and the native VLAN
 
-For a segment your switch presents untagged.
+If your switch presents one segment untagged, set a native VLAN on the trunk port.
+It tells $[prodname] which VLAN untagged traffic arriving on that interface belongs to.
+
+```yaml
+hostConnections:
+  - trunkPort:
+      interface:
+        name: eno2
+      nativeVLAN: 10
+```
+
+The native VLAN must be one the `Network` defines.
+$[prodname] rejects a `Network` that names a native VLAN not present in its `vlans` list.
 
 ### Nodes with different interface names
 
-When the trunk interface is not called the same thing on every node.
+If the trunk is not called the same thing on every node, add one host configuration entry per naming scheme and select nodes with `nodeSelector`.
+
+```yaml
+hostConfig:
+  - nodeSelector: has(hardware) && hardware == "gen2"
+    bridge:
+      managedBridge: {}
+    hostConnections:
+      - trunkPort:
+          interface:
+            name: eno2
+  - nodeSelector: ''
+    bridge:
+      managedBridge: {}
+    hostConnections:
+      - trunkPort:
+          interface:
+            name: ens1f1
+```
+
+Entries are evaluated in order and **the first match wins**.
+Everything after it is ignored for that node — entries are not merged.
+Put your most specific selectors first and a catch-all last.
 
 ## Create a NetworkAttachmentDefinition for each VLAN
 
 Multus needs one attachment definition per VLAN you want workloads on.
+The definition names the `Network` and the VLAN within it.
+
+1. Create the definition:
+
+   ```yaml
+   apiVersion: 'k8s.cni.cncf.io/v1'
+   kind: NetworkAttachmentDefinition
+   metadata:
+     name: vlan-10
+   spec:
+     config: '{
+       "cniVersion": "0.3.1",
+       "type": "calico",
+       "network": "vlan-trunk",
+       "vlan": 10,
+       "log_level": "info",
+       "datastore_type": "kubernetes",
+       "nodename_file_optional": false,
+       "ipam": {
+         "type": "calico-ipam",
+         "assign_ipv4": "true",
+         "assign_ipv6": "false",
+         "ipv4_pools": ["vlan-10-pool"]
+       },
+       "policy": {
+         "type": "k8s"
+       },
+       "kubernetes": {
+         "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
+       }
+       }'
+   ```
+
+2. Apply it in the namespace where your workloads will run, and repeat for each VLAN.
+
+Set `vlan` explicitly.
+Create the definition in the same namespace as the workloads that reference it, or reference it across namespaces with the `namespace/name` form.
 
 ## Attach a VM
 
-For a VM, declare the interface in the VM specification. Do not edit the launcher pod's annotations.
+This is the main path for the feature.
+For a VM you declare the interface in the VM specification, alongside the VM's other devices.
+Do not annotate the launcher pod that KubeVirt creates — KubeVirt builds it from the specification below.
+
+1. Add an interface to the VM's device list, and a matching network referencing the attachment definition:
+
+   ```yaml
+   apiVersion: kubevirt.io/v1
+   kind: VirtualMachine
+   metadata:
+     name: app-server
+   spec:
+     template:
+       spec:
+         domain:
+           devices:
+             interfaces:
+               - name: vlan10
+                 bridge: {}
+         networks:
+           - name: vlan10
+             multus:
+               networkName: vlan-10
+   ```
+
+2. Start the VM and confirm it received an address from the pool:
+
+   ```bash
+   kubectl get vmi app-server -o jsonpath='{.status.interfaces}'
+   ```
+
+3. From inside the VM, confirm the interface is up and the address is configured.
+   KubeVirt serves the address over DHCP inside the VM, so the guest configures itself as it would on any network.
+
+The `bridge: {}` here is KubeVirt's own interface binding, which attaches the VM to the interface $[prodname] built.
+It is unrelated to the $[prodname] L2 bridge — the two uses of the word are unfortunate but distinct.
+
+To give the VM a specific address and MAC address rather than whatever the pool offers, see [Bring a VM over with its IP and MAC](vm-identity.mdx).
 
 ### Attaching a pod instead
 
+A pod references the attachment definition through an annotation, giving the interface a name inside the pod.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: app-with-vlan10
+  annotations:
+    k8s.v1.cni.cncf.io/networks: vlan-10@vlan10
+spec:
+  containers:
+    - name: app
+      image: nginx
+```
+
+The name after `@` is the interface name inside the pod.
+It is also the name that scopes per-interface annotations, such as a requested address, so choose it deliberately.
+
 ### Attaching a workload to more than one VLAN
+
+List one attachment definition per VLAN, each with its own interface name.
+A workload gets one access port per VLAN, and each port belongs to exactly one VLAN.
+
+```yaml
+annotations:
+  k8s.v1.cni.cncf.io/networks: vlan-10@vlan10,vlan-20@vlan20
+```
+
+For a VM, add a second interface and network pair to the specification in the same way.
 
 ### Using an L2 network as the primary interface
 
-Possible for both VMs and pods, with a significant trade-off.
+A workload can take an L2 network as its primary interface instead of an additional one, set with annotations rather than through Multus.
+See the [Network resource](../../reference/resources/network.mdx) for the annotations.
+
+Understand the trade-off before you do this.
+The workload has no route to or from the Kubernetes network.
+For a pod that is normally unacceptable, because kubelet cannot reach it for readiness and liveness checks.
+For a VM meant to behave like a traditional server on a VLAN, it is often what you want.
 
 ## Verify
 
-Work outwards from the node to the fabric, so a failure tells you where to look.
+Work outward from the node to your fabric, so that a failure tells you where to look.
+
+1. Confirm the `Network` exists and $[prodname] accepted it:
+
+   ```bash
+   kubectl get network vlan-trunk -o yaml
+   ```
+
+2. On a node running one of your workloads, confirm the trunk is enslaved to the bridge:
+
+   ```bash
+   ip link show master <bridge-name>
+   ```
+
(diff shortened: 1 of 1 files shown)


calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-Use this guide when $[prodname] can take over an uplink of its own. If the only uplink on your nodes also carries the node's own IP address, start with [Prepare an existing bridge](byo-bridge.mdx) and return here afterwards.
+Use this guide when $[prodname] can take over an interface of its own.
```
fasaxc:
```suggestion
Use this guide when your L2 trunk is separate from your host's primary NIC.  For example, your hosts' primary NIC is `eth0` and your L2 trunk is `eth1`.
```

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
 
-Check all of these before you start. Each one is a hard requirement, and the failures they cause are easier to avoid than to diagnose.
+Every item here is a hard requirement.
+The failures they cause are cheaper to avoid than to diagnose.
+
+- $[prodname] running the eBPF data plane.
+- Linux kernel 5.18 or later on every node that will carry the network.
+- An active $[prodname] license.
+- [Multus](../configuring/multiple-networks.mdx) installed.
+  Multus is a third-party component and $[prodname] does not ship it.
```
fasaxc: Technically Multus is only required for secondary interfaces.  Worth saying that OpenShift includes Multus by default?

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+- KubeVirt installed, if you are attaching VMs.
+- An interface on each node that $[prodname] can take over, carrying your VLANs as a trunk, and **not** carrying the node's own IP address.
```
fasaxc: It's a little more than that: the interface we take over must be the one carrying your L2 trunk traffic

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
-Collect these values first. Getting the subnet or the interface name wrong surfaces later as a workload that will not start.
+Collect these before you write any manifests.
+A subnet or interface name that does not match reality surfaces later as a workload that will not start.
+
+For each VLAN you intend to use:
+
+- The VLAN ID.
+- The subnet on that VLAN, in CIDR notation.
+- The default gateway on that subnet.
+- The range of addresses you are willing to let $[prodname] assign.
```
fasaxc, line 90:
```suggestion
- The range of addresses you are willing to let $[prodname] assign.
- Any other reserved addresses on your subnet (e.g. DNS servers).
```
They should create IPReservation objects to cover any reserved IPs that lie within the calico IP pools (that's existing function but it's more important for L2)

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+And for the nodes:
+
+- The name of the trunk interface.
+- Whether that name is the same on every node.
+  If it is not, you need one host configuration entry per naming scheme.
 
 ## Create an IP pool for each workload VLAN
 
-The pool must exist before the `Network`, and its CIDR must line up with the VLAN's subnet.
+The pool has to exist before the `Network`.
```
fasaxc: Not true, you can create them in either order (but both must exist before you try to use them).

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+       - L2Workload
+     disableBGPExport: true
+   ```
+
+2. Apply it, and repeat for each workload VLAN.
+
+`allowedUses: L2Workload` marks the pool as belonging to an L2 network.
+Without it, $[prodname] treats the block as part of the routed pod network: it programs a blackhole route for it and advertises it over BGP, neither of which is correct for addresses that live on your VLAN.
+
+$[prodname] does not assign the first and last addresses of a subnet, or the gateway address.
```
fasaxc: Just to be clear, we automatically reserve the first and last addresses of the VLAN's subnet, not the IP pool's CIDR.  So, you can subdivide your VLAN's subnet into >1 IP pool if you need to and you don't lose any additional IPs by subdividing.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
 ## Create an IP pool for each workload VLAN
 
-The pool must exist before the `Network`, and its CIDR must line up with the VLAN's subnet.
+The pool has to exist before the `Network`.
+$[prodname] IPAM chooses a pool by checking which pools fall inside the VLAN's subnet, so a pool that does not match fails at workload creation with an error that does not obviously point back here.
+
+A VLAN that carries only host traffic does not need a pool.
+Create pools only for the VLANs your workloads will use.
+
+1. Create a pool whose CIDR sits inside the VLAN's subnet:
```
fasaxc: `inside` is a little ambiguous.  In most cases, they'll want the IP pool to be equal to the subnet.  The exception would be if they need to sub-divide the subnet into different chunks for some reason.

Note that the IP pools should cover the IPs of any VMs that you plan to migrate in.  VMs migrating in need to assign their IP in IPAM.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+
+$[prodname] does not assign the first and last addresses of a subnet, or the gateway address.
+Size the pool with that in mind.
 
 ## Create the Network resource
 
-One `Network` describes the whole L2 network: its VLANs, and how each node reaches them.
+One `Network` describes the whole L2 network: the VLANs it carries, and how each node reaches them.
+
+Prefer one `Network` carrying several VLANs over several Networks carrying one each.
```
fasaxc: Some more points to help decide:
- A trunk can only belong to one network.  One trunk should imply one network.
- Multiple trunks on one host serving different VLANs requires multiple networks right now.
- Each Network creates a VLAN-aware bridge device on each host so that's the cost.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+2. Apply it.
+3. Confirm $[prodname] created the bridge on a node that the selector matches:
+
+   ```bash
+   ip -br link show type bridge
+   ```
+
+`managedBridge: {}` tells $[prodname] to create and own the bridge.
+It derives the bridge name per node, so do not expect a name you chose.
+An empty `nodeSelector` applies the entry to every node.
```
fasaxc, line 216: IIRC the bridge name is derived from the name of the network.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
 
 ### Tagged trunk
 
+The configuration above is the strict case: every VLAN is tagged on the wire, and the switch port facing the node is configured as a trunk carrying VLANs 10 and 20.
```
fasaxc:
```suggestion
The configuration above is the strict case: every VLAN is tagged on the wire, and the `eno2` trunk port is configured as a trunk carrying VLANs 10 and 20.
```

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+$[prodname] rejects a `Network` that names a native VLAN not present in its `vlans` list.
 
 ### Nodes with different interface names
 
-When the trunk interface is not called the same thing on every node.
+If the trunk is not called the same thing on every node, add one host configuration entry per naming scheme and select nodes with `nodeSelector`.
+
+```yaml
+hostConfig:
+  - nodeSelector: has(hardware) && hardware == "gen2"
```
fasaxc:
```suggestion
  - nodeSelector: hardware == "gen2"
```
has is implicit

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+  - nodeSelector: ''
+    bridge:
+      managedBridge: {}
+    hostConnections:
+      - trunkPort:
+          interface:
+            name: ens1f1
+```
+
+Entries are evaluated in order and **the first match wins**.
```
fasaxc:
```suggestion
Entries are evaluated in order and **the first matching `nodeSelector` wins**.
```

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+Everything after it is ignored for that node — entries are not merged.
+Put your most specific selectors first and a catch-all last.
```
fasaxc: If you don't have a catch all at the end, some nodes won't match and they'll ignore the Network entirely, which can be useful (but it doesn't affect pod scheduling so you need to make sure Pods/VMs have nodeSelectors too if not all nodes have a bridge).

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+   apiVersion: 'k8s.cni.cncf.io/v1'
+   kind: NetworkAttachmentDefinition
+   metadata:
+     name: vlan-10
+   spec:
+     config: '{
+       "cniVersion": "0.3.1",
+       "type": "calico",
+       "network": "vlan-trunk",
+       "vlan": 10,
```
fasaxc, line 301: Can you highlight these somehow; the rest is the standard NAD boilerplate

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+   kubectl get vmi app-server -o jsonpath='{.status.interfaces}'
+   ```
+
+3. From inside the VM, confirm the interface is up and the address is configured.
+   KubeVirt serves the address over DHCP inside the VM, so the guest configures itself as it would on any network.
+
+The `bridge: {}` here is KubeVirt's own interface binding, which attaches the VM to the interface $[prodname] built.
+It is unrelated to the $[prodname] L2 bridge — the two uses of the word are unfortunate but distinct.
+
+To give the VM a specific address and MAC address rather than whatever the pool offers, see [Bring a VM over with its IP and MAC](vm-identity.mdx).
```
fasaxc: Haven't read the other one yet, but we should document both migration and _just_ setting IP and MAC explicitly.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
 
 ### Attaching a pod instead
```
fasaxc, line 373: Both here and above, worth spelling out "what you get".

Above, I think you'll get a VM with one interface, plumbed to a secondary interface of the Pod containing the VM.  The primary interface of the VM's pod is unused.

Here you get the primary interface on the pod network, and the secondary interface on vlan 10

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -6,60 +6,361 @@ description: Expose one of your existing VLANs to the cluster with a Calico-mana
...
+             - trunkPort:
+                 interface:
+                   name: eno2
+       vlans:
+         - vlan: { id: 10 }
+           subnets:
+             - cidr: 10.10.0.0/24
+         - vlan: { id: 20 }
+           subnets:
+             - cidr: 10.20.0.0/24
```
fasaxc, line 199: Include the gateway routes here, will almost certainly want those.

fasaxc (dismissed):

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Make this conditional: If you're [correct phrase for going the managed route), you need to install Multus. OpenShift includes Multus by default.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -2,64 +2,375 @@
...
-Check all of these before you start. Each one is a hard requirement, and the failures they cause are easier to avoid than to diagnose.
+Every item here is a hard requirement.
+The failures they cause are cheaper to avoid than to diagnose.
+
+- $[prodname] running the eBPF data plane.
+- Linux kernel 5.18 or later on every node that will carry the network.
+- An active $[prodname] license.
+- [Multus](../configuring/multiple-networks.mdx) installed.
+  Multus is a third-party component and $[prodname] does not ship it.
+- KubeVirt installed, if you are attaching VMs.
```
ctauchen, line 47: Lead with conditional.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: We can strike this statement. Claude, you've gone overboard with your pre-procedure explanation. Hard limit to two sentences. One is better.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Include these and ask if examples needed.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: @fasaxc Yes we can highlight. Which lines?

Claude if you revise before we have response, highlight what seems relevant and we'll adjust after.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Agreed. Let's spell this out as part of the 'why' for the section. "By the end, you'll get ..."

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -2,64 +2,375 @@
...
+         "type": "calico-ipam",
+         "assign_ipv4": "true",
+         "assign_ipv6": "false",
+         "ipv4_pools": ["vlan-10-pool"]
+       },
+       "policy": {
+         "type": "k8s"
+       },
+       "kubernetes": {
+         "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
```
lwr20: This path does not exist on OpenShift, so the NAD as written fails CNI ADD for every OCP reader.

On RHCOS `/etc/cni/net.d/` contains no Calico files at all; they are under `/var/run/multus/cni/net.d/`. The only working NAD on the test cluster uses `/var/run/multus/cni/net.d/calico-kubeconfig`.

Suggest noting that the path is platform-dependent and telling the reader to copy it from the node's own `10-calico.conflist`. Same fix is needed in tigera/qe#210, which flags this in prose but keeps the old path in its YAML.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
```diff
@@ -2,64 +2,375 @@
...
+Start with [Prepare an existing bridge](byo-bridge.mdx), then return here from step 3 onwards.
 
 ## Before you begin
 
-Check all of these before you start. Each one is a hard requirement, and the failures they cause are easier to avoid than to diagnose.
+Every item here is a hard requirement.
+The failures they cause are cheaper to avoid than to diagnose.
+
+- $[prodname] running the eBPF data plane.
+- Linux kernel 5.18 or later on every node that will carry the network.
```
lwr20: This excludes the platform the feature was validated on.

The test cluster is RHCOS 9.6 / kernel `5.14.0-570.78.1.el9_6` and the feature works. The requirement is really the bridge port `locked` flag (upstream 5.18), which RHEL backports — `locked on` is set on every Calico access port on `br-l2` there.

An OpenShift admin who checks `uname -r` against this line concludes they are unsupported. Suggest stating the capability rather than a version, and naming RHEL 9 / RHCOS explicitly.

lwr20 (commented):
Two technical corrections from testing this page against an OpenShift cluster on the `2026-08-20-v3-24-2-kettle` hashrelease.

lwr20:
`managedBridge` is documented here — four mentions, inside a page about connecting workloads to a VLAN — and nowhere else in the set. I think it has earned a page of its own.

The reason is a test result rather than a preference. On NetworkManager-based nodes (RHEL 8/9, RHCOS, SLE Micro), a bridge Calico owns is **immune** to a whole class of failure that a bring-your-own bridge is exposed to:

- **`managedBridge`** — Calico creates `calb-<hash>`; NetworkManager reports it `connected (externally)` and does not touch it. A full `nmcli con reload` + `con up` changed nothing, 0% packet loss throughout.
- **`existingBridge`** — the node's config and the `Network` are two sources of truth with nothing keeping them in sync. If they diverge, the trunk loses Calico's VLAN at the next reconcile and L2 stops permanently and silently: node `Ready`, pod `Running`, `Network.status` empty, no self-healing. Filed as CORE-13475; I've left the detail on #2946.

So for anyone on a NetworkManager host with no hard requirement to manage the bridge themselves, `managedBridge` is materially the safer choice — and right now that is not discoverable. A reader who wants Calico to own the bridge has to already know the field exists to find it.

Suggested shape, whether as a short page or a section promoted from here:

- What `managedBridge: {}` does — creates and owns the bridge, derives the name per node (`calb-` prefix), deletes it when no longer needed
- **When to prefer it over `existingBridge`** — no node-side bridge configuration to write or keep in step with the `Network`; nothing for NetworkManager, nmstate or config management to diverge from
- What you give up — no control over the bridge name, and the bridge is not yours to attach other things to
- A pointer from `byo-bridge.mdx` saying "if you do not need to own the bridge, prefer this"

The existing line here is good and I would keep it verbatim: *"`managedBridge: {}` tells $[prodname] to create and own the bridge. It derives the bridge name per node, so do not expect a name you chose."* It just deserves somewhere more findable than a VLAN how-to.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Taken, with your wording. I kept the sentence that follows, describing what managed mode does, which the suggestion would otherwise have replaced.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Both points in. The bullet now scopes Multus to attaching the network as an additional interface and notes that OpenShift includes it by default. The conditional itself lives on the support reference, where Multus and multiInterfaceMode are both now scoped to that path, since this guide only documents it. I have assumed multiInterfaceMode is unnecessary on the primary interface path for the same reason Multus is; tell me if that is wrong.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Fixed. Split into two bullets with the switch side first: an interface cabled to a switch port configured as a trunk carrying your workload VLANs, and a statement that Calico attaches to the trunk you already have rather than creating one or configuring the switch. The "must not carry the node's own IP" requirement is now separate.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Taken, and turned into a step. The worksheet line on its own would have collected something the reader is never told what to do with, so there is now a numbered step for creating IPReservations covering in-use addresses that fall inside a pool, skippable if nothing outside the cluster uses the VLAN.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Fixed. It now says either order, with both needing to exist before a workload attaches.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Fixed. It now says the VLAN's subnet rather than the pool's CIDR, and spells out your consequence: splitting a subnet across several pools costs no extra addresses. I dropped "size the pool with that in mind", which was advice premised on the wrong reading.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Both points in. The default is now pool equals subnet, with subdivision as the deliberate exception. There is also a warning that a pool must cover the addresses of any VMs being migrated in, since a reader who carves out a small pool breaks their migration and only finds out at VM creation, a long way from the decision. The vm-identity page has gained a matching precondition.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Added, thanks. The page now states the rule rather than a preference: a trunk belongs to exactly one Network, so a single trunk means a single Network, and each Network costs a VLAN-aware bridge on every node it selects. It also warns against a second Network on the same trunk, which is CORE-13417.

One thing I have not changed. The support reference currently says more than one trunk on a node is unsupported, which your second point suggests is wrong. I have left it as is until someone can confirm two trunks in two Networks actually works in EP2.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Fixed. It now says the bridge is named after the Network, as calb-<network name>, and is the same on every node, so the verification step tells the reader to look for calb-vlan-trunk. Pedro's output on the EP2 build confirms the pattern. The calb- hazard on the BYO bridge page now explains where the prefix comes from, which it did not before.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Taken, with one change: "the switch port facing eno2" rather than "the eno2 trunk port". Naming the interface is the improvement you were after, but the thing being configured is at the switch end, and the prerequisites now lean on that distinction after your comment above.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Taken.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Taken. The Network reference already says "first matching entry wins", so the two now agree.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Both points in. The catch-all is now presented as a choice rather than a rule, with omitting it described as the way to confine the network to a pool of hardware. Your parenthesis is a warning of its own, since it is the part readers will get caught by: node selection here does not affect scheduling, so pods and VMs need their own nodeSelector when only some nodes carry the network.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Done. The two fields are highlighted and the text now says they are the only ones specific to the reader's setup. I also trimmed log_level, nodename_file_optional and the assign_ipv4/assign_ipv6 pair, which made the live fields stand out more than the highlighting does.

While in that block I fixed the kubeconfig path, which does not exist on OpenShift. Lance hit it and Pedro's working NAD uses the same corrected path. There is also a note that cniVersion must be 1.0.0 or lower.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Agreed, and the other page is retitled because of this. It is now "Set a VM's IP and MAC address" and covers both a VM arriving with an address it already has and a new VM that needs a fixed one. The filename is unchanged so no links break, and the inbound link text is updated here and in 2944 and 2951.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Added to both examples, in the same shape so the contrast is legible. The VM case now says the VM has one interface on the VLAN and cannot use the launcher pod's primary interface, and there is a manifest showing a VM with both, since that is what most readers will want.

Please sanity check that description when you are back. It is written from your comment rather than from testing, and it is the one claim on the page I have not been able to source.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Added, and it fixes more than it looks. The gateway was collected in the worksheet, used in no manifest, and then pinged in the final verification step, so the page did not work end to end.

Every Network example now declares the gateway as a route on the subnet, written as routes with an action.nextHop, since there is no gateway field. I have also noted that the reserved gateway address comes from a route's nextHop, so a subnet declared without routes has no reserved gateway and the address can be handed to a workload. The default route is right for a VM whose only interface is on the VLAN; for a workload that keeps its primary interface on the pod network the page says to use specific routes instead.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: @aaaaaaaalex one for you on this, since Shaun is away.

The page now says a managed bridge is named calb-<network name>, which Pedro's output confirms: a Network called calico-l2 gave calb-calico-l2.

Linux caps interface names at 15 characters and calb- is five, so a Network name longer than ten characters cannot fit. Our example, vlan-trunk, is exactly ten and lands on calb-vlan-trunk at exactly fifteen.

What happens above that: is the name truncated, hashed, or is there validation on the Network that rejects it? Lance found the Network CRD carries a CEL rule on the name prefix, so there may already be something covering this. If there is a practical limit on Network names, it should be on the Network reference and I will add it.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: @aaaaaaaalex a factual check for you here, since Shaun is away and this is the one claim on the page I could not source.

The page now tells the reader what each attachment produces. For the VM example, which declares a single vlan10 interface, it says:

"a VM with one interface, on VLAN 10. KubeVirt's launcher pod still has a primary interface on the pod network, but the VM cannot use it, the VM is on your VLAN and nothing else."

Is that right? It matches KubeVirt's model as I understand it and matches Pedro's manifests, which declare both a default pod interface and a VLAN interface when both are wanted, but I have not tested it.

The reason it matters is that the pod example immediately below gets both interfaces, so the two look like the same operation and are not. If a reader wants a VM with cluster networking and a VLAN address, they have to know to declare both.

lwr20:
Correction to my comment above — I was wrong about discoverability, and I'd rather say so than leave it standing.

I claimed `managedBridge` is documented "here and nowhere else" and that a reader "has to already know the field exists to find it". That's not right. **#2944 `about-l2-bridge.mdx` covers the choice properly** under *"Which bridge mode do you need"*, including *"$[prodname] can have its own uplink"* vs *"The uplink is shared with the host"*, and it already recommends managed mode where it applies:

> *"This is the simpler option wherever it applies. There is nothing for you to configure correctly on the host, nothing that a reboot can undo, and your node's own address never moves."*

So the concept is well documented; what lives only here is the `managedBridge: {}` **field syntax**. My "needs its own page" ask is therefore much weaker than I put it — the reader is not lost.

What I'd still offer, reduced to the part that survives:

The concept page recommends managed mode on **simplicity** grounds. There's now a **robustness** argument too, which is measured rather than editorial: on NetworkManager-based nodes (RHEL, RHCOS, SLE Micro), a bridge Calico owns is immune to a failure mode a BYO bridge is exposed to. If the node's declared VLAN set drifts from the `Network`'s, the next NetworkManager or nmstate reconcile strips Calico's VLAN off the trunk and L2 dies silently and permanently — node `Ready`, pod `Running`, `Network.status` empty. Filed as CORE-13475 and reproduced on both RKE2 and OpenShift/RHCOS.

That's arguably already gestured at by *"nothing that a reboot can undo"* — the instinct in the page is right, it just doesn't name the failure. One sentence in that section would carry it, e.g.:

> There is also nothing to keep in step: with a bridge $[prodname] owns, the VLAN list on the trunk cannot drift from the `Network`, because nothing else is managing it.

Detail and the drift warning are on #2946, which is where the exposure actually lives.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Fixed. The attachment definition now uses /var/run/multus/cni/net.d/calico-kubeconfig. Pedro's working manifest from the cluster uses the same path, so that is two confirmations. I trimmed some boilerplate from that block at the same time, following Shaun's point that only the network and vlan fields are decisions.

calico-enterprise/networking/l2-bridge/connect-vlan.mdx
ctauchen: Fixed here too. The prerequisite is now the bridge port locked flag rather than a version number, noting RHEL 9 and RHCOS report 5.14 and meet it, and pointing at the fuller explanation on the support reference.

ctauchen:
Both read, and thank you for posting the correction rather than leaving the first standing.

Taking the part that survives. The sentence you proposed is now in 2944, in the bridge mode comparison, near enough verbatim: with a bridge Calico owns the VLAN list on the trunk cannot drift from the Network, because nothing else is managing it. I dropped "nothing that a reboot can undo" at the same time, for the reason in your other comment.

I have not given managedBridge its own page, on the strength of your own correction. Your field-syntax point is still fair, so if the concept page turns out not to be enough once this set is published, that is the change to make rather than a new page.

ctauchen:

PR #2948, opened 2026-08-21
DOCS-2997: Write the guide for preserving a VM IP and MAC address

ctauchen:
Fills in the guide for bringing a VM onto an L2 network with its original IP address and MAC address. Sixth in the DOCS-2997 sequence, and it depends on an L2 network already existing.

The guide states the order plainly, because getting it wrong is destructive. The address and MAC address are declared on the VM before it starts, so the VM comes up already holding them. The review of the plan flagged that an earlier outline read as though you would start the VM and then move it to its old address, which breaks every connection made in between. The page now rules that out explicitly.

Other points:

- KubeVirt's own macAddress field is honored alongside the hwAddr annotation, so the page presents both and warns that conflicting values are rejected rather than resolved. OpenShift's MAC allocator fills that field in on new VMs, which makes the conflict easy to hit by accident.
- The page recommends ipAddrs over ipAddrsNoIpam, with the reason: bypassing IPAM leaves Calico unable to tell the workload's traffic from spoofed traffic.
- The write-back annotation is documented under its current name, network-status.
- The migration section describes what the release implements rather than what a reader will experience, and says it has not been through a full test cycle. No claim of seamlessness anywhere.
- Forklift carries a warning that Calico support for L2 networks is not in a released version, so a stock installation will not do this.

Two things to check:

- The status annotation is read here from the launcher pod. Confirm that is where a reader should look for a VM, rather than somewhere on the VM or the instance.
- The live migration paragraph is deliberately conservative while CORE-13289 is open. If that is closed and there is a test result, the wording can firm up.

Changed page: https://deploy-preview-2948--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/l2-bridge/vm-identity

### DIFF
--- a/calico-enterprise/networking/l2-bridge/vm-identity.mdx
+++ b/calico-enterprise/networking/l2-bridge/vm-identity.mdx
@@ -6,42 +6,173 @@ description: Bring a KubeVirt VM onto an L2 network holding its original IP addr
 
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-A VM that keeps its address and MAC keeps its identity: DNS entries, firewall rules, license servers, and monitoring all continue to work without being updated.
+A VM that keeps its IP address and MAC address keeps its identity.
+DNS entries still resolve to it, firewall rules still match it, license servers still recognise it, and monitoring still charts it — none of which has to be updated as part of the move.
+
+This guide covers a VM you are bringing onto an L2 network for the first time.
 
 ## Before you begin
 
+- An L2 network already created and verified.
+  See [Connect workloads to an existing VLAN](connect-vlan.mdx).
+- KubeVirt installed.
+- The VM's current IP address, prefix length, and MAC address, recorded from wherever it runs now.
+- Confirmation that the target VLAN's subnet contains that IP address.
+  $[prodname] assigns the address from an IP pool inside that subnet, so an address outside it cannot be assigned.
+- Confirmation that the address is not already in use.
+  $[prodname] fails the request rather than issuing a duplicate.
+
 ## How identity is preserved
 
-The address and MAC are declared before the VM starts. Nothing is renumbered afterwards.
+The order matters, so it is worth being explicit about it.
+
+The address and the MAC address are declared **on the VM before it starts**.
+$[prodname] assigns exactly those values when the interface is created, and the VM comes up already holding them.
+
+Nothing is renumbered afterwards.
+Do not start the VM, observe what it got, and then move it to the address you wanted: that sequence breaks every connection the VM makes in between, and defeats the point of preserving the address at all.
 
 ## Set the MAC address
 
-Two mechanisms, either of which works. Setting both to conflicting values is rejected.
+There are two ways to specify a MAC address, and either is sufficient.
+
+KubeVirt's own `macAddress` field on the interface is honored by $[prodname]:
+
+```yaml
+domain:
+  devices:
+    interfaces:
+      - name: vlan10
+        bridge: {}
+        macAddress: '52:54:00:3a:1c:04'
+```
+
+The `hwAddr` annotation is the other, scoped to the interface name:
+
+```yaml
+metadata:
+  annotations:
+    cni.projectcalico.org/vlan10.hwAddr: '52:54:00:3a:1c:04'
+```
+
+Use whichever fits how you generate the VM.
+The KubeVirt field is usually the better choice, because it lives with the rest of the interface configuration.
+
+:::caution
+
+If you set both, they must agree.
+When they conflict, $[prodname] rejects the interface rather than guessing which one you meant, and the VM does not start.
+
+On OpenShift this is easy to hit by accident.
+OpenShift ships a MAC address allocator that fills in the `macAddress` field on newly created VMs, so a VM may already carry a value before you add an annotation.
+Check the specification before you annotate.
+
+:::
 
 ## Set the IP address
 
+The address is requested with the `ipAddrs` annotation, scoped to the same interface name you gave the interface.
+
+```yaml
+metadata:
+  annotations:
+    cni.projectcalico.org/vlan10.ipAddrs: '["10.10.0.42"]'
+```
+
+The value is a JSON array, and the annotation prefix is the interface name followed by a dot.
+An annotation with no interface name applies to the primary interface.
+
+$[prodname] takes the address from the pool covering that VLAN and records the assignment, so the address is reserved for as long as the VM exists.
+
 ### Use ipAddrs rather than ipAddrsNoIpam
 
-Both request a specific address. Only one leaves $[prodname] able to enforce policy on it.
+Both request a specific address.
+Only one leaves $[prodname] able to enforce policy on it.
+
+`ipAddrsNoIpam` configures the address on the interface but bypasses IPAM entirely, so $[prodname] holds no record of it.
+On an L2 network that matters more than it does on the pod network: $[prodname] uses its knowledge of a workload's address to make policy decisions and to distinguish legitimate traffic from spoofed traffic.
+An address it does not know about cannot be protected.
+
+Use `ipAddrs`.
 
 ### One address per family
 
+An interface takes at most one IPv4 address.
+IPv6 is not supported on L2 networks in this release.
+
+Addresses cannot be added to or removed from an interface after it is set up.
+To change them, delete the VM and recreate it with the values you want.
+
 ## Confirm what was configured
 
-$[prodname] writes back what it actually applied. This is the authoritative record.
+$[prodname] writes back what it actually applied, which is the authoritative record of what the VM got.
+
+1. Read the status annotation from the VM's launcher pod:
+
+   ```bash
+   kubectl get pod -l vm.kubevirt.io/name=app-server \
+     -o jsonpath='{.items[0].metadata.annotations.cni\.projectcalico\.org/vlan10\.network-status}'
+   ```
+
+2. Check the reported `ipv4`, `mac`, `network`, and `vlan` against what you asked for.
+
+An admission policy restricts this annotation so that only the CNI plugin can write it.
+If it disagrees with your request, the request was not honored — it does not mean somebody edited the record.
 
 ## Verify from the network
 
-Check from outside the cluster, because that is what the VM's identity is for.
+Check from outside the cluster.
+The point of preserving an address is that other systems reach the VM at it, so that is what to test.
+
+1. From an existing machine on the same VLAN, ping the VM at its original address.
+2. On that machine, confirm the ARP entry for the address shows the original MAC address:
+
+   ```bash
+   ip neigh show 10.10.0.42
+   ```
+
+3. Confirm the workload endpoint carries the network and VLAN you expect:
+
+   ```bash
+   kubectl get workloadendpoints -o wide
+   ```
+
+4. If the VM is registered in DNS, confirm the name still resolves to the same address and that traffic to the name reaches the VM.
 
 ## Using Forklift
 
-Where a migration tool sets these values for you.
+Forklift copies a VM's configuration to the target KubeVirt VM as part of a migration, including the values this guide sets by hand.
+Where you use it, you are unlikely to be writing these annotations yourself.
+
+:::caution
+
+$[prodname] support for L2 networks is being contributed to Forklift and is not in a released version.
+Do not expect a stock Forklift installation to preserve addresses onto an L2 network.
+If you want to try this now, talk to your support contact rather than working from the upstream documentation.
+
+:::
 
 ## What to expect during migration
 
+What follows describes what this release implements.
+It is not a statement about what a particular migration will feel like, and it has not been through a full test cycle.
+
+$[prodname] preserves the address and the MAC address across a live migration between KubeVirt nodes, and announces the VM's new location with a gratuitous ARP so the fabric updates its forwarding.
+Established TCP connections are admitted on arrival rather than dropped for want of a handshake, which is intended to let sessions continue.
+
+Bringing a VM in from outside the cluster is a different matter.
+That is an offline operation: the VM is shut down while its disks are copied, so connections in progress do not survive, and clients reconnect once it starts.
+
+Plan for a disruption window and confirm the behavior you need in your own environment before you rely on it.
+
 ## Additional resources
+
+- [Connect workloads to an existing VLAN](connect-vlan.mdx)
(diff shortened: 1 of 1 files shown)


calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+The `hwAddr` annotation is the other, scoped to the interface name:
+
+```yaml
+metadata:
+  annotations:
+    cni.projectcalico.org/vlan10.hwAddr: '52:54:00:3a:1c:04'
+```
+
+Use whichever fits how you generate the VM.
+The KubeVirt field is usually the better choice, because it lives with the rest of the interface configuration.
```
fasaxc: Actually, we found in testing that this isn't equivalent because KubeVirt doesn't know about it, so it doesn't survive live migration.  The interface one is the way to go ^^

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+
+On OpenShift this is easy to hit by accident.
+OpenShift ships a MAC address allocator that fills in the `macAddress` field on newly created VMs, so a VM may already carry a value before you add an annotation.
+Check the specification before you annotate.
+
+:::
 
 ## Set the IP address
 
+The address is requested with the `ipAddrs` annotation, scoped to the same interface name you gave the interface.
```
fasaxc: Worth explicitly saying VM's `domain.devices.interfaces.name`  (is it really `domain`?)  it's different to the multus interface name, and the cali interface name(!)

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
 ## Confirm what was configured
 
-$[prodname] writes back what it actually applied. This is the authoritative record.
+$[prodname] writes back what it actually applied, which is the authoritative record of what the VM got.
+
+1. Read the status annotation from the VM's launcher pod:
+
+   ```bash
+   kubectl get pod -l vm.kubevirt.io/name=app-server \
+     -o jsonpath='{.items[0].metadata.annotations.cni\.projectcalico\.org/vlan10\.network-status}'
```
fasaxc: I don't think this works as written, 
- The `vlan10` will be replaced by an ungusable hashed name `pod1234abcd` due to KubeVirt's internals
- The pod IP isn't in the network-status (I took it out to avoid dual-provisioning)

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+2. On that machine, confirm the ARP entry for the address shows the original MAC address:
+
+   ```bash
+   ip neigh show 10.10.0.42
+   ```
+
+3. Confirm the workload endpoint carries the network and VLAN you expect:
+
+   ```bash
+   kubectl get workloadendpoints -o wide
```
fasaxc: Can't get weps with kubectl, they're backed by the pod annotations.  To see weps, you need the (rather sad)

`MULTI_INTERFACE_MODE=multus calicoctl get wep -o wide`

Without the env var, it defaults to the wrong mode and you get junk.

fasaxc (dismissed):

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+1. Read the status annotation from the VM's launcher pod:
+
+   ```bash
+   kubectl get pod -l vm.kubevirt.io/name=app-server \
+     -o jsonpath='{.items[0].metadata.annotations.cni\.projectcalico\.org/vlan10\.network-status}'
+   ```
+
+2. Check the reported `ipv4`, `mac`, `network`, and `vlan` against what you asked for.
+
+An admission policy restricts this annotation so that only the CNI plugin can write it.
```
lwr20: No such admission policy exists in this build — CORE-13288.

There are zero Calico/Tigera ValidatingAdmissionPolicies on the cluster. This sentence tells the reader to trust a record that is writable by anyone with pod-patch rights: Alex's 20 Aug results show a forged VLAN forces the victim endpoint admin-down, and a forged MAC with a valid VLAN programs a static FDB entry.

The test strategy also lists this VAP as a release criterion, and OS-T1423 is recorded FAIL, so this needs a product decision as well as a docs change.

lwr20 (commented):
Findings from testing this page against `2026-08-20-v3-24-2-kettle` on OpenShift. The live-migration section checks out — IP and MAC survived 5 of 5 cross-node migrations.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+```yaml
+metadata:
+  annotations:
+    cni.projectcalico.org/vlan10.ipAddrs: '["10.10.0.42"]'
+```
+
+The value is a JSON array, and the annotation prefix is the interface name followed by a dot.
+An annotation with no interface name applies to the primary interface.
+
+$[prodname] takes the address from the pool covering that VLAN and records the assignment, so the address is reserved for as long as the VM exists.
```
lwr20, line 123: Not quite — the reservation outlives the VM by about 15 minutes, and that has a visible consequence.

A VMI-scoped IPAM handle survives deletion of the VM and its namespace until kube-controllers reclaims it (`Confirmed IP leak after 15m1s`). Recreate a VM with the same name inside that window and Calico silently reuses the old handle's address and ignores `ipAddrs` entirely — I requested `10.10.0.205` and got `10.10.0.203`, with no error. Details on CORE-12786.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
 
 ## Before you begin
 
+- An L2 network already created and verified.
+  See [Connect workloads to an existing VLAN](connect-vlan.mdx).
+- KubeVirt installed.
+- The VM's current IP address, prefix length, and MAC address, recorded from wherever it runs now.
+- Confirmation that the target VLAN's subnet contains that IP address.
+  $[prodname] assigns the address from an IP pool inside that subnet, so an address outside it cannot be assigned.
+- Confirmation that the address is not already in use.
```
lwr20: Related to the note on line 100: in the same-name-recreate case it neither failed the request nor issued a duplicate — it silently substituted the previous address. Worth softening this claim or scoping it.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+DNS entries still resolve to it, firewall rules still match it, license servers still recognise it, and monitoring still charts it — none of which has to be updated as part of the move.
+
+This guide covers a VM you are bringing onto an L2 network for the first time.
 
 ## Before you begin
 
+- An L2 network already created and verified.
+  See [Connect workloads to an existing VLAN](connect-vlan.mdx).
+- KubeVirt installed.
+- The VM's current IP address, prefix length, and MAC address, recorded from wherever it runs now.
```
lwr20: Prefix length is asked for here but never used — `ipAddrs` takes a bare address (`["10.10.0.42"]`). Either drop it from the preconditions or say what it is for.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
```diff
@@ -2,46 +2,187 @@
...
+Use whichever fits how you generate the VM.
+The KubeVirt field is usually the better choice, because it lives with the rest of the interface configuration.
+
+:::caution
+
+If you set both, they must agree.
+When they conflict, $[prodname] rejects the interface rather than guessing which one you meant, and the VM does not start.
+
+On OpenShift this is easy to hit by accident.
+OpenShift ships a MAC address allocator that fills in the `macAddress` field on newly created VMs, so a VM may already carry a value before you add an annotation.
```
lwr20: This covers OpenShift, but says nothing about other platforms.

Off OpenShift the MAC allocator must be installed by hand, and whether kubemacpool substitutes cleanly is an open question. The named prospect for this feature is on RKE2 with upstream KubeVirt, so a reader on that platform gets no guidance here.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
lwr20: Confirming both points empirically on `2026-08-20-v3-24-2-kettle` (OpenShift, BYO bridge, secondary L2 NIC).

Hashed name: a VM requesting via `cni.projectcalico.org/net-0.ipAddrs` had its status written back under `cni.projectcalico.org/pod900d2c01c91.network-status` — so the request key and the status key differ exactly as you describe, and the documented jsonpath returns empty.

No IP in the payload: the annotation contained `{"name":"byo-vlan10","mac":"00:50:56:85:24:ce","vlan":10}` — name/mac/vlan only, no address, matching your dual-provisioning note.

So the "Confirm what was configured" step needs rewriting on both counts: the key it tells you to read does not exist for a secondary interface, and the `ipv4` field it tells you to check is not in the payload at all. (Apologies for the duplicate top-level comment on this line — deleting it, this thread already had it covered.)

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Fixed, thanks. This is the one that would have hurt.

The page now tells readers to set the MAC on the KubeVirt interface field, with a warning that the hwAddr annotation is not carried to the target node on live migration because KubeVirt does not know about it. The annotation is documented as the option for a workload that is not a VM, or a VM that will never migrate.

This also covers the EP2 doc action on CORE-13418, so the MAC allocator is now named per platform: OpenShift ships one, elsewhere it is kubemacpool, flagged as not verified with this feature outside OpenShift.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Done, and it turned up something worse than ambiguity. Pedro's VM used the VM's interface name in the annotation, exactly as this page told him to, and came up with a pool address rather than the one he requested. Lance's VM used net-0 and the request was honoured, with the status written back under a generated name.

So the annotation appears to be keyed on the pod-side name KubeVirt generates, not the VM's interface name. I have documented that, with a table of the four names in play, and it is the first question on my list for Alex. If it is wrong, the central mechanism of this page is wrong, so worth a look when you are back.

And yes, it really is domain.devices.interfaces[].name.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Fixed on both counts, and Lance confirmed both against the EP2 build. The address now comes from the VMI status. The MAC, network and VLAN come from the annotation matched by its suffix rather than by name, since the key carries the generated name. The instruction to check ipv4 is gone, because the payload does not carry it.

I also removed the sentence saying an admission policy restricts the annotation to the CNI plugin, since CORE-13288 did not land. That will go in the release note instead.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Fixed. It is now MULTI_INTERFACE_MODE=multus calicoctl get wep -o wide, with a note that a WorkloadEndpoint is derived from pod annotations rather than stored, so kubectl cannot list it, and that without the variable calicoctl reports incorrect values rather than failing.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: @aaaaaaaalex tagging you on this one, it is the question I would most like settled before EP2 ships.

What exactly keys the per-interface annotations for a KubeVirt VM?

Two conflicting data points:

- Pedro's VM used cni.projectcalico.org/vlan1001.ipAddrs, matching the name of the interface in the VM spec, which is what this page told him to do. The VM came up with 10.200.1.2 from the pool rather than the requested 10.200.1.11.
- Lance's VM used cni.projectcalico.org/net-0.ipAddrs and the request was honoured, with the status written back under cni.projectcalico.org/pod900d2c01c91.network-status.

On that basis the page now documents the pod-side interface name that KubeVirt generates, net-0 for the first Multus secondary, rather than the VM's own interface name. Is that correct, and does the same rule apply to the hwAddr annotation?

If it is wrong, the central mechanism of this page is wrong, so I would rather correct it now than ship it.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Removed. The sentence claiming an admission policy restricts the annotation is gone from this page, and the troubleshooting page now says the opposite where it tells the reader to compare request against status: nothing restricts writes to that annotation in this release, so a disagreement is not conclusive. CORE-13288 is on my list for the release note. Agreed it needs a product decision too, which I cannot make from here.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Documented. There is now a section on recreating a VM with the same name: the handle survives the VM and its namespace for around 15 minutes, and inside that window Calico reuses the previous address and ignores ipAddrs without reporting an error. It says to wait for the reclaim or use a different name. Thanks for the measurement, "silently substituted the previous address" is the part that makes it worth a caution rather than a footnote.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Scoped. The precondition now says the address must not be in use by another workload, and points at the same-name-recreate section for the exception, where the substitution behaviour is described in full.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Fixed. Prefix length is out of the preconditions, with a line saying it comes from the VLAN's subnet in the Network rather than from anything the reader supplies.

calico-enterprise/networking/l2-bridge/vm-identity.mdx
ctauchen: Addressed, though not settled. The page now splits by platform: OpenShift ships an allocator, and on other platforms you provide one yourself, with kubemacpool named as the upstream project and flagged as not verified with this feature outside OpenShift. That matches what Shaun said in the channel while leaving your open question open. I have asked Alex to close it, since the RKE2 case is the one that matters commercially.

ctauchen:

PR #2949, opened 2026-08-21
DOCS-2997: Write the L2 network troubleshooting guide

ctauchen:
Fills in the troubleshooting guide for L2 bridge networking. Seventh in the DOCS-2997 sequence, and near the end because it references the setup guides.

The guide is organized by symptom rather than by tool, so a reader arrives with a problem and finds it.

It opens with where the signals are, because that is genuinely unusual here and looking in the obvious places wastes time. Bridge prerequisite failures appear only in the Felix log inside calico-node: no Kubernetes event, no condition on the Network, no status on the workload, no metric. Attachment failures appear in the workload status and the CNI and Multus logs instead.

The diagnostics section states that bridge state is not collected by the standard bundle and has to be gathered by hand. A support case from a cluster in this state otherwise arrives with the Felix warning and nothing else about the bridge.

The entries deliberately favour failures that are silent, late, or self-concealing:

- A bridge that does not meet its prerequisites, where Calico fails closed and workloads stay administratively down.
- A workload with two interfaces called eth0, because the multi-interface mode was never set.
- A node taken off the network mid-configuration, with the three usual causes.
- A packet capture that hides the hardware filtering problem it was opened to investigate.
- A pod that cannot reach Service addresses because connect-time load balancing cannot be disabled per interface, with the workaround.

Two notes:

- The link to the multi-interface mode section resolves once the connect-vlan PR merges. The file exists, so the link works; the anchor lands after that merge.
- The diagnostics command assumes calicoctl runs inside the calico-node pod. Worth confirming that is the current recommended invocation.

Changed page: https://deploy-preview-2949--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/l2-bridge/troubleshoot

### DIFF
--- a/calico-enterprise/networking/l2-bridge/troubleshoot.mdx
+++ b/calico-enterprise/networking/l2-bridge/troubleshoot.mdx
@@ -6,44 +6,225 @@ description: Diagnose connectivity problems on Calico Enterprise L2 bridge netwo
 
 :::note
 
-L2 bridge networking is a tech preview feature. APIs and behavior may change before GA.
+L2 bridge networking is a tech preview feature.
+APIs and behavior may change before GA.
 
 :::
 
-Most L2 problems fall into a handful of shapes. Find your symptom below.
+Most problems on an L2 network fall into a handful of shapes.
+Find your symptom below.
 
 ## Where the signals are
 
-Read this first. L2 bridge problems report themselves in an unusual place, and looking anywhere else wastes time.
+Read this first.
+L2 problems report themselves somewhere unusual, and looking in the obvious places wastes time.
+
+Problems with the **bridge itself** appear only in the Felix log inside `calico-node`.
+There is no Kubernetes event, no condition on the `Network`, no status on the workload, and no metric.
+The messages are good — each names the bridge, the `Network`, and the command that fixes the problem — but you have to read the log to see them.
+
+```bash
+kubectl logs -n calico-system -l k8s-app=calico-node --tail=200 | grep -i bridge
+```
+
+Problems **attaching a workload** appear in the workload's own status, and in the CNI and Multus logs on the node that was scheduled to run it.
+
+```bash
+kubectl describe pod <pod>
+```
+
+On the node, CNI and Multus write to `/var/log/calico/cni/` and the container runtime's log.
 
 ## Collect diagnostic information
 
-What to gather before opening a support case, including the bridge state that the standard bundle does not capture.
+Gather this before opening a support case, because one part of it is not collected automatically.
+
+1. Run the standard diagnostics bundle:
+
+   ```bash
+   kubectl exec -n calico-system <calico-node-pod> -- calicoctl cluster diags
+   ```
+
+2. Collect the bridge state by hand on each affected node.
+   The bundle does not include it:
+
+   ```bash
+   ip -d link show type bridge
+   bridge link show
+   bridge vlan show
+   bridge fdb show
+   ```
+
+3. If the problem is not obvious, raise Felix's log level and reproduce it:
+
+   ```bash
+   kubectl patch felixconfiguration default --type=merge \
+     -p '{"spec":{"logSeverityScreen":"Debug"}}'
+   ```
+
+Set it back to `Info` afterwards.
 
 ## $[prodname] will not use the bridge I prepared
 
+Workloads on the `Network` stay administratively down and never get connectivity, and the Felix log repeats a warning every few seconds.
+
+$[prodname] fails closed here by design.
+Rather than attaching workloads to a bridge that cannot isolate them, it programs nothing at all.
+That means no trunk enslavement, no VLAN membership, and no workload interfaces.
+
+Check the three properties $[prodname] requires:
+
+1. VLAN filtering is enabled:
+
+   ```bash
+   ip -d link show <bridge> | grep vlan_filtering
+   ```
+
+2. The MAC address was set explicitly rather than inherited from a port.
+   If it was not, setting it to its current value is enough:
+
+   ```bash
+   ip link set dev <bridge> address <current-address>
+   ```
+
+3. The tag protocol is 802.1Q rather than 802.1ad:
+
+   ```bash
+   ip -d link show <bridge> | grep vlan_protocol
+   ```
+
+The warning in the log names which of these failed and the command to fix it.
+See [Prepare an existing bridge](byo-bridge.mdx) for the full requirements.
+
+Two side effects are worth knowing about.
+A single broken bridge can intermittently delay unrelated `Network` resources, because $[prodname] stops processing on the first failure each cycle.
+And the failure is per-node, so a `Network` that works on most nodes and fails on one reads like flakiness rather than a configuration error.
+
 ## A workload came up with two eth0 interfaces and no connectivity
+
+The multi-interface mode is not set.
+Multus being installed is not sufficient on its own, and nothing reports the omission until this happens.
+
+```bash
+kubectl get installation default -o jsonpath='{.spec.calicoNetwork.multiInterfaceMode}'
+```
+
+If the value is not `Multus`, set it as described in [Set the multi-interface mode](connect-vlan.mdx#set-the-multi-interface-mode), then delete and recreate the affected workloads.
 
 ## The workload has no IP address, or will not start
 
+The workload stays pending or failing, and the CNI log reports an address assignment failure.
+
+Check these in order:
+
+1. **No IP pool matches the VLAN's subnet.**
+   $[prodname] chooses a pool by checking which pools fall inside the subnet on the VLAN.
+   Nothing matching means no address, and $[prodname] fails rather than assigning an address from the wrong subnet.
+2. **The pool is not marked for L2 use.**
+   A pool without `allowedUses: L2Workload` is not a candidate for an L2 interface.
+3. **A requested address is already assigned.**
+   If you asked for a specific address with `ipAddrs`, and something already holds it, the request fails rather than duplicating it.
+4. **A requested address is reserved.**
+   $[prodname] does not assign the first or last address of a subnet, or the gateway address.
+
 ## The bridge was never created on this node
+
+Other nodes have the bridge and this one does not.
+
+1. Check that a host configuration entry matches this node.
+   Remember that entries are evaluated in order and the first match wins — a broad entry earlier in the list shadows a more specific one after it.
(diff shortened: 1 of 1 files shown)


calico-enterprise/networking/l2-bridge/troubleshoot.mdx
```diff
@@ -2,48 +2,236 @@
...
-What to gather before opening a support case, including the bridge state that the standard bundle does not capture.
+Gather this before opening a support case, because one part of it is not collected automatically.
+
+1. Run the standard diagnostics bundle:
+
+   ```bash
+   kubectl exec -n calico-system <calico-node-pod> -- calicoctl cluster diags
+   ```
+
+2. Collect the bridge state by hand on each affected node.
```
fasaxc: I have a PR up to add this to calicoctl, not sure if it'll land in time for EP2

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
```diff
@@ -2,48 +2,236 @@
...
+   The bundle does not include it:
+
+   ```bash
+   ip -d link show type bridge
+   bridge link show
+   bridge vlan show
+   bridge fdb show
+   ```
+
+3. If the problem is not obvious, raise Felix's log level and reproduce it:
```
fasaxc: Not recommended as written, debug logging is perf impacting unless you combine with a filename regex

fasaxc (dismissed):

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Ok, we'll scrap this.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Ok, will keep an eye out.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
```diff
@@ -2,48 +2,236 @@
...
+The multi-interface mode is not set.
+Multus being installed is not sufficient on its own, and nothing reports the omission until this happens.
+
+```bash
+kubectl get installation default -o jsonpath='{.spec.calicoNetwork.multiInterfaceMode}'
+```
+
+If the value is not `Multus`, set it as described in [Set the multi-interface mode](connect-vlan.mdx#set-the-multi-interface-mode), then delete and recreate the affected workloads.
+
 ## The workload has no IP address, or will not start
```
lwr20, line 125: Missing the failure I actually hit here, which is a plausible first-week user error.

A NAD referencing a `Network` that no longer exists leaves the VM stuck in `Starting` indefinitely, with virt-launcher at `Init:0/3`, no restarts and no pod IP. The real cause appears only in the launcher pod's events:

```
failed to create pod network sandbox ... error adding container to network "vm-migrated-vlan10":
failed to get Network "vlan-trunk" ... resource does not exist: Network(vlan-trunk) not found
```

The VMI condition says only `GuestNotRunning`, and `kubectl get vm` shows `Starting` forever. Three VMs on the test cluster had been wedged this way for over two days. Nothing on this page points a KubeVirt reader from the VM to the launcher pod, which is where the error lives.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
```diff
@@ -2,48 +2,236 @@
...
+
+Putting an interface into promiscuous mode changes how some network cards filter incoming traffic, which can mask a hardware filtering problem for exactly as long as the capture runs.
+
+Capture without promiscuous mode, or capture from a different interface in the path:
+
+```bash
+tcpdump -p -i <interface>
+```
 
 ## Workloads lost connectivity after a Network was deleted
```
lwr20, line 238: This entry covers workloads that were already running when the Network was deleted. The other half — new workloads created against a NAD whose Network is already gone — behaves differently and is not covered; see the note on line 121.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
```diff
@@ -2,48 +2,236 @@
...
+For a pod that needs both cluster Services and an L2 network, keep its primary interface on the $[prodname] pod network and attach the L2 network as an additional interface.
+
 ## The MAC address is not what I asked for
 
+1. **The VM does not start at all, and the CNI log reports a conflict.**
+   KubeVirt's `macAddress` field and the `hwAddr` annotation are both set, to different values.
+   $[prodname] rejects the interface rather than choosing.
+   Make them agree, or remove one.
+2. **The address is not the one you asked for.**
+   Check the `network-status` annotation to see what $[prodname] actually applied, and confirm the annotation you set is scoped to the right interface name.
```
lwr20, line 221: Worth qualifying: on this build the `network-status` annotation is not protected by anything.

#2948 states an admission policy restricts it to the CNI plugin, but no such policy exists (CORE-13288), so a disagreement between request and status can also mean somebody wrote to it. That makes this check less conclusive than it reads.

lwr20 (commented):
Checked against `2026-08-20-v3-24-2-kettle` on OpenShift. This page covers the bridge gotchas well — MAC pinning, VLAN-filtering ordering, self-VID membership and the promiscuous-mode masking trap all match what I hit in practice.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
```diff
@@ -10,40 +10,241 @@ L2 bridge networking is a tech preview feature. APIs and behavior may change bef
...
+The workload stays pending or failing, and the CNI log reports an address assignment
+failure.
+
+Check these in order:
+
+1. **No IP pool matches the VLAN's subnet.** $[prodname] chooses a pool by checking which
+   pools fall inside the subnet on the VLAN. Nothing matching means no address, and
+   $[prodname] fails rather than assigning an address from the wrong subnet.
+2. **The pool is not marked for L2 use.** A pool without `allowedUses: L2Workload` is not a
+   candidate for an L2 interface.
```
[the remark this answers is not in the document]
lwr20: `L2Workload` is valid — this troubleshooting step is correct as written.

On the EP2 build (`2026-08-20-v3-24-2-kettle`) the IPPool CRD enum accepts it:

```
"HostSecondaryInterface"  "L2Workload"  "LoadBalancer"  "Tunnel"  "Workload"
```

and a live pool on the validation cluster uses it: `l2-vlan10 -> allowedUses: ["L2Workload"]`.

The useful half of this comment is the other one though: `reference/resources/ippool.mdx:52` really does still list only Workload/Tunnel/HostSecondaryInterface/LoadBalancer, and the allowed-combinations list at lines 70-71 doesn't cover it either. So the reference page is the stale one, not this page.

Worth fixing alongside this docs set, since #2947 tells readers to create a pool with `allowedUses: ["L2Workload"]` and the reference page currently implies that is not a legal value.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
lwr20, line 125: This section has no entry for an unsupported `cniVersion` in the NAD, which presents exactly as the symptom in the heading.

A NAD declaring `cniVersion: 1.1.0` fails CNI ADD — Calico's plugin supports up to `1.0.0`. The pod stays in `ContainerCreating` and the only signal is in pod events:

```
plugin type="multus" ... failed (add): error adding container to network "<nad>":
incompatible CNI versions; config is "1.1.0", plugin supports [... "1.0.0"]
```

Nothing on the NAD, the Network, or (for a VM) the VMI says anything — a reader has to know to run `kubectl describe pod` on the sandbox. Suggest a row pointing there and naming the supported list.

Tested on RKE2 v1.33.5+rke2r1, EE `2026-08-20-v3-24-2-kettle`, Multus v4.3.0. Claude-assisted, verified on a live cluster.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Checked: CORE-13422 is scoped to EP3 and carries no EP2 fix version, so the manual collection step stands for this release. I have made the wording neutral about the omission and named which nodes to collect from, so it is a small edit when your calicoctl change lands.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Fixed. The step now sets logDebugFilenameRegex alongside logSeverityScreen, with a note that debug logging across the whole of Felix is expensive and worse on a cluster that is already unhealthy.

I have published "bridge" as the regex rather than a single filename, so it cannot silently match nothing, and mentioned bridge_network_mgr.go as the narrower option. If there is a better file set, say and I will pin it.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Added. There is now an entry for an attachment definition naming a Network that does not exist, including that a VM presents as Starting indefinitely with its launcher pod at Init:0/3, and that the error is only in the launcher pod's events rather than on the VM or the VMI. That last part is the bit a KubeVirt reader would not think to look for.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Covered. The deleted-Network entry now distinguishes the two halves: workloads that were already running lose an interface, and workloads created after the Network is gone never get one, with a link to the entry that describes that case.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Qualified. That step now says nothing restricts writes to the annotation in this release, so a disagreement between request and status can also mean something else wrote to it. The claim that prompted this is deleted from 2948.

calico-enterprise/networking/l2-bridge/troubleshoot.mdx
ctauchen: Added, as its own entry under the same symptom, saying the plugin supports up to 1.0.0 and that a later version fails at interface creation with the pod stuck in ContainerCreating. The ceiling is now also a prerequisite on the support reference and a note next to the attachment definition example in 2947.

ctauchen:

PR #2951, opened 2026-08-21
DOCS-2997: Point existing pages at L2 bridge networking

ctauchen:
Updates five existing pages so they account for the new L2 bridge networking set. Last in the DOCS-2997 sequence. Small diffs, but two of them are correctness fixes rather than additions.

The KubeVirt networking page uses the word bridge for KubeVirt's interface binding mode, which connects a VM to its own pod interface. That is a different thing from an L2 bridge network, and the two pages now sit next to each other in the sidebar. A note disambiguates them.

The same page states that live migration requires BGP networking without overlay. That is true of the pod network path it describes, but not of a VM on an L2 network, whose address is reachable on the reader's own VLAN. The claim is now scoped, because left as it was it reads as contradicting the L2 story.

The remaining three are additions:

- The pod MAC address and specific IP address pages documented only the flat form of their annotations. Both now cover the per-interface form. The MAC page also points at KubeVirt's own macAddress field and notes that conflicting values are rejected. The IP address page explains why an L2 workload should use ipAddrs rather than ipAddrsNoIpam.
- The multiple networks page points Multus readers at the L2 path and at the extra installation setting it requires.
- The eBPF use cases page records L2 bridge networking as a consumer of the eBPF data plane, with its higher kernel floor.

Reviewers may want to check the scoping of the BGP statement in particular. I am confident it does not apply to L2, but the sentence has been there a while and someone may have meant it more broadly.

Changed pages:

- https://deploy-preview-2951--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/kubevirt/kubevirt-networking
- https://deploy-preview-2951--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/configuring/pod-mac-address
- https://deploy-preview-2951--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/ipam/use-specific-ip
- https://deploy-preview-2951--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/configuring/multiple-networks
- https://deploy-preview-2951--calico-docs-preview-next.netlify.app/calico-enterprise/next/operations/ebpf/use-cases-ebpf


ctauchen:

PR #2959, opened 2026-08-24
Note which v3 CRD manifest to apply per Kubernetes version

caseydavenport:
The v3 CRD bundle declares its MutatingAdmissionPolicy resources at `admissionregistration.k8s.io/v1`, which Kubernetes only serves from 1.36 onwards. Clusters on 1.34 and 1.35 need the new `v1beta1` bundle instead, so every page that applies a v3 bundle now splits the step by Kubernetes version, with one code block per bundle.

The Helm pages install that bundle from the `projectcalico.org.v3` chart instead. `helm template` uses client-side capabilities unless you pass `--validate`, so those three commands always rendered the beta API version regardless of the cluster. They now pass it.

This also restores the 1.34 prerequisite wording on both vNext `native-v3-crds.mdx` pages. https://github.com/tigera/docs/pull/2805 (v3 CRDs as new-install default) reverted them to "Kubernetes 1.32 or later" while the published 3.32 and 3.23-2 snapshots kept the correct text.

Depends on the manifests being published first, in https://github.com/projectcalico/calico/pull/13619 (publish v1beta1 manifests) and https://github.com/tigera/calico-private/pull/13422 (same, Enterprise).


calico-enterprise/getting-started/upgrading/upgrading-calico-to-calico-enterprise/upgrade-to-tsee/helm.mdx
```diff
@@ -73,6 +74,8 @@ The following steps assume the Calico deployment is installed on `tigera-operato
    kubectl create -f $[filesUrl]/manifests/eck-operator-crds.yaml
    ```
 
+   <V3CRDMapVersion />
```
ctauchen: This seems to be a note to tell Helm users they don't need to do anything. Perhaps we can scrap this and similar ones elsewhere?

calico-enterprise/getting-started/install-on-clusters/windows-calico/rancher.mdx
```diff
@@ -34,6 +36,8 @@ The following steps will outline the installation of $[prodname] networking on t
    kubectl create -f $[filesUrl]/manifests/tigera-operator.yaml
    ```
 
+   <V3CRDMapVersion />
```
ctauchen: To make this easier for users to follow, let's conditionalize the step. So something like this:

1. Install the Tigera Operator and custom resource definitions.

If your cluster is based on Kubernetes 1.36 or later:
code

If your cluster is based on Kubernetes 1.34 or 1.35:
code

ctauchen (commented):
Two suggestions:
1. Make this a regular conditional in the operator/crd apply step. For this step, we don't need to get into the details about why.
2. I don't think we need to mention this at all in Helm installs. Again, we don't need to get into why, and there's nothing that needs doing.

calico-enterprise/getting-started/upgrading/upgrading-calico-to-calico-enterprise/upgrade-to-tsee/helm.mdx
caseydavenport: Yep, done! We we use the manifest to apply the CRDs here, instead of helm itself - so needed to swap to using helm to generate the CRDs instead - but that's a better experience.

calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
```diff
@@ -96,7 +96,7 @@ To install a standard $[prodname] cluster with Helm:
 1. Install the necessary custom resource definitions.
 
    ```bash
-   helm template calico-crds projectcalico.org.v3-$[chart_version_name].tgz | kubectl apply --server-side -f -
+   helm template calico-crds projectcalico.org.v3-$[chart_version_name].tgz --validate | kubectl apply --server-side -f -
```
caseydavenport, line 99: This `--validate` is what triggers helm template to use real cluster state to determine the API version to use.

caseydavenport:
Thanks @ctauchen ! Think this one is ready now

calico-enterprise/getting-started/install-on-clusters/kubernetes/helm.mdx
ctauchen: Ah right, I missed that. Thanks.

ctauchen (commented):
LGTM

PR #2964, opened 2026-08-25
DOCS-3014: Serve Markdown to clients that ask for it

ctauchen:
PR 5 on DOCS-3014, on top of the merged #2963. This is the one with runtime behaviour, so it is deliberately off on the production domain until we have proof from a preview.

Appending .md to a URL only reaches agents that know to try it. Of seven coding agents surveyed in February 2026, Claude Code, Cursor and OpenCode announce Accept: text/markdown but never guess a suffix, so content negotiation is the only way to reach them. The other four announce nothing useful and are unreachable either way.

A Netlify edge function serves the Markdown twin when the client genuinely prefers Markdown. Its declaration carries a header condition, so the function does not run at all for ordinary browser traffic and costs nothing on the overwhelming majority of requests.

Preferring Markdown is not the same as mentioning it. A client may list text/markdown below text/html as a fallback, so quality values are compared rather than the header being substring-matched. Ties go to Markdown, on the grounds that a client naming it and ranking it level with HTML is signalling it can use it.

Scope shrank from the original plan. The edge function was also going to set Content-Type and X-Robots-Tag on .md paths, but production showed both are already correct: Netlify serves .md as text/markdown unprompted, and the /*.md rule in _headers applies the noindex. So this is negotiation only, and this PR adds two files and changes no configuration.

On the production gate. Negotiation is decided from the request hostname, allow-listed rather than deny-listed, so docs.tigera.io stays on HTML and an unrecognised host does not negotiate either. Netlify excludes Accept from Netlify-Vary because of its cardinality; the standard Vary header is documented as feeding cache keys, but if it does not, the failure mode is a browser served raw Markdown out of the CDN, and that is worth proving before it reaches production. Promoting it is one line.

Deciding that in the function rather than in configuration is deliberate. Environment plumbing proved the least reliable part of this: CONTEXT is undefined under netlify dev, which also ignores [context.dev.environment] in netlify.toml and does not forward process environment to edge functions. Keeping the decision in the function means it behaves the same everywhere.

Worth knowing for review, because it cost four failed builds on two sites. Netlify validates the edge function manifest at bundle time and permits only GET, POST, PUT, PATCH, DELETE and OPTIONS. HEAD is not on that list, and including it fails the whole build rather than just the function. I had added HEAD purely so curl -I would exercise the function during testing. There is now a test asserting the declaration only uses methods Netlify accepts, because the mistake is invisible in review and surfaces only at deploy time. Note also that netlify build --dry validates configuration but does not run the bundling step; netlify build --offline does, and reproduces the failure locally.

Consequence: HEAD requests get the HTML page's headers. Agents use GET, so this only affects curl -I probing, which needs curl -sD - -o /dev/null against a GET instead.

Verified against the local emulator: an agent header returns text/markdown with Vary: Accept, a canonical Link header and noindex; a browser header returns HTML; Markdown ranked below HTML returns HTML; a page with no twin falls through to HTML rather than erroring; a category index doc resolves despite its trailing slash; llms.txt is out of scope.

What to check on the deploy preview, which is what local testing cannot cover:

    P=https://deploy-preview-2964--tigera.netlify.app/calico/latest/networking/configuring/bgp
    curl -sD - -o /dev/null -H 'Accept: text/markdown, text/html, */*' $P   # text/markdown, Vary: Accept
    curl -sD - -o /dev/null -H 'Accept: text/html' $P                       # text/html
    curl -sD - -o /dev/null $P                                              # text/html
    # then repeat the last two, twice each, in both orders, watching for a poisoned cache entry

A browser request returning Markdown at any point means Vary: Accept is not being honoured, and negotiation should not be promoted to production.

Tests: 19 unit tests over the Accept parsing, twin-path derivation, host gating and the declaration itself, including the exact headers Claude Code, Cursor and OpenCode send, a real browser header, an explicit q=0, and a media type that merely contains the string "text/markdown".

### DIFF
--- a/netlify/edge-functions/markdown-negotiation.js
+++ b/netlify/edge-functions/markdown-negotiation.js
@@ -1,0 +1,151 @@
+/**
+ * Serve the Markdown twin to clients that ask for Markdown.
+ *
+ * Appending ".md" to a URL only helps agents that know to try it. Several do not:
+ * of seven coding agents surveyed in February 2026, Claude Code, Cursor and OpenCode
+ * announce `Accept: text/markdown` but never guess a suffix, while Codex, Gemini CLI,
+ * Copilot and Windsurf announce nothing useful. Content negotiation is the only way
+ * to reach the first group.
+ *
+ * The `header` condition in the config below means this function does not run at all
+ * for ordinary browser traffic, so it costs nothing on the overwhelming majority of
+ * requests.
+ *
+ * Deliberately does not run on the production domain yet. Netlify excludes `Accept`
+ * from `Netlify-Vary` because of its cardinality, and while the standard `Vary` header
+ * is documented as feeding cache keys, the failure mode if it does not — a browser
+ * served raw Markdown out of the CDN — is bad enough to prove on a preview first.
+ */
+
+/**
+ * Whether to negotiate on this host.
+ *
+ * Allow-listed rather than deny-listed, so an unrecognised host means "do not
+ * negotiate". Netlify preview and branch deploys are served from *.netlify.app;
+ * docs.tigera.io is not, and stays on HTML until the caching behaviour is proven.
+ * Promoting it is a deliberate one-line change here.
+ *
+ * This is decided from the request rather than from an env var because environment
+ * plumbing turned out to be the least reliable part of this: CONTEXT is undefined
+ * under `netlify dev`, which also ignores [context.dev.environment] in netlify.toml.
+ *
+ * @param {URL} url
+ * @returns {boolean}
+ */
+export function negotiationEnabled(url) {
+  return url.hostname.endsWith('.netlify.app') || url.hostname === 'localhost';
+}
+
+/**
+ * Quality value for a media type in an Accept header, or null if absent.
+ *
+ * @param {string} accept
+ * @param {string} mediaType
+ * @returns {number | null}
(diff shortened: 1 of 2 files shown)


ctauchen:
Verification results from the deploy preview, since this is the part local testing cannot cover.

Negotiation works on the real CDN:

    agent header (text/markdown, text/html, */*)  -> text/markdown
    browser header                                -> text/html
    no Accept header                              -> text/html
    markdown ranked below html                    -> text/html

Vary: Accept is honoured. Twelve interleaved requests, agent and browser alternating in both orders, twice each: every one returned the correct representation, with no cache entry leaking across. Netlify merges its own value with ours, so the negotiated response carries "vary: Accept-Encoding, Accept" and also the canonical Link header and X-Robots-Tag: noindex.

Fallbacks behave:

    page with no twin (Swagger)  -> text/html, not an error
    category index doc           -> text/markdown despite the trailing slash
    /llms.txt                    -> text/plain, out of scope entirely

Production is untouched: docs.tigera.io returns text/html for the agent header, as the host allow-list intends.

That is the evidence needed to promote negotiation to production, which is a one-line change to the host check whenever we want it. I would suggest letting this sit on preview for a little while first, since the risk it guards against is a caching behaviour that only shows up under real traffic patterns.

PR #2966, opened 2026-08-25
DOCS-3014: Advertise the Markdown twin on the page itself

ctauchen:
Last PR on DOCS-3014, on top of the merged #2965. Twins have been reachable only by guessing the URL convention or by asking for Markdown in an Accept header. Each doc page now points at its own twin, for machines and for people.

For machines, a link tag in the page head:

    <link rel="alternate" type="text/markdown" href="/calico/latest/networking/configuring/bgp.md" />

so a crawler or agent holding the HTML can discover the Markdown without guessing. The href is relative on purpose: siteConfig.url is the production domain, so an absolute one would point every deploy preview at docs.tigera.io.

For people, a split button sitting on the page title's baseline. Copying is the common case — it is what someone pasting a page into an assistant wants — so it is the primary target, and the chevron opens a menu with the same copy action plus a link to view the Markdown. The whole control is hidden below 996px, where the reading column has no room to spare. This is the only user-visible change in the entire stack, so it is worth looking at on the preview rather than only in the diff.

Neither half renders for a page without a twin, so nothing advertises a 404. Those are declared as path suffixes in customFields.markdownTwinExclusions, one entry covering the same page across every product and version: the eight client-rendered Swagger API browsers, and the placeholder page the multi-instance 'default' docs instance requires. The generator checks that list against reality on every build and warns if the two disagree in either direction. It checks every docs instance rather than only the ones it generates for, and that distinction is what found the /docs placeholder: the theme renders on any docs page, including instances the plugin skips.

Three things came out of review and are worth calling out.

Copying was broken in Safari. Awaiting a fetch before calling navigator.clipboard.writeText forfeits the user gesture WebKit requires, so the write rejected with NotAllowedError — and the window.open fallback was popup-blocked for the same reason, so the button did nothing at all. The fetch now stays inside a ClipboardItem, which is the sanctioned way to keep the gesture. Confirmed by driving WebKit directly: the old code raises NotAllowedError there, the new code succeeds. The earlier verification on this PR was Chromium-only, which is how it was missed.

The menu is Chakra's rather than hand-rolled. Chakra already wraps every page via src/theme/Root.js, and its Menu brings the keyboard contract that role=menu promises: arrow keys move between items, opening from the keyboard focuses the first, Escape closes and returns focus to the trigger. The hand-rolled version declared role=menu and implemented none of that, which puts NVDA and JAWS into application mode where arrow keys then do nothing — worse than declaring no role at all. It is styled from Infima variables rather than the repo's Chakra Menu theme: that theme hard-codes color: tigeraBlack on the list with no colour-mode branch, and nothing else in the repo renders a Chakra Menu, so it is untested. Without an explicit colour the item titles come out dark on dark.

Icons are Docusaurus's own @theme/Icon/Copy, Success and ExternalLink rather than hand-drawn SVGs. They are already on the page — the code block copy buttons use them — so the reader meets the same glyph for the same action.

The CSS deliberately does not build on Infima's .dropdown__* classes. src/css/custom.css restyles those globally for the navbar, including a blue hover with white text and a grey panel in dark mode, so adopting them here would mean overriding all of it back: more CSS, not less.

Also stops the Algolia crawler indexing twins, which has to land with this rather than after it, since this is the first thing to link to them. The pattern is anchored, "\.md$" rather than ".md$", because stop_urls are regular expressions and an unescaped dot would also match a route ending in "amd".

Needs a human step: nothing in the repo reads algolia-crawler-config.json — it is a checked-in copy of what goes in the crawler console. The stop_urls entry needs applying there too, ideally before this merges.

Not needed, in case it comes up: Google is already handled. The /*.md X-Robots-Tag: noindex rule went in with #2958 and is live in production today.

Verified against a full build: 2,921 pages advertise a twin and none of those links is dangling; excluded pages advertise nothing; removing an entry from the exclusion list makes the build warn, naming the page. Driven in a browser: the clipboard receives the page's own twin in both Chromium and WebKit, keyboard navigation behaves, and the control sits on the h1 baseline in both colour modes.

On the deploy preview:

- https://deploy-preview-2966--tigera.netlify.app/calico/latest/networking/configuring/bgp — control on the title line, rel=alternate in the head
- https://deploy-preview-2966--tigera.netlify.app/calico/latest/networking/ — a category index doc, whose twin drops the trailing slash
- https://deploy-preview-2966--tigera.netlify.app/calico/latest/reference/rest-api-reference — no control and no link tag, since it has no twin

Worth checking the menu in both colour modes and with the keyboard, and that Copy page lands on your clipboard — in Safari as well as Chrome, given the above.

With this merged, DOCS-3014 is complete. The remaining decision is whether to promote content negotiation to production, which is one line in the edge function's host check and better made after the preview has seen real traffic.

### DIFF
--- a/algolia-crawler-config.json
+++ b/algolia-crawler-config.json
@@ -1,7 +1,7 @@
 {
   "index_name": "calico",
   "start_urls": ["https://docs.tigera.io/calico", "https://docs.tigera.io/calico-enterprise", "https://docs.tigera.io/calico-cloud"],
-  "stop_urls": [],
+  "stop_urls": ["\\.md$"],
   "selectors": {
     "lvl0": {
       "selector": "(//ul[contains(@class,'menu__list')]//a[contains(@class, 'menu__link menu__link--sublist menu__link--active')]/text() | //nav[contains(@class, 'navbar')]//a[contains(@class, 'navbar__link--active')]/text())[last()]",
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -586,6 +586,18 @@ export default async function createAsyncConfig() {
     ],
     customFields: {
       isTesting: process.env.TESTING || false,
+      // Doc routes that render entirely client-side, so the generator produces no
+      // Markdown twin for them and nothing should link to one. Matched as path
+      // suffixes, so a single entry covers the page across every product and
+      // version. The llms-txt plugin checks this against the pages it actually
+      // skipped and warns on drift, so the list cannot silently go stale.
+      markdownTwinExclusions: [
+        // Swagger API browsers: rendered client-side, so they convert to nothing.
+        '/reference/rest-api-reference',
+        // The placeholder page the multi-instance 'default' docs instance requires.
+        // No Markdown is generated for that instance at all.
+        '/docs',
+      ],
     },
   };
 
--- a/src/components/MarkdownActions/__test__/markdownTwin.test.js
+++ b/src/components/MarkdownActions/__test__/markdownTwin.test.js
@@ -1,0 +1,50 @@
+import { hasMarkdownTwin, twinPathFor } from '../../../utils/markdownTwin';
+
+describe('twinPathFor', () => {
+  it('appends .md to a leaf doc route', () => {
+    expect(twinPathFor('/calico/latest/networking/configuring/bgp')).toBe(
+      '/calico/latest/networking/configuring/bgp.md'
+    );
+  });
+
+  it('drops the trailing slash a category index doc carries', () => {
+    // Its canonical URL ends in a slash; the twin is a sibling file, not a child.
+    expect(twinPathFor('/calico/latest/networking/')).toBe('/calico/latest/networking.md');
+  });
+});
+
+describe('hasMarkdownTwin', () => {
+  const EXCLUSIONS = ['/reference/rest-api-reference'];
+
+  it('is true for an ordinary doc page', () => {
+    expect(hasMarkdownTwin('/calico/latest/about', EXCLUSIONS)).toBe(true);
+  });
+
+  it('is false for a client-rendered page, in every product and version', () => {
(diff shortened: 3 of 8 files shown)


ctauchen:
Went through the Copilot comments. Most are against commits superseded by the Chakra rewrite, but two were live and chasing one of them turned up a third bug that had been hiding.

Fixed — fetch does not reject on HTTP errors. Correct, and a regression I introduced: the pre-Chakra version checked response.ok and I dropped the check when rewriting for ClipboardItem. A 404 would have put an HTML error page on the clipboard and reported success. Now verified by stubbing a 404: the clipboard is left untouched and the button does not say "Copied".

Fixed — the reset timeout was never cleared. Also correct. It is now held in a ref, cleared before each new one and on unmount, so a quick second click or a navigation inside the two seconds cannot leave a timer setting state behind it.

Found while fixing the first one, and the more serious of the three. The twin is served as text/markdown, and Chromium rejects a ClipboardItem whose key does not match its blob's type — so the ClipboardItem path had been failing in Chromium the whole time. It went unnoticed because the old code's catch silently rescued it through the writeText path; my refactor removed that rescue and exposed it. My earlier note that copying was verified in Chromium and WebKit was true, but in Chromium it was true for the wrong reason. The blob is now re-wrapped as text/plain, and both engines are verified independently.

Declined — 'Copy page' is ambiguous, suggest 'Copy as Markdown'. The visible label matches the reference design this control was asked to follow, and the button's accessible name is already the explicit "Copy page as Markdown", so screen reader users get the unambiguous version. Happy to change it if others prefer the longer label.

Already addressed by the rewrite, listed for completeness: role='menu' without the keyboard contract, aria-haspopup='menu', mousedown versus pointerdown for click-away, and the popup-blocked window.open fallback. Chakra's Menu now provides the keyboard behaviour and the dismissal handling, and the window.open fallback is gone — it was blocked after an await for the same reason the clipboard write was, which is what made the button do nothing in Safari.

PR #2968, opened 2026-08-26
DOCS-3021: Render the technology preview table from the data file

ctauchen:
First iteration on DOCS-3021. data/feature-status.yaml landed in DOCS-3009 with no consumer, so the same statuses are now maintained twice: once in the data file and once as a hand-typed Markdown table in each version's release notes. Seven tables across two products, retyped every release. They have already drifted — the Enterprise 3.24 table is missing the L2 bridge networking row that the data file records.

This adds the machinery and uses it in one place, the Calico Open Source 3.32 technology preview table.

A build-time plugin, src/plugins/docusaurus-plugin-feature-status, parses the YAML once and puts it on plugin global data. Parsing there rather than in a webpack loader keeps the file out of every page bundle that does not use it, and getPathsToWatch means editing a status during yarn start updates the tables without a restart.

The component is src/components/FeatureStatusTable, registered in MDXComponents so pages need no import. It takes no arguments:

    <TechPreviewTable />

The product and the three-release window both come from the page's own docs context. The product is the docs plugin id, which is already the data file's product key. The window ends at the page's own version, with the Enterprise suffix stripped so 3.24-1 and 3.24-2 both resolve to line 3.24. That means a frozen snapshot keeps the window it was cut with — the 3.31 page still shows 3.29 to 3.31 after 3.33 ships — and no version numbers are written into the page.

The Next version has no release line to derive a window from. Rather than guess at the next minor, the component renders nothing and warns. The Next release notes already have an empty preview section, so nothing regresses there.

Row order is the one thing that could not be taken from the data. The hand-written order was not derivable and the two products disagreed: Open Source put the oldest preview last and Native v3 CRDs before Istio ambient mode, Enterprise did the reverse. Rows are now ordered by the release a feature first appeared in, oldest first, with ties keeping data-file order. The only visible consequence in 3.32 is that nftables data plane moves from the third row to the first.

Cell values, the legend, and row inclusion are unchanged. The legend still glosses only the statuses a given table actually uses, so a table with no GA cells does not explain what GA means — that is not decoration, it is what the Enterprise 3.21 table does today.

Verified against a full build. The Markdown twin of the 3.32 page, generated from the rendered HTML, reads:

    | Feature                             | 3.30 | 3.31 | 3.32 |
    | ----------------------------------- | ---- | ---- | ---- |
    | nftables data plane                 | TP   | GA   | GA   |
    | Calico Ingress Gateway              | TP   | GA   | GA   |
    | Flow logs API and Whisker           | TP   | TP   | TP   |
    | Istio ambient mode                  | –    | –    | TP   |
    | Native v3 CRDs                      | –    | –    | TP   |
    | Selector-scoped Felix configuration | –    | –    | TP   |

    TP = technology preview, GA = generally available, – = not available in that release.

Nineteen unit tests cover the logic module and the component, including a case that asserts the whole 3.32 table and one that pins the Enterprise 3.24 ordering. The logic lives in its own module so the carry-forward and ordering rules are testable without rendering.

Scoped deliberately to what renders this one table. There is no props interface, no configurable column count, and no status filter — those arrive with the second caller that needs them, not before.

Left for follow-ups:

- the deprecated and removed variant, which needs a status filter parameter that this table does not
- the other six tables, once the ordering change has been reviewed once
- whether the Next release notes should carry the table at all

On the deploy preview:

- https://deploy-preview-2968--tigera.netlify.app/calico/latest/release-notes/ — the technology preview table, rendered by the component. The deprecated and removed table below it is still hand-written Markdown, and the two should be indistinguishable.

### DIFF
--- a/calico_versioned_docs/version-3.32/release-notes/index.mdx
+++ b/calico_versioned_docs/version-3.32/release-notes/index.mdx
@@ -4,6 +4,8 @@ title: Calico Open Source release notes
 sidebar_label: Release notes
 ---
 
+import TechPreviewTable from '@site/src/components/FeatureStatusTable';
+
 # Calico Open Source 3.32 release notes
 
 Learn about the new features, bug fixes, and other updates in this release of $[prodname].
@@ -88,16 +90,7 @@ This release adds support for Kubernetes 1.36.
 This table shows the status of features that were in technology preview at any point in the releases covered here.
 Technology preview features are for evaluation and feedback only and are **not supported for production use**. Behavior, APIs, and configuration may change before a feature reaches general availability.
 
-| Feature | 3.30 | 3.31 | 3.32 |
-| --- | --- | --- | --- |
-| Flow logs API and Whisker | TP | TP | TP |
-| Calico Ingress Gateway | TP | GA | GA |
-| nftables data plane | TP | GA | GA |
-| Native v3 CRDs | – | – | TP |
-| Istio ambient mode | – | – | TP |
-| Selector-scoped Felix configuration | – | – | TP |
-
-TP = technology preview, GA = generally available, – = not available in that release.
+<TechPreviewTable />
 
 ## Deprecated and removed features
 
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -548,6 +548,7 @@ export default async function createAsyncConfig() {
           editUrl: generateEditUrl,
         },
       ],
+      './src/plugins/docusaurus-plugin-feature-status',
       [
         './src/plugins/docusaurus-plugin-llms-txt',
         {
--- a/src/components/FeatureStatusTable/__test__/featureStatus.test.ts
+++ b/src/components/FeatureStatusTable/__test__/featureStatus.test.ts
@@ -1,0 +1,113 @@
+import fs from 'fs';
+import path from 'path';
+import { parse } from 'yaml';
+
+import { buildLegend, buildRows, cellLabel, releaseWindow } from '../featureStatus';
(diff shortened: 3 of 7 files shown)


ctauchen:
Responding to the Copilot review, since one of its points is wrong and I do not want it to mislead a human reviewer.

The claim that useDocsVersion() returns only version metadata and not pluginId is incorrect. It returns PropVersionMetadata, declared in @docusaurus/plugin-content-docs as:

    export type PropVersionMetadata = Pick<
      VersionMetadata,
      'label' | 'banner' | 'badge' | 'className' | 'isLast' | 'noIndex'
    > & {
      /** ID of the docs plugin this version belongs to. */
      pluginId: string;
      /** Name of this version. */
      version: string;
      ...
    };

Both fields are there, so useActivePlugin is not needed. The build is independent confirmation: the 3.32 page renders columns 3.30, 3.31, 3.32 against product calico, neither of which is possible if either field were undefined. Nothing throws in that case either — releaseWindow would simply fail to match and the component would render nothing.

Two points from the review were fair and are now applied:

- Terminology. The comments called the unversioned version "next", which is the sidebar label, not the version name. They now say current and explain that the site labels it Next, which matches what the code compares against.
- Calico Cloud versions. The earlier releaseLine() would have turned 23-2 into 23. That function no longer exists; releaseWindow requires a dotted major.minor, so a Cloud version does not match and the component renders nothing and warns, the same safe path as current. Added a test pinning that.

Also removed the MDXComponents registration. The component is imported directly in the release notes now, rather than being available site-wide to every page.

PR #2980, opened 2026-08-28
DOCS-3021: Render the deprecated and removed tables from the data file

ctauchen:
Sits on top of #2979 and will shrink to a single commit once that merges. Review the top commit only.

Every release-notes page carries a second hand-typed table, listing what has been deprecated or removed, with the same duplication and drift problem as the preview table. This renders it from data/feature-status.yaml too. With this and #2979 merged, no feature status is typed into a release-notes page anywhere.

Almost nothing new was needed. buildRows gains an include parameter, and the status filter is the only thing separating the two tables: one selects rows that were in preview during the window, the other rows that were deprecated or removed. The window, the product, carry-forward, ordering, and the legend are already shared. That parameter is the one deliberately left out of #2968 as speculative, on the grounds that it should arrive with a second caller rather than before one. This is the second caller.

Both statuses select a row, so a feature deprecated earlier and removed inside the window stays visible through the release that removed it. Enterprise 3.24 is the case that matters: compliance reporting reads Deprecated, Deprecated, Removed.

Both components are now named exports, so a page imports the pair on one line rather than a default and a name:

    import { TechPreviewTable, DeprecatedFeaturesTable } from '@site/src/components/FeatureStatusTable';

Only the table and its legend are replaced. The prose above each table and the per-release list of deprecations below it are untouched.

Verified the same way as #2979, against what docs.tigera.io serves today. Across all seven deprecated tables there is no difference in any column header, in any row, or in any cell. Two things change:

- Row order, on the four Enterprise pages, by the same first-appearance rule as everywhere else. The Open Source tables happen to already be in that order.
- Every deprecated legend gains a gloss for the dash. The legend set is fixed per table, so the deprecation table always names generally available, scheduled for removal, no longer present, and not available in that release. No table published today has a dash cell, but a feature introduced and deprecated inside the same window would produce one, and the reader should not meet a symbol the legend does not name.

The tables already carried the same legend regardless of content, so this keeps them consistent with each other and with the preview tables, which #2979 puts on the same footing.

On the deploy preview:

- https://deploy-preview-2980--tigera.netlify.app/calico/latest/release-notes/ — Open Source, both tables on one page
- https://deploy-preview-2980--tigera.netlify.app/calico-enterprise/3.24/release-notes/ — Enterprise 3.24, the widest tables and the only ones with a removed feature
- https://deploy-preview-2980--tigera.netlify.app/calico-enterprise/3.22/release-notes/ — Enterprise 3.22, whose deprecated table is the one that loses two legend glosses

The rendered tables should be indistinguishable from hand-written Markdown ones, since they are plain tables picking up the same theme styles.

### DIFF
--- a/calico-enterprise_versioned_docs/version-3.21-2/release-notes/index.mdx
+++ b/calico-enterprise_versioned_docs/version-3.21-2/release-notes/index.mdx
@@ -4,7 +4,7 @@ title: Calico Enterprise release notes
 sidebar_label: Release notes
 ---
 
-import TechPreviewTable from '@site/src/components/FeatureStatusTable';
+import { TechPreviewTable, DeprecatedFeaturesTable } from '@site/src/components/FeatureStatusTable';
 
 # Calico Enterprise 3.21 release notes
 
@@ -83,12 +83,7 @@ Review the following table to see what features have recently been deprecated or
 Deprecated features are still present and supported, but they are scheduled for removal.
 You should consider migrating away from a deprecated feature before it is removed.
 
-| Feature | 3.19 | 3.20 | 3.21 |
-| --- | --- | --- | --- |
-| Compliance reporting | GA | Deprecated | Deprecated |
-| Fortinet integration | Deprecated | Deprecated | Deprecated |
-
-GA = generally available, Deprecated = scheduled for removal, Removed = no longer present.
+<DeprecatedFeaturesTable />
 
 ## Release details
 
--- a/calico-enterprise_versioned_docs/version-3.22-2/release-notes/index.mdx
+++ b/calico-enterprise_versioned_docs/version-3.22-2/release-notes/index.mdx
@@ -4,7 +4,7 @@ title: Calico Enterprise release notes
 sidebar_label: Release notes
 ---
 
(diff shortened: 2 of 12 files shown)


ctauchen:
The readonly finding was correct and is fixed: buildRows now takes readonly FeatureStatus[], since it does not mutate the argument.

Worth recording how it survived to review, because the same gap will hide the next one. Nothing in the pipeline type-checks this code. Jest transforms through swc, and the site builds through swc as well, so both strip types without checking them. Running the repo's own tsc fails before it reaches any source, on two pre-existing tsconfig problems: @types/node is listed in types but not installed, and suppressImplicitAnyIndexErrors was removed in TypeScript 5.

Verified the fix rather than assuming it. Type-checking the two files against a minimal config reports the error before the change and nothing after:

    src/components/FeatureStatusTable/index.tsx(35,56): error TS2345:
      Argument of type 'readonly FeatureStatus[]' is not assignable to
      parameter of type 'FeatureStatus[]'.

Fixing the repo tsconfig so CI can type-check at all is worth its own ticket. It is out of scope here, but this PR is evidence that a type error can reach review with every check green.

PR #2987, opened 2026-08-31
docs: serve Forklift manifests from downloads.tigera.io

radTuti:
The Forklift install manifests were served from `docs.tigera.io/files`, whose path carries no version, so the next Forklift release would overwrite the 3.24 files in place. This moves them to the release bucket and points the docs at `$[filesUrl]`, which the release process stamps per version. The manifests in the previous PR also named image tags that were never pushed; the copies now in the bucket carry the tags this release actually shipped.

Draft until the manifests are live at `downloads.tigera.io/ee/v3.24.0-2.0/forklift/` — merging before that would leave the install pages pointing at missing files. Note `filesUrl` on main is `ee/master`, so these links resolve only in the versioned 3.24 docs.


src/utils/linkChecker.js
```diff
@@ -27,6 +27,9 @@ const defaultSkipList = [
   // debs directory listing 403s (index forbidden) even when the packages exist.
   /^https?:\/\/downloads\.tigera\.io\/ee\/debs\/v3\.23\/?$/,
   /^https?:\/\/docs\.tigera\.io\/calico\/latest\/networking\/kubevirt\/?$/,
+  // Forklift manifests: filesUrl on main points at ee/master, which the release
+  // process only populates when a version is cut. Live in the versioned docs.
+  /^https?:\/\/downloads\.tigera\.io\/ee\/master\/forklift\/.*/,
```
ctauchen, line 32: Don't we want this to fail when the files aren't available?

ctauchen (approved):

src/utils/linkChecker.js
radTuti: only if it is not master.

PR #2995, opened 2026-09-02
Correct the Forklift install pages after the RKE2 test run

ctauchen:
A test run of the Forklift v2.12.5-v3.24.0-2.0 release build on a five-node RKE2 cluster (CE 3.24-2, BPF, KubeVirt) found five problems in the install pages. Migration itself worked end to end. Each problem was true on a uniform, well-provisioned cluster and wrong or unusable on a real one, which matters because this non-OLM route exists for RKE2 customers.

Mixed-cluster limitation, step 4 of the Kubernetes page. The four virt_v2v profile fields belong to the one ForkliftController, so they apply to every conversion pod in the cluster, and there is no per-node or per-Plan override. The page said to set only the seccomp pair "on nodes that do not enforce AppArmor", which quietly assumes every node is the same. A cluster whose conversion-capable nodes are mixed has no valid setting: naming the AppArmor profile makes the kubelet refuse the conversion pod where AppArmor is absent, and omitting it lets the pod start where AppArmor is enforced, and the remount then fails. The page now gives the two valid configurations and points to convertorNodeSelector on the Plan for mixed clusters. Both configurations were verified separately on the test cluster.

Storage prerequisite, Before you begin. The pages asked only for CDI and a StorageClass. Migrations also fail before conversion starts when the volume cannot honor the file locking qemu-img takes on an image it opens, which NFS-backed storage needs mount options for. This cost a day on the release candidate.

Address-preservation annotations. The verification step named only the interface-scoped keys. A primary NIC, mapped by a type: pod item with a calico block, carries the keys with no interface name instead, so a reader following the page concluded that preservation had failed when it had worked. Both forms are real, and the page now shows one per case, with a pointer to the existing guidance on choosing the interface name.

Profile installer log, step 4 of the Kubernetes page. The page promised a log that "says which it did on each node", then gave a command that aggregates every pod into unlabelled lines. On a mixed cluster that is exactly the information needed for the step that follows. It now reads one pod at a time, under each node's name.

The calico struct, step 5. Its fields and a complete type: multus item were shown, but never a complete item containing calico, which left its place in the YAML ambiguous. One complete map, with a type: pod item and a type: multus item, now shows calico as a property of destination, matching the CRD.

Both the 3.24-2 and the unversioned copies of both pages are updated, since nothing in CI compares them.

Not in this PR, and raised separately: a sixth issue in the report, the ForkliftController never reaching Successful on Kubernetes, needs no doc change. It was fixed in tigera/forklift #2 and #3, and the published images were re-cut from that fix, so there is nothing to document around.

Changed pages on the deploy preview, 3.24-2:

https://deploy-preview-2995--tigera.netlify.app/calico-enterprise/3.24/networking/kubevirt/install-forklift-kubernetes

https://deploy-preview-2995--tigera.netlify.app/calico-enterprise/3.24/networking/kubevirt/install-forklift-openshift

The unversioned copies hold the same content and build on the next preview:

https://deploy-preview-2995--calico-docs-preview-next.netlify.app/calico-enterprise/next/networking/kubevirt/install-forklift-kubernetes

### DIFF
--- a/calico-enterprise/networking/kubevirt/install-forklift-kubernetes.mdx
+++ b/calico-enterprise/networking/kubevirt/install-forklift-kubernetes.mdx
@@ -43,6 +43,9 @@ Upstream Forklift's guest conversion fails on clusters whose nodes carry the run
   ```
 
 - CDI and a default StorageClass, for migrations that import disks. Neither is needed to install Forklift.
+  The StorageClass backing the migrated disks must support the file locking `qemu-img` takes on an image it opens.
+  A volume that cannot honor those locks fails the migration before conversion starts.
+  NFS-backed storage needs mount options that permit locking, such as `nfsvers=3,nolock` or NFSv4, set on the StorageClass rather than only on the export.
 
 ## How to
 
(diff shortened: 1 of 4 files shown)


ctauchen:
PTAL @lwr20

PR #3001, opened 2026-09-02
EV-6699: Document license agent removal in 3.24.0-3

ti-afra:
The license agent is removed in Calico Enterprise 3.24.0-3.0 and is no longer maintained. No license-agent image is published for this release or later.

- Release notes: add a "Deprecated and removed features" entry, plus an "Upgrade notes" subsection pointing to the platform-specific cleanup steps.
- Upgrade guides (helm, operator, openshift): document how to uninstall the agent, preferring the versioned manifest from the release being upgraded from, with a manual resource cleanup fallback. Each links to the license metrics documentation.
- Redirects: point the removed license-agent metrics page at the license options page for the latest and 3.24 paths. These are non-forced 301s, so they stay dormant while 3.23 remains latest and activate once 3.24.0-3.0 ships.

### DIFF
--- a/calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
+++ b/calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
@@ -54,6 +54,28 @@ owned resources being garbage collected by Kubernetes.
 
 $[prodname] creates a default-deny for the calico-system namespace. If you deploy workloads into the calico-system namespace, you must create policy that allows the required traffic for your workloads prior to upgrade.
 
+### License agent removed
+
+The license agent has been removed in $[prodname] v3.24.0-3. If you previously installed
+it, uninstall it using the manifest from the version you are upgrading
+**from** — it is still available at its versioned URL. For example, if upgrading from v3.23.2:
+
+  ```bash
+  kubectl delete -f https://downloads.tigera.io/ee/v3.23.2/manifests/licenseagent.yaml
+  ```
+
+If you no longer have that manifest:
+
+  ```bash
+  kubectl delete servicemonitor calico-lic-monitor -n tigera-prometheus --ignore-not-found
+  kubectl delete networkpolicies.projectcalico.org calico-system.tigera-license-agent-access -n tigera-license-agent --ignore-not-found
+  kubectl delete networkpolicies.projectcalico.org allow-tigera.tigera-license-agent-access -n tigera-license-agent --ignore-not-found
+  kubectl delete clusterrole,clusterrolebinding tigera-license-agent --ignore-not-found
+  kubectl delete namespace tigera-license-agent --ignore-not-found
+  ```
+
+To learn about the license metrics, see [Operator metrics](../../../../operations/monitor/metrics/operator-metrics.mdx).
+
 ## Upgrade from 3.19 or later
 
 :::note
--- a/calico-enterprise/release-notes/index.mdx
+++ b/calico-enterprise/release-notes/index.mdx
@@ -30,6 +30,7 @@ This version of Calico Enterprise is based on [Calico Open Source $[openSourceVe
 
 - The compliance reporting feature has been removed from the $[prodname] web console.
 - $[prodname] is moving the `projectcalico.org/v3` API from the aggregated API server to native Kubernetes CRDs. Both mechanisms will be supported until native v3 CRDs are compatible with all supported platforms, after which the aggregated API server will be removed.
+- The license agent has been removed and is no longer maintained.
 
 ## Technology Preview features
 
@@ -52,6 +53,14 @@ Migrate in this order:
 3. After the upgrade, re-point monitoring, RBAC, and external DNS from `tigera-gateway` to each Gateway's namespace. If you used merged gateways, plan for one load balancer, DNS record, and certificate per gateway.
 
 :::
+
+### License agent removed
+
+If you previously installed the license agent, remove it from your cluster before or after upgrading.
+No `license-agent` image is published for this release or later.
+For platform-specific steps, see the upgrade guides for [Helm](../getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx#license-agent-removed),
+[the operator](../getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/operator.mdx#license-agent-removed),
+or [OpenShift](../getting-started/upgrading/upgrading-enterprise/openshift-upgrade.mdx#license-agent-removed).
 
 ## Release details
 
--- a/calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/operator.mdx
+++ b/calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/operator.mdx
@@ -77,6 +77,28 @@ These steps differ based on your cluster type. If you are unsure of your cluster
 
 :::
 
+### License agent removed
+
+The license agent has been removed in $[prodname] v3.24.0-3. If you previously installed
+it, uninstall it using the manifest from the version you are upgrading
+**from** — it is still available at its versioned URL. For example, if upgrading from v3.23.2:
+
+  ```bash
+  kubectl delete -f https://downloads.tigera.io/ee/v3.23.2/manifests/licenseagent.yaml
+  ```
+
+If you no longer have that manifest:
+
+  ```bash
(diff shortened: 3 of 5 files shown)


calico-enterprise/release-notes/index.mdx
```diff
@@ -53,6 +54,12 @@ Migrate in this order:
 
 :::
 
+### License agent
```
rene-dekker: Maybe clarify i the title that it has been removed.

In the paragraph, please direct them towards the operator metrics in `calico-enterprise/operations/monitor/metrics/operator-metrics.mdx`

ti-afra: changed the title to `License agent removed` and updated the link to point to operator metrics instead of license metrics.

rene-dekker (approved):

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
```diff
@@ -54,6 +54,28 @@ owned resources being garbage collected by Kubernetes.
...
+
+The license agent has been removed in $[prodname] v3.24.0-3. If you previously installed
+it, uninstall it using the manifest from the version you are upgrading
+**from** — it is still available at its versioned URL. For example, if upgrading from v3.23.2:
+
+  ```bash
+  kubectl delete -f https://downloads.tigera.io/ee/v3.23.2/manifests/licenseagent.yaml
+  ```
+
+If you no longer have that manifest:
```
ctauchen, line 67: This seems to be a second, perhaps unnecessary method. We know that we host manifests for all published versions. So can't we just rely on the manifest being available?

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
```diff
@@ -54,6 +54,28 @@ owned resources being garbage collected by Kubernetes.
...
 
+### License agent removed
+
+The license agent has been removed in $[prodname] v3.24.0-3. If you previously installed
+it, uninstall it using the manifest from the version you are upgrading
+**from** — it is still available at its versioned URL. For example, if upgrading from v3.23.2:
+
+  ```bash
+  kubectl delete -f https://downloads.tigera.io/ee/v3.23.2/manifests/licenseagent.yaml
+  ```
```
ctauchen, line 65: What does this procedure look like for Helm installations?

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
```diff
@@ -54,6 +54,28 @@ owned resources being garbage collected by Kubernetes.
 
 $[prodname] creates a default-deny for the calico-system namespace. If you deploy workloads into the calico-system namespace, you must create policy that allows the required traffic for your workloads prior to upgrade.
 
+### License agent removed
+
+The license agent has been removed in $[prodname] v3.24.0-3. If you previously installed
```
ctauchen, line 59: > If you previously installed it

Under what circumstances would a CE users NOT have installed it? It seems to me that this applies to everyone using Calico Enterprise. Or is there a way to have a license without the agent?

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
```diff
@@ -54,6 +54,28 @@ owned resources being garbage collected by Kubernetes.
 
 $[prodname] creates a default-deny for the calico-system namespace. If you deploy workloads into the calico-system namespace, you must create policy that allows the required traffic for your workloads prior to upgrade.
 
+### License agent removed
```
ctauchen, line 57: Some general questions:
1) What happens if you don't remove the license agent?
2) Does it make a difference when I do this? Before or after an upgrade? I'm guessing we do this first, then perform the upgrade. 
3) I presume that this removal step needs to remain in the CE docs until 3.27, when it will no longer apply because we will stop supporting upgrades from 3.24. Is that right?
4) Is this just cleanup, or does this constitute a breaking change?

ctauchen (commented):
Thanks for this. I have a few questions and suggestions. 

I'm making changes to how we document these upgrade-related procedures (mostly for breaking changes, but also this). I'm still working through the structure, but it's likely to result in an "Upgrade notes" page that includes all the things people need to consider before upgrading.

If it's all right with you, I'll repurpose the material here and include it with that work. After I check details about the license agent changes.

ctauchen:
@ti-afra One last check: Is there anything that a user needs to do to maintain the function of the license? I assume that we're just changing the system that deals with licenses, and that everything should continue to work as before with no special procedures. Is that right?

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
ti-afra: fair enough. I tried to be helpful to user but if it is too much explanation, I am okay with dropping this section and only refer them to find the correct manifest to do the uninstall.

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
ti-afra: installing license agent is not part of the default installation, it comes as an additional step and therefore not included in the helm installations as well. the instructions to install license agent for reference: https://docs.tigera.io/calico-enterprise/3.21/operations/monitor/metrics/license-agent

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
ti-afra: The license agent has no operational role in the cluster. It runs as a standalone Deployment whose only job is to read the LicenseKey and expose Prometheus gauges — it does not gate, enforce or enable any feature. A Calico Enterprise cluster without it is fully functional; the only difference is those metrics.

Its replacement is already running on every cluster installed from our manifests. tigera-operator exports `tigera_operator_license_expiry_timestamp_seconds` and `tigera_operator_license_valid`, both labelled by package, and ships PrometheusRule alerts on them — `LicenseExpiringWarning` at 30 days and `LicenseExpiringCritical` at 7 days or when the license is invalid. The agent shipped no alerts, so this is more coverage than before, not less.

The one real impact is on users whose dashboards or alerts reference the agent's metric names; those need repointing at the operator metrics.

calico-enterprise/getting-started/upgrading/upgrading-enterprise/kubernetes-upgrade-tsee/helm.mdx
ti-afra: **1. What happens if you don't remove it?**

It keeps running and keeps exporting its metrics. Nothing in the release manages it — the operator doesn't reconcile it, nothing upgrades or removes it — so it becomes an orphan: not built, not tested, not patched. CVEs accumulate in an image that never gets rebuilt.

**2. Before or after the upgrade?**

Either works.

**3. Keep the note until 3.27?**

That's correct. When 3.24 is the oldest supported cluster, we are not supporting this upgrade path anymore.

**4. Cleanup or breaking change?**

No functionality breaks — license validity is still observable via the operator metrics. But it's a user-visible removal, not silent cleanup: the manifest URL will 404, and dashboards or alerts on the agent's metric names need repointing. Should be a deprecation/removal in the release notes.

ti-afra:
> @ti-afra One last check: Is there anything that a user needs to do to maintain the function of the license? I assume that we're just changing the system that deals with licenses, and that everything should continue to work as before with no special procedures. Is that right?

Correct — no user action needed. Licensing itself is unchanged: the LicenseKey resource, enforcement, and renewal all work exactly as before. The license agent never took part in any of that; it only exported metrics. The only change is which metrics report license expiry.

ti-afra:
> Thanks for this. I have a few questions and suggestions.
> 
> I'm making changes to how we document these upgrade-related procedures (mostly for breaking changes, but also this). I'm still working through the structure, but it's likely to result in an "Upgrade notes" page that includes all the things people need to consider before upgrading.
> 
> If it's all right with you, I'll repurpose the material here and include it with that work. After I check details about the license agent changes.

Of course — please structure it however works best for the docs.

One clarification: are you taking everything from this PR, or just the release note and upgrade note? And will you close this PR when that's merged?

PR #3005, opened 2026-09-03
DOCS-3027: Remove unreferenced images from static/img

ctauchen:
Removes 105 files from static/img that no page, component, stylesheet, or config references. That is 34 MB, and static/img drops from 108 MB to 74 MB.

Jira: https://tigera.atlassian.net/browse/DOCS-3027
Most of these date from DOCS-1206, which copied images into every product tree regardless of which product used them.

How it was verified:

- Matched all 446 files in static/img against every md, mdx, js, jsx, ts, tsx, json, css, svg, yaml, html and txt file in the repo, then confirmed with git grep that none of the candidates are referenced on any of the 12 active upstream branches either.
- Checked for consumers outside this repository, since files under static/img are served at stable public URLs. A number of our logos are used elsewhere. None of them is deleted here, and one file was restored to this branch after that check.
- Full local build before and after. The build stays clean, the set of image URLs referenced by the built site is unchanged, and the count of webpack-emitted images under assets/images stays at 177.
- Loaded 107 pages from the local build in Chromium and checked every img element for load failures: 593 images, no failures. The 107 pages cover all 39 pages that reference images through the unchecked src="/img/..." form, the 30 most image-heavy pages, and the pages nearest the deleted files. A control page referencing a deleted image was correctly reported as broken.

Nothing to look at on the deploy preview, since no page changed.

Held back for a follow-up:

- 34 brand, product, and customer logos are unreferenced by the docs but stay for now. Some are known to be in use outside this repository, and for the rest we cannot show that they are unused.

Flags:

- Nothing in the build reports unused images, which is how these accumulated. Markdown image references are resolved by webpack and do fail the build when a file is missing, but the 154 JSX src="/img/..." references are plain strings that nothing validates.
- Files under static/img are a public interface, not private to this repo. Worth documenting somewhere.
- The crawler skip-list entry for https://docs.tigera.io/img/calico-logo-2026-badge.png is now stale and can be removed. Left alone here to keep this PR to deletions.


ctauchen:
PTAL @breff This list doesn't include the files you recently asked about (for use on CC auth page). But if you are aware of anything else being used, please let me know.

PR #3009, opened 2026-09-04
Correct what survives the switch to eBPF mode

tomastigera:
The *Enable eBPF mode* section says preexisting connections "should not be disrupted". That holds only for TCP, so the page currently promises more than the dataplane delivers.

Replaced with what actually happens, per protocol: TCP connections survive including any address translation, because the translation lives in the Linux conntrack entry rather than in the `kube-proxy` rules Calico removes; non-TCP traffic is treated as new, so a UDP flow to a service with several backends can move to a different one; and connections over a VXLAN overlay do not survive, because the device is recreated in eBPF mode and the kernel drops the conntrack state with it.

Verified against the dataplane code and covered by a new functional verification test in projectcalico/calico#13778, which holds a TCP connection through a service across the switch and confirms it survives the removal of the `kube-proxy` rules.

Product Version(s): all — the sentence is identical and equally wrong in every version directory, so all 12 copies are corrected. Open Source 3.31 and 3.30 additionally note that a connection to a node port with a backend on another node is lost where the filter FORWARD policy is DROP: the fix for that (projectcalico/calico#13778) is in 3.33 and 3.32 but was not backported further.

Issue: CORE-12056

Link to docs preview: https://deploy-preview-3009--tigera.netlify.app

SME review:

DOCS review:

Additional information: the wording change came out of an investigation into which connections survive the iptables to eBPF switch. Written with Claude Code; I have reviewed every line and can explain the change.

### DIFF
--- a/calico-cloud/operations/ebpf/enabling-ebpf.mdx
+++ b/calico-cloud/operations/ebpf/enabling-ebpf.mdx
@@ -240,8 +240,17 @@ resource to `"BPF"`.
 kubectl patch installation.operator.tigera.io default --type merge -p '{"spec":{"calicoNetwork":{"linuxDataplane":"BPF"}}}'
 ```
 
-When enabling eBPF mode, preexisting connections continue to use the non-BPF datapath; such connections should
-not be disrupted, but they do not benefit from eBPF mode’s advantages.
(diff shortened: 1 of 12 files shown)


ctauchen:
Thanks for the updates, LGTM

PR #3028, opened 2026-09-15
DOCS-3047: Use the TigeraSecureEnterprise variant where the operator rejects the alias

ctauchen:
Installation examples told readers to set spec.variant to CalicoEnterprise. The operator only accepts that value from v1.43.0, which ships with Calico Enterprise 3.24, so on 3.23 and Calico Cloud 23 the API server rejects it. This is a partial revert of https://github.com/tigera/docs/pull/2628.

https://tigera.atlassian.net/browse/DOCS-3047

Changed in full, because the value is rejected:

- Calico Enterprise 3.23, which is the latest published version. Every 3.23 patch ships operator v1.42.x, whose CRD enum is Calico and TigeraSecureEnterprise only.
- Calico Cloud 23, which is based on Enterprise v3.23.1 and uses operator-cloud release-v1.42, with the same enum.

Changed in later versions, upgrade page only:

- The OpenShift page for upgrading Calico to Calico Enterprise, in 3.24-1, 3.24-2 and next. Both values work on 3.24, but during an upgrade the older operator is still running and reconciles the new CRDs back to its own schema, so the patch fails depending on timing. See https://tigera.atlassian.net/browse/EV-6874. The underlying fix, https://tigera.atlassian.net/browse/EV-6875, is still open and will not help upgrades from 3.23.

Left on CalicoEnterprise deliberately:

- All other 3.24 and next pages. The value is valid there, and none of them is reachable from an upgrade path.
- The unversioned calico-cloud tree, which is not published.
- The generated API references for Enterprise 3.24 and next, which already document both spellings.

The Calico Cloud API reference is included because PR 2628 hand-edited that generated file and nobody has regenerated it since. The four reverted lines restore what generation from v1.42 produces.

Preview pages:

- 3.23 OpenShift upgrade, the page in the ticket: https://deploy-preview-3028--calico-docs-preview-next.netlify.app/calico-enterprise/latest/getting-started/upgrading/upgrading-calico-to-calico-enterprise/upgrade-to-tsee-openshift
- 3.23 nftables, which quotes the downloaded custom-resources.yaml: https://deploy-preview-3028--calico-docs-preview-next.netlify.app/calico-enterprise/latest/operations/nftables
- 3.24 OpenShift upgrade: https://deploy-preview-3028--calico-docs-preview-next.netlify.app/calico-enterprise/3.24/getting-started/upgrading/upgrading-calico-to-calico-enterprise/upgrade-to-tsee-openshift
- Calico Cloud operator checklist, which shows status.variant: https://deploy-preview-3028--calico-docs-preview-next.netlify.app/calico-cloud/get-started/operator-checklist

Flags:

- The Calico Cloud part is inferred from the shared operator lineage. Worth a check with the Cloud team.


ctauchen:
PTAL @caseydavenport

PR #3030, opened 2026-09-16
Make the preexisting-connection text true per version

tomastigera:
Follow-up to #3009. That PR put one paragraph into 12 version directories, but three of its claims are not true in every version, and its own later wording reached only the 4 Open Source copies before it merged.

**VXLAN.** The device is only recreated where `vxlan_mgr.go` sets `FlowBased`. That code does not exist on Calico 3.30, Calico Enterprise 3.21 or Calico Cloud, so the stated cause is absent and the effect is unverified — the bullet is removed there. On 3.31, 3.32 and Enterprise 3.22/3.23 it is gated on `BPFEnabled && BTFEnabled`, so the bullet now names the BTF condition. Master, 3.33, Enterprise master and 3.24-x are unconditional and keep the plain wording.

**Forwarded flows.** The FORWARD accept restored by projectcalico/calico#13778 matches `MarkSeenFallThrough` plus ESTABLISHED,RELATED — any established flow, exactly like the `KUBE-FORWARD` accept it replaces. So on versions without it, every flow forwarded between two host interfaces is lost on a FORWARD-DROP node, including an external IP with a remote backend and traffic the node merely routes. #3009 said only node ports.

**Non-TCP.** `conntrack.h` gives the mid-flow fallthrough to TCP alone; everything else goes to full policy evaluation as a new flow. A flow allowed only as the reply of an already-established connection is therefore dropped, not just moved to another backend.

Also drops "the `kube-proxy` rules that $[prodname] removes" — with `BPFKubeProxyIptablesCleanupEnabled=false` it removes nothing — and applies #3009's terminology and apostrophe fixes to the 8 Enterprise and Cloud copies.

Product Version(s): all 12 directories, with the text now differing by version as above. The exception sentence goes to Calico 3.30, 3.31 and both Calico Cloud copies; Cloud is treated as not carrying the fix because its `releases.json` pins v3.21.0-1.0 and `release-calient-v3.21-1` has been frozen since February. Calico 3.32/3.33 and Enterprise 3.21/3.22/3.23 have the fix merged but not yet released, so their pages describe the next patch.

Issue: CORE-12056

Link to docs preview:

SME review:

DOCS review:

Additional information: found by a code review of #3009 against the Felix source on each product branch. Written with Claude Code; I have reviewed every line and can explain the change.


ctauchen (approved):
LGTM

PR #3032, opened 2026-09-17
[TSLA-11720] Derive the cc-operator image from the Calico Cloud version being installed

marcosmmb:
Product Version(s):
Calico Cloud v24.0.0 and later (applies to the current docs and version-23-2)

Issue:
https://tigera.atlassian.net/browse/TSLA-11720

Link to docs preview:

- /calico-cloud/get-started/setup-private-registry
- /calico-cloud/get-started/checklist

SME review:

DOCS review:

Additional information:

cc-operator is now released per Calico Cloud version and tagged with it, so the
image list takes the operator tag from the version you select instead of deriving
it with `helm show chart`. The old command always returned the newest chart's
operator even when an older version was selected, so this also fixes every older
version in the list.

Also in here: install state `complete` → `done` on the checklist page.

⚠️ Do not merge until cc-operator `v24.0.0` is tagged and `promote-prod` has run.
Until the live chart index serves the lockstep chart, Helm resolves `v1.27.2`,
which asks for `cc-operator:v1.27.2` — so the new image list would break
private-registry installs. The kubectl path is already per-version and unaffected.

Merge checklist:
- [ ] Test have passed


calico-cloud/get-started/install-cluster.mdx
```diff
@@ -59,6 +59,10 @@ You can quickly connect a cluster to Calico Cloud by generating a unique kubectl
    Monitor the status under **Connection Status**.
    When the status changes to **Connected**, installation is complete and your cluster is connected to Calico Cloud.
 
+:::note
+`installer.calicoCloudVersion` also selects the operator image, and an operator rejects an `Installer` that asks for a different version. To change version, generate a new command rather than editing this value.
```
tmjd: Why do you think this is helpful? Because it has always been the case that when following these directions it would be necessary to generate a new command to install a different version.

calico-cloud/get-started/install-private-registry.mdx
```diff
@@ -14,7 +14,7 @@ You can perform a Helm installation from images stored on a private registry.
 * You have at least one cluster that meets our [system requirements](system-requirements.mdx).
 * You have kubectl access to the cluster.
 * You have installed Helm 3.0 or later on your workstation.
-* You have [added the Calico Cloud images to a private registry](setup-private-registry.mdx), and you have the following information about the registry:
+* You have [added the Calico Cloud images to a private registry](setup-private-registry.mdx) for the version you are installing, including its operator image, and you have the following information about the registry:
```
tmjd: I don't think the operator image is anything special for setting up the private registry, so I'm not sure why we should mention it specifically here.

src/___new___/components/VersionedCloudImageList/index.js
```diff
@@ -41,13 +45,11 @@ const CodeBlockSelector = () => {
...
       </FormControl>
       <CodeBlock
-        title={`Helm command to add images for Calico Cloud ${selectedVersion}`}
-        language='javascript'
+        title={`Images for Calico Cloud ${selectedVersion}`}
+        language='bash'
       >
-        {`helm repo add calico-cloud ${variables.clouddownloadbase}/charts
-INSTALLER_IMAGE="${variables.cloudoperatorimage}:$(helm show chart calico-cloud/calico-cloud | grep version: | sed -e 's/version: *//' -e 's/+/-g/')"
+        {`OPERATOR_IMAGE=${variables.cloudoperatorimage}:${operatorTag(selectedVersion)}
```
tmjd: I don't think there is any reason to have OPERATOR_IMAGE separate from the IMAGES list/array. So if this image is added to that list then OPERATOR_IMAGE variable can be removed and the additional changes added to copy it can be removed also.

calico-cloud/get-started/setup-private-registry.mdx
```diff
@@ -50,6 +49,8 @@ Add the required $[prodname] images to a private registry.
 
 <VersionedCloudImageList />
 
+`OPERATOR_IMAGE` is the $[prodname] operator, which is released per $[prodname] version and tagged with it; `IMAGES` is everything else. Mirror both for each version you plan to install.
```
tmjd: As I suggested elsewhere, the image should be part of the IMAGES, so there is no need for this line.

calico-cloud/get-started/setup-private-registry.mdx
```diff
@@ -64,12 +65,14 @@ IMAGEPATH=""
...
-| `quay.io/tigera/cc-operator:v4.5.6` | `my.registry/` |               | `my.registry/tigera/cc-operator:v4.5.6`      |
-| `quay.io/tigera/cc-operator:v4.5.6` | `my.registry/` | `custom-path` | `my.registry/custom-path/cc-operator:v4.5.6` |
+| Original image                       | Image Registry | Image Path    | Private registry image                                |
+| ------------------------------------ | -------------- | ------------- | ----------------------------------------------------- |
+| `quay.io/tigera/typha:v1.2.3`        | `my.registry/` |               | `my.registry/tigera/typha:v1.2.3`                     |
+| `quay.io/tigera/typha:v1.2.3`        | `my.registry/` | `custom-path` | `my.registry/custom-path/typha:v1.2.3`                |
+| `quay.io/tigera/cc-operator:v23.0.2` | `my.registry/` |               | `my.registry/calico-cloud-public/cc-operator:v23.0.2` |
+| `quay.io/tigera/cc-operator:v23.0.2` | `my.registry/` | `custom-path` | `my.registry/custom-path/cc-operator:v23.0.2`         |
+
+The operator is the exception: with no image path set, it is pulled from the `calico-cloud-public` path in your registry rather than `tigera`.
```
tmjd: Was this table correct before? If so then we should fix thing so it continues to use `tigera` as its path.

calico-cloud/get-started/setup-private-registry.mdx
```diff
@@ -83,6 +86,7 @@ for image in ${IMAGES[@]}; do
   img_base=$(echo ${image} | sed "s#^.*/\([^/]*/[^/]*$\)#\1#")
   crane cp ${image} ${REGISTRY}${img_base} || break
 done
+crane cp ${OPERATOR_IMAGE} ${REGISTRY}calico-cloud-public/${OPERATOR_IMAGE##*/}
```
tmjd: As mentioned this shouldn't be needed.

calico-cloud/get-started/setup-private-registry.mdx
```diff
@@ -93,6 +97,7 @@ for image in ${IMAGES[@]}; do
   img_base=$(echo ${image} | sed "s#^.*/##")
   crane cp ${image} ${REGISTRY}${IMAGEPATH}/${img_base} || break
 done
+crane cp ${OPERATOR_IMAGE} ${REGISTRY}${IMAGEPATH}/${OPERATOR_IMAGE##*/}
```
tmjd: ditto

calico-cloud/get-started/upgrade-cluster.mdx
```diff
@@ -9,11 +9,15 @@ To upgrade a managed cluster to the latest version of $[prodname]:
...
 1. If your organization uses multiple projects to group managed clusters, click the **Project** menu and select the project you want your cluster to be part of.
 1. For the cluster you want to upgrade, select **Actions** > **Reinstall**.
-1. In the **Reinstall Cluster** dialog, select a newer version of $[prodname] from the list, click **Reinstall**, and copy the generated kubectl command.
+1. In the **Reinstall Cluster** dialog, select a newer version of $[prodname] from the list, click **Reinstall**, and copy the generated command.
 1. From a terminal, paste and run the command.
    The cluster's status under **Connection Status** changes to **Disconnected: Installing**.
    When the upgrade is complete, the status changes to **Connected**.
 
+:::important
+If you install from a private registry, [mirror the new version's images](setup-private-registry.mdx), including its operator image, before you run the command.
```
tmjd: Do you think there is a reason to call out the installer image here? If so please change the message to say installer rather than operator.

calico-cloud_versioned_docs/version-23-2/get-started/checklist.mdx
tmjd, line 1: you shouldn't be making changes to the version 23-2 docs for the cc-operator updates. A new set of version docs will be created from the unversioned docs.

Makefile
```diff
@@ -264,22 +264,17 @@ build-crd-reference-docs: $(CRD_DOC_GENERATOR)
...
-run-update-cloud-image-list:
-	RUN_UPDATE_CLOUD_IMAGE_LIST=1 PRODUCT=calico-cloud make update-cloud-image-list
-	for x in $$(ls calico-cloud_versioned_docs/); do \
-		RUN_UPDATE_CLOUD_IMAGE_LIST=1 PRODUCT=calico-cloud_versioned_docs/$$x make update-cloud-image-list; \
-	done
-	@if [ "$$(git diff --stat ./calico-cloud*/**/get-started/connect/setup-private-registry.mdx)" != "" ]; then \
-	echo "You might need to run 'make run-update-cloud-image-list' and commit the changes"; exit 1; fi
+	@if [ -z "$(CC_VERSION)" ] || [ -z "$(MANIFEST_VERSION)" ]; then \
+		echo "usage: make update-cloud-image-list CC_VERSION=vXX.Y.Z MANIFEST_VERSION=v3.XX.Y-N"; false; fi
+	./scripts/update-cloud-image-list.sh "$(CC_VERSION)" "$(MANIFEST_VERSION)"
```
tmjd: @ctauchen have you been using this make target to update the image list? My memory is that it didn't work on Mac so you just used a curl command to fetch the list and update the `src/___new___/data/ccImageLists.js` file when there were new updates needed.
I'll leave it up to Chris if we want to update this target and add the script or if not remove the make target(s).

tmjd (commented):
I think one thing we should do is stop calling cc-operator, just 'operator' we already have something else that gets referred to as 'operator' (tigera/operator). I think we should continue referring to it in the docs as the 'installer' since that is more descriptive for the user.

I don't know if we want to start the release notes but I think we need to include something in the release notes for private registry users that when they update to use the latest helm chart they'll also need to ensure they pull the cc-operator images for older CC versions they'll be installing because they'll need the CC v2X.Y.Z image that the helm chart will be attempting to use rather than the v1.x image that would have been used for all the installs.

Makefile
ctauchen: Yeah, I've just been pulling things manually with curl. I've been meaning to clear out a lot of unused content from the Makefile.

calico-cloud/get-started/checklist.mdx
```diff
@@ -20,14 +20,20 @@ kubectl get installer default --namespace calico-cloud -o jsonpath --template '{
 {"clusterName":"my-new-cluster","state":"installing"}
 ```
 
-After `state` is `complete`, $[prodname] is properly installed.
+`state` moves through `waiting` and `installing` to `done`. If `state` is `error`, `message` says why.
```
ctauchen: Is this verified? The doc is stale, so certainly could be. Rather than take us through everything, I'd suggest we just report the end state. So keep the original line but change complete to done, if that's true.

calico-cloud/get-started/checklist.mdx
```diff
@@ -20,14 +20,20 @@ kubectl get installer default --namespace calico-cloud -o jsonpath --template '{
...
 
 ## Check logs for fatal errors
 
-Check that the $[prodname] operator is running and that logs do not have any fatal errors.
+Check that the $[prodname] operator is running and that logs do not have any fatal errors. The deployment is named `calico-cloud-controller-manager` on a kubectl install and `calico-cloud` on a Helm install, so select it by label. The first line reports which version the operator installs.
```
ctauchen: I'm not sure what this is for. Do we need it?

calico-cloud/get-started/install-automated.mdx
```diff
@@ -56,7 +56,7 @@ The following parameters are required for all Calico Cloud installations.
 | Parameter | Value | Example | Description |
 | -- | -- | -- | -- |
 | `installer.clusterName` | string | `cluster-name` | The name given to your managed cluster in Calico Cloud. |
-| `installer.calicoCloudVersion` | string | `$[cloudUserVersion]` | The version of Calico Cloud you're installing. |
+| `installer.calicoCloudVersion` | string | `$[cloudUserVersion]` | The version of Calico Cloud you're installing. This also selects the operator image, which is tagged with that version. |
```
ctauchen: As Erik mentions later, this extra information about the operator image isn't really needed.

calico-cloud/get-started/install-automated.mdx
```diff
@@ -75,6 +75,10 @@ If you're using a private registry, you must set the following parameters.
 | `installer.imagePath` | string | `path/to/directory` | The path to a directory in your private registry that contains images required to install Calico Cloud. |
 | `imagePullSecrets.name` | string | `secret-name` | The name of the image pull secret you will use to allow access from the `calico-cloud` namespace to your private registry. |
 
+:::important
+The operator image is pulled at the tag named by `installer.calicoCloudVersion`, so [mirror that version's operator image](setup-private-registry.mdx) first. With `installer.imagePath` unset, it is pulled from the `calico-cloud-public` path in your registry rather than `tigera`.
+:::
+
```
ctauchen: This seems unnecessary, as Erik mentions later.

calico-cloud/get-started/upgrade-cluster.mdx
```diff
@@ -9,11 +9,15 @@ To upgrade a managed cluster to the latest version of $[prodname]:
 1. From the web console, go to the **Managed Clusters** page.
 1. If your organization uses multiple projects to group managed clusters, click the **Project** menu and select the project you want your cluster to be part of.
 1. For the cluster you want to upgrade, select **Actions** > **Reinstall**.
-1. In the **Reinstall Cluster** dialog, select a newer version of $[prodname] from the list, click **Reinstall**, and copy the generated kubectl command.
+1. In the **Reinstall Cluster** dialog, select a newer version of $[prodname] from the list, click **Reinstall**, and copy the generated command.
```
ctauchen: I've actually got a PR up now that addresses this.

Makefile
ctauchen: I think that automating this is a good goal, but I'd prefer to look at that separately to this PR.

ctauchen (commented):
Thanks for this @marcosmmb ! I mostly echo Erik's comments. This seems to be a change that happens almost entirely in the background, with few user-facing changes. Some of the changes here fix places where we name the old image. But many changes are just signposting the new image.

If there's something I'm missing, do let me know. Are there things that users need to know about this change?

ctauchen:
@tmjd You happy with this now? I think you're better positioned to review the new flow.

tmjd (approved):

ctauchen (approved):
LGTM, thanks!

